PR #3, opened 2023-04-27
ModelConfiguration

Atticus1806:
This is a first implementation of the ModelConfiguration base class and some basic tests. 
Feel free to discuss, since we didn't have a strong opinion on the decisions most of the time.
E.g. we removed the `metadata` helpers since we decided they are most likely not used.
Also naming is up for discussion as well as documentation etc.

### DIFF
--- a/i6_models/assemblies/util.py
+++ b/i6_models/assemblies/util.py
@@ -1,6 +1,7 @@
 from dataclasses import MISSING, dataclass
 from typing import Any, List, Optional
 
+
 @dataclass
 class ModelConfiguration:
     """base dataclass that supported fetching attributes and metas"""
@@ -19,7 +20,7 @@ class ModelConfiguration:
 
     def _get_default(self, attribute_name: str) -> Any:
         if hasattr(self, attribute_name):
-                return getattr(self, attribute_name)
+            return getattr(self, attribute_name)
 
         f = self.__dataclass_fields__[attribute_name]
         if not f.default_factory is not MISSING:
@@ -43,15 +44,3 @@ class ModelConfiguration:
                     setattr(config, k, getattr(args, k))
 
             return config
-
-@dataclass
-class TestConfiguration(ModelConfiguration):
-
-    num_layers: int = 5
-    hidden_dim: int = 256
-    name: str = "Cool Model Configuration"
-
-class TestNestedConfiguration(ModelConfiguration):
-
-    encoder_config: ModelConfiguration = TestConfiguration(num_layers=4, hidden_dim=3, name="HSG")
-    hidden_dim: int = 333
--- a/i6_models/assemblies/util.py
+++ b/i6_models/assemblies/util.py
@@ -4,7 +4,7 @@ from typing import Any, List, Optional
 
 @dataclass
 class ModelConfiguration:
-    """base dataclass that supported fetching attributes and metas"""
+    """base dataclass that supported fetching attributes"""
 
     _name: Optional[str] = None
 
@@ -32,6 +32,11 @@ class ModelConfiguration:
 
     @classmethod
     def from_namespace(cls, args):
+        """
+        Generates a ModelConfiguration from a given dataclass.
+        If its already of the correct instance it will return the input object, otherwise it creates an object
+        from the matching attributes.
+        """
         if isinstance(args, cls):
             return args
         else:
--- a/i6_models/assemblies/util.py
+++ b/i6_models/assemblies/util.py
@@ -1,0 +1,72 @@
+from dataclasses import MISSING, dataclass
+from typing import Any, List, Optional
+
+@dataclass
+class ModelOptions:
+    """base dataclass that supported fetching attributes and metas"""
+
+    _name: Optional[str] = None
+
+    @staticmethod
+    def name():
+        return None
+
+    def _get_all_attributes(self) -> List[str]:
+        return [k for k in self.__dataclass_fields__.keys()]
+
+    def _get_meta(
+        self, attribute_name: str, meta: str, default: Optional[Any] = None
+    ) -> Any:
+        return self.__dataclass_fields__[attribute_name].metadata.get(meta, default)
+
+    def _get_name(self, attribute_name: str) -> str:
+        return self.__dataclass_fields__[attribute_name].name
+
+    def _get_default(self, attribute_name: str) -> Any:
+        if hasattr(self, attribute_name):
+            if str(getattr(self, attribute_name)).startswith("${"):
+                return str(getattr(self, attribute_name))
+            elif str(self.__dataclass_fields__[attribute_name].default).startswith(
+                "${"
+            ):
+                return str(self.__dataclass_fields__[attribute_name].default)
+            elif (
+                getattr(self, attribute_name)
+                != self.__dataclass_fields__[attribute_name].default
+            ):
+                return getattr(self, attribute_name)
+
+        f = self.__dataclass_fields__[attribute_name]
+        if not f.default_factory is not MISSING:
+            return f.default_factory()
+        return f.default
+
+    def _get_type(self, attribute_name: str) -> Any:
+        return self.__dataclass_fields__[attribute_name].type
+
+    def _get_help(self, attribute_name: str) -> Any:
+        return self._get_meta(attribute_name, "help")
+
+    def _get_argparse_const(self, attribute_name: str) -> Any:
+        return self._get_meta(attribute_name, "argparse_const")
+
+    def _get_argparse_alias(self, attribute_name: str) -> Any:
+        return self._get_meta(attribute_name, "argparse_alias")
+
+    def _get_choices(self, attribute_name: str) -> Any:
+        return self._get_meta(attribute_name, "choices")
+
+    @classmethod
+    def from_namespace(cls, args):
+        if isinstance(args, cls):
+            return args
+        else:
+            config = cls()
+            for k in config.__dataclass_fields__.keys():
+                if k.startswith("_"):
+                    # private member, skip
+                    continue
+                if hasattr(args, k):
+                    setattr(config, k, getattr(args, k))
+
+            return config
--- a/i6_models/assemblies/util.py
+++ b/i6_models/assemblies/util.py
@@ -23,7 +23,7 @@ class ModelConfiguration:
             return getattr(self, attribute_name)
 
         f = self.__dataclass_fields__[attribute_name]
-        if not f.default_factory is not MISSING:
+        if f.default_factory is not MISSING:
             return f.default_factory()
         return f.default
 
--- a/i6_models/assemblies/util.py
+++ b/i6_models/assemblies/util.py
@@ -2,7 +2,7 @@ from dataclasses import MISSING, dataclass
 from typing import Any, List, Optional
 
 @dataclass
-class ModelOptions:
+class ModelConfiguration:
     """base dataclass that supported fetching attributes and metas"""
 
     _name: Optional[str] = None
@@ -14,26 +14,11 @@ class ModelOptions:
     def _get_all_attributes(self) -> List[str]:
         return [k for k in self.__dataclass_fields__.keys()]
 
-    def _get_meta(
-        self, attribute_name: str, meta: str, default: Optional[Any] = None
-    ) -> Any:
-        return self.__dataclass_fields__[attribute_name].metadata.get(meta, default)
-
     def _get_name(self, attribute_name: str) -> str:
         return self.__dataclass_fields__[attribute_name].name
 
     def _get_default(self, attribute_name: str) -> Any:
         if hasattr(self, attribute_name):
-            if str(getattr(self, attribute_name)).startswith("${"):
-                return str(getattr(self, attribute_name))
-            elif str(self.__dataclass_fields__[attribute_name].default).startswith(
-                "${"
-            ):
-                return str(self.__dataclass_fields__[attribute_name].default)
-            elif (
-                getattr(self, attribute_name)
-                != self.__dataclass_fields__[attribute_name].default
-            ):
                 return getattr(self, attribute_name)
 
         f = self.__dataclass_fields__[attribute_name]
@@ -43,18 +28,6 @@ class ModelOptions:
 
     def _get_type(self, attribute_name: str) -> Any:
         return self.__dataclass_fields__[attribute_name].type
-
-    def _get_help(self, attribute_name: str) -> Any:
-        return self._get_meta(attribute_name, "help")
-
-    def _get_argparse_const(self, attribute_name: str) -> Any:
-        return self._get_meta(attribute_name, "argparse_const")
-
-    def _get_argparse_alias(self, attribute_name: str) -> Any:
-        return self._get_meta(attribute_name, "argparse_alias")
-
-    def _get_choices(self, attribute_name: str) -> Any:
-        return self._get_meta(attribute_name, "choices")
 
     @classmethod
     def from_namespace(cls, args):
@@ -70,3 +43,15 @@ class ModelOptions:
                     setattr(config, k, getattr(args, k))
 
             return config
+
+@dataclass
+class TestConfiguration(ModelConfiguration):
+
+    num_layers: int = 5
+    hidden_dim: int = 256
+    name: str = "Cool Model Configuration"
+
+class TestNestedConfiguration(ModelConfiguration):
+
+    encoder_config: ModelConfiguration = TestConfiguration(num_layers=4, hidden_dim=3, name="HSG")
+    hidden_dim: int = 333
--- a/.github/workflows/model_tests.yml
+++ b/.github/workflows/model_tests.yml
@@ -1,0 +1,25 @@
+name: model_tests
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+jobs:
+  test-jobs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          repository: "rwth-i6/i6_models"
+          path: "i6_models"
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+          cache: 'pip'
+      - run: pip install pytest
+      - name: Test Models
+        run: |
+          set -e
+          pytest i6_models/tests
--- a/tests/assemblies_tests/test_util.py
+++ b/tests/assemblies_tests/test_util.py
@@ -1,0 +1,44 @@
+from dataclasses import dataclass
+
+from i6_models.assemblies.util import ModelConfiguration
+
+def test_simple_configuration():
+    @dataclass
+    class TestConfiguration(ModelConfiguration):
+        num_layers: int = 5
+        hidden_dim: int = 256
+        name: str = "Cool Model Configuration"
+
+    test_cfg = TestConfiguration(num_layers=12, name="Even Cooler Model Configuration")
+    assert test_cfg.hidden_dim == 256
+    assert test_cfg.name == "Even Cooler Model Configuration"
+    assert test_cfg.num_layers == 12
+    test_cfg.num_layers = 7
+    assert test_cfg.num_layers == 7
+
+def test_nested_configuration():
+
+    @dataclass
+    class TestConfiguration(ModelConfiguration):
+        num_layers: int = 5
+        hidden_dim: int = 256
+        name: str = "Cool Model Configuration"
+
+    @dataclass
+    class TestNestedConfiguration(ModelConfiguration):
+        encoder_config: TestConfiguration = TestConfiguration(num_layers=4, hidden_dim=3, name="encoder_config")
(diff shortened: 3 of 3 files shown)


Marvin84:
Just an additional comment that we discussed: the presented structure should have all we need for an atomic Config object. We can then construct nested configs for a better definition of the structure and hierarchy of/between different parameters.

i6_models/assemblies/util.py
```diff
@@ -0,0 +1,51 @@
...
+        f = self.__dataclass_fields__[attribute_name]
+        if f.default_factory is not MISSING:
+            return f.default_factory()
+        return f.default
+
+    def _get_type(self, attribute_name: str) -> Any:
+        return self.__dataclass_fields__[attribute_name].type
+
+    @classmethod
+    def from_namespace(cls, args):
```
michelwi, line 34: do we really want to silently drop all config params that do not match? This seems error prone to me.

Why do we even need such a config class converter?

i6_models/assemblies/util.py
```diff
@@ -0,0 +1,51 @@
+from dataclasses import MISSING, dataclass
```
curufinwe, line 1: This should not live in assemblies. It can also be used in parts. I would put it in /i6_models/config.py

i6_models/assemblies/util.py
Atticus1806: So the usecase I found to not drop this was that you could have one big custom Config to use for your model which then initializes each subconfig you later need for the model.
So e.g. `TransformerConfig` can then initialize `TransformerEncoder` and `TransformerDecoder` etc. 
But if this makes sense or should be dropped is def. up for discussion.

i6_models/config.py
```diff
@@ -0,0 +1,51 @@
...
+    def _get_default(self, attribute_name: str) -> Any:
+        if hasattr(self, attribute_name):
+            return getattr(self, attribute_name)
+
+        f = self.__dataclass_fields__[attribute_name]
+        if f.default_factory is not MISSING:
+            return f.default_factory()
+        return f.default
+
+    def _get_type(self, attribute_name: str) -> Any:
```
albertz, line 30: Where is this function used?

i6_models/config.py
```diff
@@ -0,0 +1,51 @@
...
+    def name() -> Optional[str]:
+        return None
+
+    def _get_all_attributes(self) -> List[str]:
+        return [k for k in self.__dataclass_fields__.keys()]
+
+    def _get_name(self, attribute_name: str) -> str:
+        return self.__dataclass_fields__[attribute_name].name
+
+    def _get_default(self, attribute_name: str) -> Any:
```
albertz, line 21: Where is this function used?

i6_models/config.py
```diff
@@ -0,0 +1,51 @@
+from dataclasses import MISSING, dataclass
```
albertz, line 1: I would add some docstring explaining (high-level) what the logic of the config would be.

i6_models/config.py
Atticus1806: I don't understand. This really depends on your usecase. Here this is just abstract implementation for use in derived configs. Same goes for similar comments.

i6_models/config.py
Atticus1806: Do you want an example (e.g. one of the classes from the tests)? Because for now this is not really set I think. For now we wanted a base definition model configurations are built upon.

i6_models/config.py
```diff
@@ -0,0 +1,51 @@
...
+
+@dataclass
+class ModelConfiguration:
+    """base dataclass that supported fetching attributes"""
+
+    _name: Optional[str] = None
+
+    @staticmethod
+    def name() -> Optional[str]:
+        return None
```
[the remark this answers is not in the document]
Atticus1806: If others agree this can def. be changed. For now this is just similar to the Fairseq reference implementation. But I assume that in this case yes each config should overwrite this if use is desired.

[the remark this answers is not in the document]
Atticus1806: For most of the functions here (and the ones I dropped) I found that they are barely used, but still I thought they provide nice optional functionality.

JackTemaki:
@Atticus1806 can you edit the initial post to include a link to the fairseq reference?

curufinwe (approved):

i6_models/config.py
```diff
@@ -0,0 +1,51 @@
+from dataclasses import MISSING, dataclass
+from typing import Any, List, Optional
+
+
+@dataclass
+class ModelConfiguration:
+    """base dataclass that supported fetching attributes"""
```
vieting, line 7:
```suggestion
    """base dataclass that supports fetching attributes"""
```

i6_models/config.py
```diff
@@ -0,0 +1,51 @@
...
+        f = self.__dataclass_fields__[attribute_name]
+        if f.default_factory is not MISSING:
+            return f.default_factory()
+        return f.default
+
+    def _get_type(self, attribute_name: str) -> Any:
+        return self.__dataclass_fields__[attribute_name].type
+
+    @classmethod
+    def from_namespace(cls, args):
```
vieting, line 34: Can you please add type annotation?

i6_models/config.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        Generates a ModelConfiguration from a given dataclass.
+        If its already of the correct instance it will return the input object, otherwise it creates an object
```
vieting, line 37:
```suggestion
        If it's already of the correct instance, it will return the input object, otherwise it creates an object
```

i6_models/config.py
albertz: But you must have some basic understanding of how this model config would be used, or not?

albertz: It would also be helpful to document the inspiration from Fairseq.

i6_models/config.py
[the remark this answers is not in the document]
albertz: But still, I don't understand the purpose of this function. It just returns `None`. So is this supposed to be overwritten? I don't understand the purpose. It this just for debugging purpose, or does this have any actual semantic meaning?

i6_models/config.py
albertz: I don't see any other code which uses this. So if some code is not used, I would remove it. We should not have code here which might be used later by other code but is not used yet.

i6_models/config.py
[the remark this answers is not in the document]
Atticus1806: So this is what I understand. But then I was not able to find a Config which overwrites this, so maybe we can remove it for now.

i6_models/config.py
Atticus1806: This class is supposed to offer basic functionality for configs and while you are right that we should carefully decide which functions are helpful and which are not, I think `_get_type` is something that will be used. Similar to the other 3 private functions. Regarding the name, maybe we should remove that for now, I am not sure.

i6_models/config.py
Atticus1806: > But you must have some basic understanding of how this model config would be used

Yes, construction for example as in the tests. But for example then the later use for models and integration into our pipeline is not fully decided from my understanding. What I was asking was, whether you want an example or a rather conceptual explaining here.

i6_models/config.py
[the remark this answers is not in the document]
albertz: Yes, I would remove code which we don't need for now. Once we need it, we can add it.

i6_models/config.py
Atticus1806: I am not 100% sure about the typing myself. So from what I understand this can be any object on which has attributes, so maybe `Any` is correct here?

i6_models/config.py
albertz: Ok, but then my question is, where will it be used, how will it be used? I don't really see any documentation here, nor any usage, so it's hard to just guess the potential usage. It would help to see some actual code where it is used, and maybe also some documentation.

As this is a private method, I assume the usage will only be internal. So what internal code would use it? Why did you not include this basic functionality?

i6_models/config.py
albertz: If possible, both, a conceptual explaining, and a minimal example.
I'm fine with the most minimal (1 line of code) example you can think of.
The conceptual explanation can also be short for now (one sentence).
A reference to the Fairseq code would also be very helpful.

albertz: I also mean a costring for the module itself.

i6_models/config.py
Atticus1806: An example would be to check for proper initialization of the config. E.g. you would want `num_layers` to always be int. So instead of checking for each parameter individually you can check whether the value the user gives is the same type as the field.

i6_models/config.py
Atticus1806: So how the config then later is used in the module? This is highly up to the user and also not set yet. 
The idea discussed in the meeting was to provide a base config class similar to Fairseq, which can then be used to build the first modules. But this can be added later once we have a "common" way of using these configs.

i6_models/config.py
albertz: Who do you mean by "you"? The user? I thought this is only for internal use, i.e. the user would never use it. It the user is supposed to use it, it should not be marked as private, i.e. not start with underscore.

Is type checking supposed to be done anyway automatically? I would have expected that all types would always be checked, or not? But where is that code?

i6_models/config.py
albertz: I'm fine with whatever minimal example you can think of, and whatever most minimal explanation you can give.

albertz: Looks good!

i6_models/config.py
Atticus1806: >Is type checking supposed to be done anyway automatically?

No. This was just some example I can up with.

>I thought this is only for internal use

Yes. From this e.g. `ConformerConfig` is derived where then `_get_type` would be used (internally). Type checking was just an example I came up with.

Atticus1806: I think the major discussion point here is, which functions this base class should provide and which are deemed not helpful for now. Maybe we can discuss this in todays meeting, or other can comment? 
I don't really have a strong opinion on those function, those where just the ones I thought would be helpful looking at the Fairseq class.

albertz: So, using the example:
```python
@dataclass
class ExampleConfig(ModelConfiguration):
    hidden_dim: int = 256
    name: str = "Example Configuration"

class ExampleModule(Module):
    def __init__(self, cfg: ExampleConfig)
        self.hidden_dim = cfg.hidden_dim
```
Now when the user writes:
```python
mod = ExampleModule(ExampleConfig(hidden_dim=2.5))
```
I thought that some code here would check the types and raise an exception here that `hidden_dim` is wrong. But this would be handled by `ModelConfiguration` automatically for all attribs. This is why I thought we have this `_get_type` method. Otherwise I still don't really understand why we need the `_get_type` method.

Or do you think that in `ExampleModule`, there would be an extra check like:
```python
assert isinstance(cfg.hidden_dim, cfg._get_type("hidden_dim"))
```
But this would be bad, as you would access a private method from outside. Also, this is not really nice to read. In this example, I would simply prefer:
```python
assert isinstance(cfg.hidden_dim, int)
```
Even if you would replicate the type `int` both here and in the `ExampleConfig` definition.

Or is this not the purpose of the function? I still don't really see the purpose. The only purpose which makes sense for me is if this would be used by some automatic type checking for all attribs. But that's why I was asking for some examples of potential usages.

SimBe195: > do you mean we should add code to the config class to do type checking automatically?

I think it's a good idea to force correct typing in the config and throw an exception if this is violated.

albertz: I was also wondering whether `dataclass` itself already supports such a functionality. Maybe by defining `_get_type`, this is actually the way how `dataclass` would handle it? I don't know. That's why I was asking here.

albertz: [Relevant StackOverflow question](https://stackoverflow.com/questions/50563546/validating-detailed-types-in-python-dataclasses).

michelwi: a quick search brought me to this library that enables type checking at runtime with a decorator 
https://pypi.org/project/typeguard/

we could just add this to the dataclass and all should be checked automatically

albertz: > a quick search brought me to this library that enables type checking at runtime with a decorator https://pypi.org/project/typeguard/
> 
> we could just add this to the dataclass and all should be checked automatically

Are you sure this works here?

And this would not need the `_get_type` method here, right?

Atticus1806: I am currently trying to set `typeguard` up, but I think it is not that straight forward.

Atticus1806: ESPnet actually wraps the `check_type` function. 
https://espnet.github.io/espnet/_modules/espnet2/utils/build_dataclass.html

albertz: See my StackOverflow link above. It is specifically about possible solutions for `dataclass`.

i6_models/config.py
```diff
@@ -0,0 +1,63 @@
...
+
+class ExampleModule(Module):
+    __init__(self, cfg: ExampleConfig)
+        self.hidden_dim = cfg.hidden_dim
+
+This config can then be used in the construction of the model to provide parameters.
+Approach is inspired by the Fairseq: https://github.com/facebookresearch/fairseq/blob/main/fairseq/dataclass/configs.py
+"""
+
+from dataclasses import MISSING, dataclass
```
SimBe195, line 17:
```suggestion
from dataclasses import MISSING, dataclass, fields
import typeguard
```

Atticus1806:
I just added typechecks and a test which does that. Additionally I removed the private functions, since the opinion seems to be that we only add them when we need them.

i6_models/config.py
```diff
@@ -0,0 +1,63 @@
...
+            return getattr(self, attribute_name)
+
+        f = self.__dataclass_fields__[attribute_name]
+        if f.default_factory is not MISSING:
+            return f.default_factory()
+        return f.default
+
+    def _get_type(self, attribute_name: str) -> Any:
+        return self.__dataclass_fields__[attribute_name].type
+
```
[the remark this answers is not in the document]
Atticus1806: The difference between the two is that this way its always enforced, while with the decorator one could not apply it to certain configs and have those not type checked. I guess enforcing it is preferred?

tests/test_config.py
```diff
@@ -0,0 +1,85 @@
...
+        name: str = "Cool Model Configuration"
+
+    from typeguard import TypeCheckError
+
+    TestConfiguration(num_layers=2, hidden_dim=1)
+    try:
+        TestConfiguration(num_layers=2.0, hidden_dim="One")
+        assert False, "Typing should not allow this"
+    except TypeCheckError:
+        print("Typing Test worked")
```
SimBe195, line 85:
```suggestion
    with pytest.raises(TypeCheckError):
        TestConfiguration(num_layers=2.0, hidden_dim="One")
```

i6_models/config.py
vieting: Yes, `Any` should be OK.

tests/test_config.py
```diff
@@ -0,0 +1,83 @@
...
+def test_config_typing():
+    @dataclass
+    class TestConfiguration(ModelConfiguration):
+        num_layers: int = 4
+        hidden_dim: int = 13
+        name: str = "Cool Model Configuration"
+
+    from typeguard import TypeCheckError
+
+    TestConfiguration(num_layers=2, hidden_dim=1)
```
vieting, line 81:
```suggestion
```
This isn't needed, is it?

i6_models/config.py
```diff
@@ -0,0 +1,56 @@
...
+                    setattr(config, k, getattr(args, k))
+
+            return config
+
+    def _validate_types(self) -> None:
+        for field in fields(type(self)):
+            try:
+                typeguard.check_type(getattr(self, field.name), field.type)
+            except typeguard.TypeCheckError as e:
+                raise typeguard.TypeCheckError(f'In field "{field.name}" of "{type(self)}": {e}')
```
albertz, line 53: `e` is considered an ambiguous var name. I would use `exc`.
```suggestion
            except typeguard.TypeCheckError as exc:
                raise typeguard.TypeCheckError(f'In field "{field.name}" of "{type(self)}": {exc}')
```

SimBe195 (approved):

i6_models/config.py
```diff
@@ -0,0 +1,56 @@
...
+
+@dataclass
+class ModelConfiguration:
+    """
+    Base dataclass that supports fetching attributes.
+    In order to enforce typing in a Config, decorate that Config with @enforce_types.
+    """
+
+    @classmethod
+    def from_namespace(cls, args: Any):
```
albertz, line 29: Sorry to ask a bit late on this, but now I wonder: What is this function `from_namespace` actually used for?

albertz, line 29:
```suggestion
    def from_namespace(cls, args: Any) -> ModelConfiguration:
```

i6_models/config.py
```diff
@@ -0,0 +1,56 @@
...
+    hidden_dim: int = 256
+    name: str = "Example Configuration"
+
+class ExampleModule(Module):
+    __init__(self, cfg: ExampleConfig)
+        self.hidden_dim = cfg.hidden_dim
+
+This config can then be used in the construction of the model to provide parameters.
+Approach is inspired by the Fairseq: https://github.com/facebookresearch/fairseq/blob/main/fairseq/dataclass/configs.py
+"""
```
albertz, line 15: Add empty line. And also import future annotation support.
```suggestion
"""

from __future__ import annotations
```

i6_models/config.py
Atticus1806: I would use it to initialize some configs with other existing ones. If I have a larger config with more parameters and want to initialize e.g. my encoder config with it, it can be done this way without having having to manually set all parameters again.

albertz: Is this really a good idea? What is the benefit of doing this? I assume that you use this larger config for multiple different modules? But then I can imagine that there can be very nasty bugs or problems. E.g. when some option is maybe interpret differently depending on module (`dropout`, `momentum` and some others can be interpret as 1 - x or x), or some option name is not really consistent?

Didn't we actually discourage to do this for RASR, i.e. setting `*.momentum = 0.1` or so?

Can you add a docstring explaining this?

Atticus1806: I think also @michelwi wasn't a big fan of this either. This is what I would imagine to use it for, but if people rather prefer not to have this, due to the high chance of errors, then I can just remove it (and possible re-add it if we want this later on).
But then I would remove the Fairseq reference, because that was the last Fairseq part :D

albertz: It's a bit difficult to discuss on this without seeing actual examples. Maybe you can find some in Fairseq?

@mmz33 do you know whether Zoltan uses this?

It still make sense to keep the Fairseq reference, but remove the "inspired from" part, just list it as a possible alternative, or extension. It might also make sense to list some other alternatives.

In general, it would be a good idea to have a short comparison to some of the popular alternatives. I know Lingvo also has such a configuration system (@mmz33 can maybe say sth about it). But I think there are also a few popular standalone packages for managing the configuration.

Btw, sorry if this question is a bit late: Did we actually look at some of these popular standalone packages, and did we consider to use just something existing, instead of doing our own thing here? Why did we decide against reusing sth existing?

Ok, when our `ModelConfiguration` is actually really minimal, like it is now, esp after removing `from_namespace`, it is not really needed.

Atticus1806: The Fairseq Transformer makes use of it but I feel its hart to follow the code because it not really documented, e.g:
https://github.com/facebookresearch/fairseq/blob/3f6ba43f07a6e9e2acf957fc24e57251a7a3f55c/fairseq/models/transformer/transformer_decoder.py#L467

In the first meeting we talked about this a bit and decided that we wanted to stay close to the style Fairseq does it.

albertz: I rather meant examples of configurations where options are just put into a flat namespace. To see whether such configuration is easy to understand, or whether it rather makes things ambiguous.

michelwi: > Btw, sorry if this question is a bit late: Did we actually look at some of these popular standalone packages, and did we consider to use just something existing, instead of doing our own thing here? Why did we decide against reusing sth existing?

We did talk about it. We decided to base the config on the fairseq config system.

> I think also @michelwi wasn't a big fan of this either. 

Yes i am very much afraid that this leads to silent bugs where a user thinks they have sweet a config option but in fact it has been dropped or misinterpreted.

Atticus1806: I can see that. Also I could not really find an easy to read example, but then I never worked with Fairseq and so maybe I just dont know where to look. But I will remove it for now then.

albertz: > We decided to base the config on the fairseq config system.

Ok, basing it on this, but not really using it directly. I really mean to use sth existing, not maintaining our own code for this.

But I also mean other independent packages which are just for this purpose.

What alternatives did you consider? It would be good to list alternatives also in the docstring, to understand later why some decision was made.

PR #4, opened 2023-05-04
Implement conformer convolution part

mmz33:


mmz33:
I think there is an issue with even kernel size. I am testing this. I will add tests later.

Atticus1806 (changes requested):
I think, we decided for this to be an `assembly`, so the it should be in the corresponding folder. 
Furthermore we wanted the kind of Normalization in the middle (So the batchnorm in your case) to be configureable. 

I really like, that you put all the shapes for each operation for the first half of the forward, could you also do that for the second half? Even if they would not change I feel like seeing shapes for depthwise and pointwise conv is helpful.

Good job on the tests!

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
+from torch import nn
+
+
+class ConformerConvolutionV1(nn.Module):
+    """
+    Conformer convolution module.
+    see also: https://github.com/espnet/espnet/blob/713e784c0815ebba2053131307db5f00af5159ea/espnet/nets/pytorch_backend/conformer/convolution.py#L13
+    """
+
+    def __init__(self, channels, kernel_size, dropout=0.1):
```
albertz, line 10: I think we should use proper type annotations.
```suggestion
    def __init__(self, channels: int, kernel_size: int, dropout: float = 0.1):
```

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        :param int channels: number of channels for conv layers
+        :param int kernel_size: kernel size of conv layers
```
albertz, line 13: Types can be removed here.
```suggestion
        :param channels: number of channels for conv layers
        :param kernel_size: kernel size of conv layers
```

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
...
+        """
```
albertz, line 14: `dropout` not documented.

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
...
+        super().__init__()
+
+        # kernel size has to be odd to get same input length when using odd strides. here we use stride 1.
```
albertz, line 17: Why is that? Even with even kernel sizes, you can get the same input length.

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
...
+        self.pointwise_conv2 = nn.Conv1d(
+            in_channels=channels,
+            out_channels=channels,
+            kernel_size=1,
+        )
+        self.layer_norm = nn.LayerNorm(channels)
+        self.batch_norm = nn.BatchNorm1d(channels)
+        self.dropout = nn.Dropout(dropout)
+
+    def forward(self, tensor):
```
albertz, line 41: Typing

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
+from torch import nn
```
albertz, line 1:
```suggestion
from __future__ import annotations
from torch import nn
```

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class ConformerConvolutionV1(nn.Module):
```
albertz, line 4: Do we really want to put `V1` postfix on every single module now?

I think it makes more sense to introduce versions only once we really have multiple versions. For many of our modules, we might never get to that point.

tests/test_conformer.py
```diff
@@ -0,0 +1,16 @@
+from i6_models.parts.conformer.convolution import ConformerConvolutionV1
+import torch
+
+
+def test_conformer_convolution():
```
albertz, line 5:
```suggestion
def test_conformer_convolution_output_shapes():
```

i6_models/parts/conformer/convolution.py
albertz, line 4: In RETURNN-common, you called this `ConformerConvBlock`, not `ConformerConvolution`. Why this name change?

i6_models/parts/conformer/convolution.py
albertz, line 10: In RETURNN-common, dropout was not part of this module, but rather done inside the `ConformerEncoderLayer`. Why this change?

albertz, line 10: In RETURNN-common, we had `norm` as an additional (required) option here, to make it configurable.

There was no default for `norm`. Rather, the default was handled via the outer `ConformerEncoderLayer`. The default was `BatchNorm`.

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        :param int channels: number of channels for conv layers
+        :param int kernel_size: kernel size of conv layers
+        """
+        super().__init__()
+
+        # kernel size has to be odd to get same input length when using odd strides. here we use stride 1.
+        assert (kernel_size - 1) % 2 == 0
+
+        self.pointwise_conv1 = nn.Conv1d(
```
albertz, line 20: Why to use `Conv1d` for this and not `Linear`?

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
...
+            in_channels=channels,
+            out_channels=2 * channels,
+            kernel_size=1,
+        )
+        self.depthwise_conv = nn.Conv1d(
+            in_channels=channels,
+            out_channels=channels,
+            kernel_size=kernel_size,
+            padding=(kernel_size - 1) // 2,
```
albertz, line 29: Why not just `padding = "same"`? That might also solve the problem from the comment above about odd kernel sizes?

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
...
+        self.layer_norm = nn.LayerNorm(channels)
+        self.batch_norm = nn.BatchNorm1d(channels)
+        self.dropout = nn.Dropout(dropout)
+
+    def forward(self, tensor):
+        """
+        :param torch.Tensor tensor: input tensor of shape [B,T,F]
+        :return: torch.Tensor of shape [B,T,F]
+        """
+        out_tensor = self.layer_norm(tensor)
```
albertz, line 46: I don't like that you already call this `out_tensor` here.

I would either just stick to always calling it `tensor`, or use more meaningful names.

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
...
+
+        # conv layers expect shape [B,F,T] so we have to transpose here
+        out_tensor = out_tensor.transpose(1, 2)  # [B,F,T]
+
+        out_tensor = self.pointwise_conv1(out_tensor)  # [B,2F,T]
+        out_tensor = nn.functional.glu(out_tensor, dim=1)  # [B,F,T]
+
+        out_tensor = self.depthwise_conv(out_tensor)
+        out_tensor = nn.functional.silu(self.batch_norm(out_tensor), inplace=True)  # apply swish activation
```
albertz, line 55: I would probably split this up into two separate calls.

i6_models/parts/conformer/convolution.py
albertz: Ah, you need the input for residual connection. Ok then.

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
...
+            kernel_size=kernel_size,
+            padding=(kernel_size - 1) // 2,
+            groups=channels,
+        )
+        self.pointwise_conv2 = nn.Conv1d(
+            in_channels=channels,
+            out_channels=channels,
+            kernel_size=1,
+        )
+        self.layer_norm = nn.LayerNorm(channels)
```
albertz, line 37: In RETURNN-common, layer-norm was not part of this module, but rather it was done in the outer module, in `ConformerEncoderLayer`. Why this change?

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
...
+        self.batch_norm = nn.BatchNorm1d(channels)
```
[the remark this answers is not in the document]
albertz:
```suggestion
        self.norm = norm
```

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
+from torch import nn
+
+
+class ConformerConvolutionV1(nn.Module):
+    """
+    Conformer convolution module.
+    see also: https://github.com/espnet/espnet/blob/713e784c0815ebba2053131307db5f00af5159ea/espnet/nets/pytorch_backend/conformer/convolution.py#L13
+    """
+
+    def __init__(self, channels: int, kernel_size: int, dropout: float = 0.1):
```
JackTemaki, line 10: We better do not add any defaults, this only causes unknown hidden behavior.

i6_models/parts/conformer/convolution.py
JackTemaki: Yes, this was decided. This makes it more obvious that versioning is done per class individually, and we have a consistent pattern.

i6_models/parts/conformer/convolution.py
JackTemaki: This is more efficient as you do not need to transpose again, also it seems that logically people treat this as convolution, even if it is basically a simple linear operation.

JackTemaki (changes requested):

i6_models/parts/conformer/convolution.py
albertz: To clarify: For every single module, we want to add this?

i6_models/parts/conformer/convolution.py
albertz: We are transposing anyway. Only the place of transpose changes.

I would also assume that `Linear` is faster than `Conv1d`.

i6_models/parts/conformer/convolution.py
JackTemaki: @curufinwe 

I understood yes.

mmz33:
> I think, we decided for this to be an `assembly`, so the it should be in the corresponding folder.
> Furthermore we wanted the kind of Normalization in the middle (So the batchnorm in your case) to be configureable.

@Atticus1806 No this should be in `parts`. We agreed that V1 of convolution module use by default pytorch BN.

i6_models/parts/conformer/convolution.py
mmz33: When using even kernel sizes, i think this would work when using padding `kernel_size / 2` but again you have issues with odd kernel sizes. I am not sure i tried to test this and i had issues with that.

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,59 @@
...
+
+        out_tensor = self.pointwise_conv1(out_tensor)  # [B,2F,T]
+        out_tensor = nn.functional.glu(out_tensor, dim=1)  # [B,F,T]
+
+        out_tensor = self.depthwise_conv(out_tensor)
+        out_tensor = nn.functional.silu(self.batch_norm(out_tensor), inplace=True)  # apply swish activation
+
+        out_tensor = self.pointwise_conv2(out_tensor)
+
+        return tensor + self.dropout(out_tensor.transpose(1, 2))
```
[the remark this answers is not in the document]
albertz: For reference: https://github.com/rwth-i6/returnn_common/pull/58

Although this was only the starting point. Since then, it was also further developed.

i6_models/parts/conformer/convolution.py
mmz33: I tried that and it did not work. I also get this warning msg:
```
Warning from pytorch:
Using padding='same' with even kernel lengths and odd dilation may require a zero-padded copy of
the input be created.
```

i6_models/parts/conformer/convolution.py
mmz33: espnet also has this check: https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/conformer/convolution.py#L26

albertz: With `padding = "same"`, you would add `(kernel_size - 1) // 2` padding to the left, and `ceildiv(kernel_size - 1, 2)` padding to the right.

JackTemaki: With even kernel sizes you can not do symmetric padding, this is why there is the `padding: "same"` option that handles this automatically. But I would also be fine with disallowing even Kernel sizes, as this is usually something not so nice to do when you want to keep the sequence length.

i6_models/parts/conformer/convolution.py
mmz33: yes i also understood the same.

i6_models/parts/conformer/convolution.py
albertz: This is just a warning about performance. It says that the performance might be suboptimal in this case, as it requires an additional copy. It should still work though.

i6_models/parts/conformer/convolution.py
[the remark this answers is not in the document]
mmz33: Yeah i would also put residual connection not here. I will fix this.

i6_models/parts/conformer/convolution.py
albertz: Ok I'm not sure this is not overkill. I would have sticked to not adding `V1` to anything, and only add `V2` once we really get to the point that we add a new incompatible version. Which should really be a rare case.

Atticus1806:
> We agreed that V1 of convolution module use by default pytorch BN.

I see, then I guess this is an oversight in the meeting notes, there it still says assembly.

i6_models/parts/conformer/convolution.py
albertz: Using `padding="same"` is just the same for odd kernel sizes as what we have now, and additionally allows even kernel sizes, even if they are maybe not so nice. I don't really understand why we artificially want to have such restriction. Esp if you want to do some hyper param search over kernel sizes.

tests/test_conformer.py
```diff
@@ -0,0 +1,16 @@
...
+        x = torch.randn(batch, time, features)
+        conformer_conv_part = ConformerConvolutionV1(channels=features, kernel_size=kernel_size, dropout=dropout)
+        y = conformer_conv_part(x)
+        return y.shape
+
+    assert get_output_shape(1, 50, 100) == (1, 50, 100)  # test with batch size 1
+    assert get_output_shape(10, 50, 250) == (10, 50, 250)
+    assert get_output_shape(10, 1, 50) == (10, 1, 50)  # time dim 1
+    assert get_output_shape(10, 10, 20, dropout=0.0) == (10, 10, 20)  # dropout 0
+    assert get_output_shape(10, 10, 20, kernel_size=3) == (10, 10, 20)  # kernel size 3
```
SimBe195, line 16: Should we test only the shapes or also the values using a very small example tensor?

i6_models/parts/conformer/convolution.py
mmz33: From a quick search, `Linear` seems to be indeed faster than `Conv1d` as Albert said. So i think better to use to Linear then. I saw that most pytorch implementations use Conv1d and that's why i used this. But we can use `Linear` if it is faster.

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,62 @@
...
+from torch import nn
+
+
+class ConformerConvolutionV1(nn.Module):
+    """
+    Conformer convolution module.
+    see also: https://github.com/espnet/espnet/blob/713e784c0815ebba2053131307db5f00af5159ea/espnet/nets/pytorch_backend/conformer/convolution.py#L13
+    """
+
+    def __init__(self, channels: int, kernel_size: int, dropout: float = 0.1, activation: nn.Module = nn.SiLU()):
```
albertz, line 12: `norm` is missing here as an option.

albertz, line 12: `activation` can also be a `Callable`. Thus the type should be `Union[Callable[[torch.Tensor], torch.Tensor], nn.Module]`. The default should be `nn.functional.silu`, not `nn.SiLU()`.

albertz: Either no default here, and default via the outer module. Or default to batch norm. But not like `norm=nn.BatchNorm()`, because you don't want to reuse the same shared instance.

albertz, line 12: Do we really want to have `dropout = 0.1` as default?

albertz, line 12: Actually, `dropout` should be completely removed here, right? And then you apply dropout in the outer module, just like applying the residual connection.

albertz (changes requested):
Some more comments. But also see the previous non-resolved comments.

i6_models/parts/conformer/convolution.py
JackTemaki: Why would you prefer that?

JackTemaki: In my opinion there should be no defaults.

i6_models/parts/conformer/convolution.py
JackTemaki: If padding="same" is really not supported in ONNX, as noted in `https://stackoverflow.com/questions/68565147/pytorch-unable-to-export-trained-model-as-onnx` we should use the integer version here.

JackTemaki:
In my understanding we discussed to have a straight-forward, non-configurable (e.g. hard-coded BatchNorm and SiLU/Swish) standard Conformer implementation that can be used as starting point and implementation reference, and the main goal was to get people involved and start the process of contributing to modules, not to have a "perfect" (or returnn-common compatible) configurable Conformer implementation, but just a basic initial one. Thus the direct introduction of versioning like V1, because it is clear we would re-work it later.

albertz:
> In my understanding we discussed to have a straight-forward, non-configurable (e.g. hard-coded BatchNorm and SiLU/Swish) standard Conformer implementation that can be used as starting point and implementation reference, and the main goal was to get people involved and start the process of contributing to modules, not to have a "perfect" (or returnn-common compatible) configurable Conformer implementation, but just a basic initial one. Thus the direct introduction of versioning like V1, because it is clear we would re-work it later.

But where do you draw the line of non-configurable? Dropout and kernel size hardcoded or configurable? Num layers and dim configurable? But then why not the type of activation function or type of normalization configurable?

I would say that it should at least cover the standard variants of the Conformer we use in our current best setups. This for example definitely means that activation function and type of norm must be configurable, as we use for example squared-relu in some of the best setups, and layer-norm instead of batch-norm as well in some cases. If this is not configurable, we could not even reproduce some existing setups. I'm not sure if it would really be so useful then.

I also don't really see such a big problem in having this configurable. It doesn't really add any complexity. It just adds one argument to `__init__`, nothing else.

I don't think it make sense to have a V1 which is not useful and would not be used as-is. I would have expected that the purpose of this Conformer implementation is that it would actually be used. So it covers the standard cases we have.

I also don't really see the point in deliberately making it incompatible to our existing implementations (RETURNN-common, RETURNN-frontend, or old net-dict based).

Btw, we actually should maybe check this: Given that we keep type of activation and type of norm configurable, are there other things missing to cover our current best Conformer setups? This is a question to anyone having good Conformer setups, e.g. @christophmluscher, @ZhouW321, @mmz33.

i6_models/parts/conformer/convolution.py
albertz: Because that makes it more flexible and cleaner. E.g. in Transformer, it is common that people play around with the order of layer-norm, dropout and residual connection. I have seen people also playing around with this for Conformer. It is always better and cleaner to keep it minimal and not have this hard-coded at this place.

If you want to have your own custom `ConformerEncoderLayer`, or we have some `ConformerEncoderLayerV2`, it can probably still reuse `ConformerConvolutionV1`, as it becomes more unlikely that any further changes are needed.

It makes it more composable.

Atticus1806:
Hijacking this to ask a general question regarding all 3 PRs: 
So we agreed to change to `ModelConfigurations` as input? Meaning that I can expect 3 configuration classes to exist for the `ConformerBlockV1`?

JackTemaki:
> Hijacking this to ask a general question regarding all 3 PRs: So we agreed to change to `ModelConfigurations` as input? Meaning that I can expect 3 configuration classes to exist for the `ConformerBlockV1`?

Yes

tests/test_conformer.py
mmz33: I think this is difficult to do.

i6_models/config.py
```diff
@@ -29,7 +29,23 @@ class ModelConfiguration:
     def _validate_types(self) -> None:
         for field in fields(type(self)):
             try:
-                typeguard.check_type(getattr(self, field.name), field.type)
+                if isinstance(field.type, str):
```
Atticus1806, line 32: Hmm, why is this marked new? This is in main right?

albertz: Better rebase it.

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,68 @@
...
+
+import torch
+from torch import nn
+from i6_models.config import ModelConfiguration
+from typing import Callable, Union, Any, Type
+
+
+@dataclass
+class ConformerConvolutionV1Config(ModelConfiguration):
+    channels: int
```
Atticus1806, line 13: Please add documentation to the parameters, similar to #6

Also lets talk about the defaults for activation and dropout in the meeting in 15 minutes.

mmz33: why i need to add documentation here. it is already documented in the class constructor. this is only used for `ConformerConvolutionV1` class. so this would be a duplicate.

Atticus1806: Is it? I think you removed it when changing to the config format.
Also I think it makes sense to have this here so when building the config e.g. PyCharm shows the doc.

mmz33: oh it was removed yes.

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,80 @@
...
+        :return: torch.Tensor of shape [B,T,F]
+        """
+        tensor = self.layer_norm(tensor)
+        tensor = self.pointwise_conv1(tensor)  # [B,T,2F]
+        tensor = nn.functional.glu(tensor, dim=-1)  # [B,T,F]
+
+        # conv layers expect shape [B,F,T] so we have to transpose here
+        tensor = tensor.transpose(1, 2)  # [B,F,T]
+        tensor = self.depthwise_conv(tensor)
+        if isinstance(self.norm, nn.LayerNorm):
```
albertz, line 69: There should be no such check on the type.

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,80 @@
...
+class ConformerConvolutionV1Config(ModelConfiguration):
+    channels: int
+    """number of channels for conv layers"""
+    kernel_size: int
+    """kernel size of conv layers"""
+    dropout: float
+    """dropout probability"""
+    activation: Union[nn.Module, Callable[[torch.Tensor], torch.Tensor]]
+    """activation function applied after norm"""
+    norm: Union[nn.BatchNorm1d, nn.LayerNorm]
```
albertz, line 21: Why only those two types? Why can't you accept any module here? And maybe even `Callable`?

tests/test_conformer.py
```diff
@@ -0,0 +1,25 @@
+from i6_models.parts.conformer.convolution import ConformerConvolutionV1, ConformerConvolutionV1Config
+import torch
+import torch.nn as nn
+
+
+def test_conformer_convolution_output_shape():
+    def get_output_shape(
+        batch, time, features, norm=nn.BatchNorm1d, kernel_size=31, dropout=0.1, activation=nn.functional.silu
```
albertz: `norm` should not be the class but an instance.

i6_models/parts/conformer/convolution.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        :param model_cfg: model configuration for this module
+        """
+        super().__init__()
+
+        channels = model_cfg.channels
+        kernel_size = model_cfg.kernel_size
+        dropout = model_cfg.dropout
+        activation = model_cfg.activation
+        norm = model_cfg.norm
```
albertz, line 42: I would not create these locals vars here. Why not directly access `model_cfg` in the code below? That looks simpler to me.

i6_models/parts/conformer/norm.py
```diff
@@ -0,0 +1,28 @@
+import torch
+import torch.nn as nn
+
+
+class LayerNorm(nn.Module):
```
albertz, line 6: You should derive from `nn.LayerNorm`, to have it compatible.

i6_models/parts/conformer/norm.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    Generic LayerNorm that accepts any input shape
+    see here: https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html
+    """
+
+    def __init__(self, normalized_shape):
+        """
+        :param normalized_shape: shape for normalization
```
albertz, line 14: I don't understand this. What is this for?

i6_models/parts/conformer/norm.py
```diff
@@ -0,0 +1,28 @@
...
+
+        self.normalized_shape = normalized_shape
+        self.layer_norm = nn.LayerNorm(normalized_shape)
+
+    def forward(self, tensor: torch.Tensor) -> torch.Tensor:
+        """
+        :param tensor: input tensor of any shape
+        """
+        shape = tensor.shape
+        tensor = tensor.reshape(-1, self.normalized_shape)
```
albertz, line 26: Why are you reshaping? I thought the purpose is to transpose, not to reshape?

i6_models/parts/conformer/norm.py
```diff
@@ -0,0 +1,28 @@
+import torch
+import torch.nn as nn
+
+
+class LayerNorm(nn.Module):
+    """
+    Generic LayerNorm that accepts any input shape
+    see here: https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html
```
albertz, line 9: I don't exactly understand what you mean.

I thought this layer norm was just to accept the (N, C, *) dim order, instead of (N, *, C) dim order. So it is not really about accepting any shape, but just this specific different dim order.

mmz33: ok then i made it very generic. i did not know that we restrict it to 3 tensor shape. anw this is simpler for me. i will just do this.

albertz: I'm not saying to restrict it to 3 dims. I just say, this is about the dim order, not the shape.

mmz33: ok i think i misunderstood it. i will fix this.

i6_models/parts/conformer/convolution.py
mmz33: I think this is fine. I would prefer to see the list of model configurations in the constructor.

i6_models/parts/conformer/norm.py
albertz: `LayerNormNC` or `LayerNormBC` or `LayerNormBatchFeatMajor` or so. Or simply keep `LayerNorm` and just explain it in the doc.

Btw, ESPnet has exactly this [here](https://espnet.github.io/espnet/_modules/espnet/nets/pytorch_backend/transformer/layer_norm.html).

mmz33: ah i see. should we follow then the implementation from ESPNet?

mmz33: I made it simpler.

albertz: Yep looks good.

i6_models/parts/conformer/convolution.py
mmz33: ok i made it simpler.

Atticus1806 (approved):
Nice!

i6_models/parts/conformer/norm.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    LayerNorm that accepts [N,C,*] tensors and normalizes over C (channels) dimension.
+    see here: https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html
+    """
+
+    def __init__(self, channels: int):
+        """
+        :param channels: number of channels for normalization
+        """
+        super(LayerNormNC, self).__init__(channels)
```
SimBe195, line 15:
```suggestion
        super().__init__(channels)
```

i6_models/parts/conformer/norm.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def forward(self, tensor: torch.Tensor) -> torch.Tensor:
+        """
+        :param tensor: input tensor with shape [N,C,*]
+        :return: normalized tensor with shape [N,C,*]
+        """
+        return super(LayerNormNC, self).forward(tensor.transpose(1, -1)).transpose(1, -1)
```
SimBe195, line 22:
```suggestion
        return super().forward(tensor.transpose(1, -1)).transpose(1, -1)
```

curufinwe (approved):

PR #6, opened 2023-05-05
Implement ConformerFeedForwardV1 Part

Judyxujj:


i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,45 @@
+from torch import nn
```
albertz, line 1:
```suggestion
from __future__ import annotations
import torch
from torch import nn
```

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class ConformerFeedForwardv1(nn.Module):
```
albertz, line 4:
```suggestion
class ConformerFeedForwardV1(nn.Module):
```

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    Conformer feedforward module
+    """
+
+    def __init__(self, input_dim, hidden_dim, dropout=0.1):
```
albertz, line 9: Typing

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        :param int input_dim: input dimension
+        :param int hidden_dim: hidden dimension (normally set to 4*input_dim as suggested by the paper)
+        :param float dropout: dropout probability
+        """
+        super().__init__()
+
+        self.layer_norm = nn.LayerNorm(normalized_shape=input_dim)
```
albertz, line 17: In our earlier implementation in RETURNN-common, we decided to have layer-norm not as part of this module, but rather as part of the outer `ConformerEncoderLayer`.

i6_models/parts/conformer/feedforward.py
albertz, line 4: In our earlier implementation in RETURNN-common, we decided to call this `ConformerPositionwiseFeedForward`. Why this name change?

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,45 @@
...
+
+        self.linear_layer_1 = nn.Linear(
```
albertz, line 19: We earlier called this `linear_ff`. Why this name change? I think it would help to keep it compatible.

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,45 @@
...
+            in_features=input_dim, out_features=hidden_dim, bias=True
+        )
+
+        self.swish_activation = nn.SiLU()
+
+        self.linear_layer_2 = nn.Linear(
```
albertz, line 25: Same here. Call it `linear_out`.

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        super().__init__()
+
+        self.layer_norm = nn.LayerNorm(normalized_shape=input_dim)
+
+        self.linear_layer_1 = nn.Linear(
+            in_features=input_dim, out_features=hidden_dim, bias=True
+        )
+
+        self.swish_activation = nn.SiLU()
```
albertz, line 23: This was configurable in our earlier implementation. So I would just call it `activation`. And I would make it configurable.

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,45 @@
...
+
+        self.linear_layer_2 = nn.Linear(
+            in_features=hidden_dim, out_features=input_dim, bias=True
+        )
+
+        self.dropout = nn.Dropout(p=dropout)
+
+    def forward(self, tensor):
```
albertz, line 31: Typing

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,45 @@
...
+        out_tensor = self.layer_norm(tensor)  # [B,T,F]
+
+        out_tensor = self.linear_layer_1(out_tensor)  # [B,T,F]
+        out_tensor = self.swish_activation(out_tensor)  # [B,T,F]
+
+        out_tensor = self.linear_layer_2(out_tensor)  # [B,T,F]
+
+        out_tensor = self.dropout(out_tensor)  # [B,T,F]
+
+        return tensor
```
albertz, line 45: So here you do not have a residual connection, which is consistent to our earlier implementation.

However, we should have it in any case consistent here. So it depends what we decide for `ConformerConvolutionV1` (or `ConformerConvBlock`) [here](https://github.com/rwth-i6/i6_models/pull/4/files#r1185864931).

JackTemaki:
@albertz please stop adding each comment as itself, and collect them in one review and send them at once.

albertz:
> @albertz please stop adding each comment as itself, and collect them in one review and send them at once.

I think that the way I do it is easier, as you can work already as you go. At least, I would prefer it that way, both as the reviewer, and also as the developer.

vieting:
> > @albertz please stop adding each comment as itself, and collect them in one review and send them at once.
> 
> I think that the way I do it is easier, as you can work already as you go. At least, I would prefer it that way, both as the reviewer, and also as the developer.

A plus for Nick's approach is that other subscribers don't get spammed with mails for every comment.

i6_models/parts/conformer/feedforward.py
albertz: `linear_ff_1` is still not the same as `linear_ff`.

i6_models/parts/conformer/feedforward.py
albertz: `linear_ff_2` is still not the same as `linear_out`.

i6_models/parts/conformer/feedforward.py
Judyxujj: The name used here is decided in the last implementation meeting

albertz: So why was it decided that way? Why the name change?

Judyxujj: I think most of the people including me are not aware of the name ConformerPositionwiseFeedForward is used in RETURNN-common. If we want to use the consistent name as in RETURNN-common, I can change the name here

albertz: @mmz33 was working on that, and then me as well. We had lots of earlier discussions on all of this, like naming of modules, etc. So I wonder a bit why we ignore all of that now.

albertz: For reference: https://github.com/rwth-i6/returnn_common/pull/58

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,40 @@
...
+
+        self.linear_out = nn.Linear(in_features=hidden_dim, out_features=input_dim, bias=True)
+
+        self.dropout = nn.Dropout(p=dropout)
+
+    def forward(self, tensor: torch.Tensor):
+        """
+        :param torch.Tensor tensor: input tensor of shape [B,T,F]
+        :return: torch.Tensor of shape [B,T,F]
+        """
```
albertz, line 31: Typing still misses the return value.
Typing in the docstring is also not necessary and should be avoided.
I would also add that `F` refers to the `input_dim`.
```suggestion
    def forward(self, tensor: torch.Tensor) -> torch.Tensor:
        """
        :param tensor: shape [B,T,F], F=input_dim
        :return: shape [B,T,F], F=input_dim
        """
```

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,40 @@
...
+        super().__init__()
+
+        self.linear_ff = nn.Linear(in_features=input_dim, out_features=hidden_dim, bias=True)
+
+        self.swish_activation = nn.SiLU()
+
+        self.linear_out = nn.Linear(in_features=hidden_dim, out_features=input_dim, bias=True)
+
+        self.dropout = nn.Dropout(p=dropout)
```
albertz, line 25: Similar as in the `ConformerConvolutionV1` (see [here](https://github.com/rwth-i6/i6_models/pull/4#discussion_r1185852351)), we probably do not want dropout here, but rather in the outer `ConformerEncoderLayer` module.

i6_models/parts/conformer/feedforward.py
Judyxujj: Thanks for the reference, I have now also made activation configurable as in RETURNN-common

albertz: I would have expected that we stick to our earlier discussions and decisions, and not redo all of those discussions, which are all not really so important anyway.

And introducing inconsistencies to our other existing implementation for no good reason is also not nice.

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,42 @@
...
+        :param input_dim: input dimension
+        :param hidden_dim: hidden dimension (normally set to 4*input_dim as suggested by the paper)
+        :param dropout: dropout probability
+        :param activation: activation function
+        """
+        super().__init__()
+
+        self.linear_ff = nn.Linear(in_features=input_dim, out_features=hidden_dim, bias=True)
+
+        self.activation = activation()
```
albertz, line 23:
```suggestion
        self.activation = activation
```

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,42 @@
...
+import torch
+from torch import nn
+
+
+class ConformerPositionwiseFeedForward(nn.Module):
+    """
+    Conformer feedforward module
+    """
+
+    def __init__(self, input_dim: int, hidden_dim: int, dropout: float = 0.1, activation: Callable[[torch.Tensor], torch.Tensor] = nn.SiLU):
```
albertz:
```suggestion
    def __init__(self, input_dim: int, hidden_dim: int, dropout: float = 0.1, activation: Callable[[torch.Tensor], torch.Tensor] = nn.functional.silu):
```

albertz: Same comment for default `dropout` value here.
But then, as written in another comment, `dropout` should better not be part of this module.

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import annotations
+from typing import Callable
+import torch
+from torch import nn
+
+
+class ConformerPositionwiseFeedForward(nn.Module):
```
albertz, line 7: So, do we need `V1` here now or not?

albertz (changes requested):

albertz:
Btw, for Black, you can configure PyCharm to automatically apply it when you save. See [here](https://github.com/rwth-i6/returnn/wiki/PyCharm-Configuration).

i6_models/parts/conformer/feedforward.py
JackTemaki: @albertz in earlier discussions and decisions nearly no one took part. We specifically started over having as much people involved as possible. I definitely see it as a problem that you can not join the Tuesday meeting, and we should find a solution for that, e.g. have another date.

Also if the discussions are not important, then I do not understand the large number of comments from you here.

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Conformer feedforward module
+    """
+    
+    def __init__(
+        self,
+        input_dim: int,
+        hidden_dim: int,
+        dropout: float = 0.1,
+        activation: Callable[[torch.Tensor], torch.Tensor] = nn.SiLU,
```
Atticus1806, line 17: Is `SiLU` the same default as used in the paper? Or is this already an improvement everyone is using?

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def forward(self, tensor: torch.Tensor):
+        """
+        :param torch.Tensor tensor: input tensor of shape [B,T,F]
+        :return: torch.Tensor of shape [B,T,F]
+        """
+        out_tensor = self.linear_ff(tensor)  # [B,T,F]
+
+        out_tensor = self.activation(out_tensor)  # [B,T,F]
+
```
Atticus1806, line 43: If I am not mistaken there is dropout between activation and linear layer in the original paper.

i6_models/parts/conformer/feedforward.py
albertz: > Also if the discussions are not important, then I do not understand the large number of comments from you here.

It should not be inconsistent to the existing implementation.

Also, we already had some experience with the earlier decisions, and some things were changed later on. So now re-deciding that again first takes takes time, and then also ignores that earlier experience.

So, I would stick to the original decisions, and be consistent to the existing implementation, unless there are really good reasons to change some parts.

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,48 @@
...
+        self.linear_out = nn.Linear(in_features=hidden_dim, out_features=input_dim, bias=True)
+
+        self.dropout = nn.Dropout(p=dropout)
+
+    def forward(self, tensor: torch.Tensor) -> torch.Tensor:
+        """
+        :param tensor: shape [B,T,F], F=input_dim
+        :return: shape [B,T,F], F=input_dim
+        """
```
Atticus1806, line 39: I think there should be `layernorm` before the first linear here.

Atticus1806 (changes requested):

i6_models/parts/conformer/feedforward.py
Judyxujj: yes, but in RETURNN-common, the layer normalisation is applied in the outer ConformerBlock instead of in the Feedforward Module. So Albert suggested we do in the consistent way as in RETURNN-common

i6_models/parts/conformer/feedforward.py
Judyxujj: Yes, the nn.SiLU activation is swish activation. See [here](https://pytorch.org/docs/stable/generated/torch.nn.SiLU.html)

i6_models/parts/conformer/feedforward.py
Judyxujj: Nice catch! I overlooked this dropout, I have added one more dropout now.

i6_models/parts/conformer/feedforward.py
Atticus1806: Ah, I did not know that, thanks!

i6_models/parts/conformer/feedforward.py
albertz: In the outer `ConformerEncoderLayer`. Just like the final dropout and the residual connection. So this here is as minimal as possible. We discussed this some time ago with @mmz33.

SimBe195:
Some tests should also be added, similar to #4.

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,50 @@
...
+
+        out_tensor = self.activation(out_tensor)  # [B,T,F]
+
+        out_tensor = self.dropout(out_tensor)  # [B,T,F]
+
+        out_tensor = self.linear_out(out_tensor)  # [B,T,F]
+
+        out_tensor = self.dropout(out_tensor)  # [B,T,F]
+
+        return tensor
```
Atticus1806, line 50:
```suggestion
        return out_tensor
```

Judyxujj:
> Some tests should also be added, similar to #4.

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class ConformerPositionwiseFeedForwardV1(nn.Module):
+    """
+    Conformer feedforward module
+    """
+
+    def __init__(
+        self,
+        input_dim: int,
```
Atticus1806, line 14: Please change the input to take some kind of ConformerFeedForwardConfig (c.f. #8 for inspiration) and also define that Config in this file, as agreed on in the last meeting.

Atticus1806 (changes requested):

i6_models/parts/conformer/feedforward.py
Judyxujj: The inputs are defined as Config class now

Atticus1806 (approved):
Thanks a lot for the extensive work, looks good!

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+        :param tensor: shape [B,T,F], F=input_dim
+        :return: shape [B,T,F], F=input_dim
+        """
+        tensor = self.layer_norm(tensor)
+        tensor = self.linear_ff(tensor)  # [B,T,F]
+        tensor = self.activation(tensor)  # [B,T,F]
+        tensor = self.dropout(tensor)  # [B,T,F]
+        tensor = self.linear_out(tensor)  # [B,T,F]
+        tensor = self.dropout(tensor)  # [B,T,F]
```
michelwi, line 47: if the same `self.dropout` is used after `linear_out` and before, will this make a difference as to using two different dropouts or does this not matter? 
(e.g. will always the same dimensions be dropped)

Atticus1806: Good point. Removed my approval for now since this is quite important (also for the other PRs) to clarify.

Judyxujj: I have made a small test to verify that the second dropout would randomly mask out channel based on the result of the first dropout (two dropouts work independently)
```
>>> dropout = torch.nn.Dropout(p=0.5)
>>> output = dropout(output)
>>> output = dropout(input)
>>> output
tensor([[0., 2., 0., 0., 0., 2., 0., 2., 2., 0.]])
>>> output = dropout(output)
>>> output
tensor([[0., 0., 0., 0., 0., 0., 0., 4., 4., 0.]])
```

albertz: Just look at the code ([here](https://pytorch.org/docs/stable/_modules/torch/nn/modules/dropout.html#Dropout)).

But to make the code more explicit, it might anyway be a good idea to use the functional variant instead: [`torch.nn.functional.dropout`](https://pytorch.org/docs/stable/generated/torch.nn.functional.dropout.html)

Atticus1806 (approved):

i6_models/parts/conformer/feedforward.py
albertz: In any case, I would avoid calling the same module here again. That's very uncommon for PyTorch, unless you explicitly intend to have param sharing.

JackTemaki: > it might anyway be a good idea to use the functional variant instead

Please be aware though that you then need to pass `training` explicitly and Dropout will no longer automatically be only enabled in training.

albertz: Just `tensor = torch.nn.functional.dropout(tensor, p=self.dropout, training=self.training)` (assuming `self.dropout` is now the float for the dropout prob).

michelwi (approved):

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,48 @@
...
+
+@dataclass
+class ConformerPositionwiseFeedForwardV1Config(ModelConfiguration):
+    input_dim: int
+    """input dimension"""
+    hidden_dim: int
+    """hidden dimension (normally set to 4*input_dim as suggested by the paper)"""
+    dropout: float
+    """dropout probability"""
+    activation: Callable[[torch.Tensor], torch.Tensor]
```
curufinwe, line 19: This should default to `torch.nn.functional.silu`

i6_models/parts/conformer/feedforward.py
```diff
@@ -0,0 +1,48 @@
...
+from i6_models.config import ModelConfiguration
+
+
+@dataclass
+class ConformerPositionwiseFeedForwardV1Config(ModelConfiguration):
+    input_dim: int
+    """input dimension"""
+    hidden_dim: int
+    """hidden dimension (normally set to 4*input_dim as suggested by the paper)"""
+    dropout: float
```
curufinwe, line 17: Do we want to have a default value for dropout?

curufinwe: I see that @JackTemaki commented in another PR for not having any defaults, for dropout I think it's OK. for the activation function I think there should be a default, as it's part of the paper and usually not something people "tune" as a hyper-parameter.

Judyxujj: Okay, I have just added nn.functional.silu as the default activation

curufinwe (approved):

PR #5, opened 2023-05-05
Implement ConformerFeedForwardV1 part

Judyxujj:

PR #7, opened 2023-05-05
add MHSA module

kuacakuaca:
pretty much the same as in torchaudio.models.conformer. 
1. Relative positional encoding is missing, i'm not sure if it should be added and if yes, do we implement it by ourselves or just import it from somewhere else. 
2. Should more parameters be added, e.g. key_dim, value_dim?


i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,41 @@
+import torch
+
+class ConformerMHSAV1(nn.Module):
```
michelwi, line 3:
```suggestion
class ConformerMHSAV1(torch.nn.Module):
```
would this work without torch?

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+    def forward(self, input: torch.Tensor):
+
+        residual = input
+
+        # layer norm, Multi-head self attention with dropout and residual connection
+        output = torch.nn.LayerNorm(self.embed_dim)
+        output, _ = self.mhsa(output)
+        output = torch.nn.Dropout(self.dropout)
```
michelwi, line 36:
```suggestion
        output = torch.nn.Dropout(self.dropout)(output)
```
I think this guy would need to know about what should be dropped out. a quick googling suggest this is the right way to tell it

michelwi (changes requested):

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,36 @@
+import torch
```
albertz, line 1:
```suggestion
from __future__ import annotations
import torch
```

albertz (changes requested):
Why do we need this at all?

Again, as in the other PRs (#4, #6), I would keep consistent to our existing implementation (see https://github.com/rwth-i6/returnn_common/pull/58 for the initial code, but then also check the current code, also in the RETURNN frontend, also check with @mmz33). We discussed many things about the naming of modules and also their behavior.

In this specific case, you would use a standard generic `RelPosSelfAttention`/`RelPositionMultiHeadedAttention`. Or the standard PyTorch [`MultiheadAttention`](https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html).

Also, this module here would not add layer-norm, or the residual connection, or dropout. That would all be handled by the outer module (`ConformerEncoderLayer`).

So, rather than reimplementing the already existing `MultiheadAttention`, what is missing is sth like `RelPositionMultiHeadedAttention`. We have that in RETURNN-common / RETURNN-frontend as well already, so I would suggest to just follow that implementation, unless there are good reasons to do sth differently. But I would keep at least the naming, arguments and variable names consistent.

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,36 @@
...
+        dropout: float = 0.1,
+    ):
+        super().__init__()
+
+        self.layernorm = torch.nn.LayerNorm(embed_dim)
+        self.mhsa = torch.nn.MultiheadAttention(embed_dim, num_att_heads, dropout=att_weights_dropout)
+        self.dropout = torch.nn.Dropout(dropout)
+
+    def forward(self, input: torch.Tensor):
+
```
Atticus1806, line 27: Even though some/most of it might be trivial, can you please add shapes and input/output shapes, similar to the forward in the other two PRs?

SimBe195:
Some tests should also be added, similar to https://github.com/rwth-i6/i6_models/pull/4.

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,36 @@
...
+        att_weights_dropout: float = 0.1,
+        dropout: float = 0.1,
+    ):
+        super().__init__()
+
+        self.layernorm = torch.nn.LayerNorm(embed_dim)
+        self.mhsa = torch.nn.MultiheadAttention(embed_dim, num_att_heads, dropout=att_weights_dropout)
+        self.dropout = torch.nn.Dropout(dropout)
+
+    def forward(self, input: torch.Tensor):
```
JackTemaki, line 26: It is required to pass a mask to the forward. A mask is a B x T sized boolean Tensor, where 0 means "within sequence" and 1 means "outside sequence". See https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html#torch.nn.MultiheadAttention.forward

JackTemaki (changes requested):

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import annotations
+from dataclasses import dataclass
+from typing import Optional, Callable
+import torch
+
+from i6_models.config import ModelConfiguration
+
+
+@dataclass
+class ConformerMHSAV1Config(ModelConfiguration):
```
Atticus1806, line 10: Please add documentation as stated in https://github.com/rwth-i6/i6_models/pull/8#discussion_r1192185282 so it shows up in the IDE.

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,42 @@
...
+    input_dim: int  # input dim and total dimension for query/key and value projections, should be dividable by `num_att_heads`
+    num_att_heads: int  # number of attention heads
+    key_padding_mask: Optional[
```
Atticus1806, line 13: I think this has to be input to the `forward` and not part of the Config since this depends on padding/batching, so it must not be static. In [torchaudio](https://pytorch.org/audio/main/generated/torchaudio.models.Conformer.html) they call it `lengths`.

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,42 @@
...
+        torch.Tensor
+    ] = None  # binary or float mask deciding if a key position should ignored or not, of shape (B, T_k)
+    att_weights_dropout: float = 0.1  # attention weights dropout
```
Atticus1806, line 16: Have we decided whether to add defaults for some values like dropout yet? I know there was a discussion, but I didn't follow the results. @JackTemaki @albertz ?

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,42 @@
...
+        self.layernorm = torch.nn.LayerNorm(cfg.input_dim)
+        self.mhsa = torch.nn.MultiheadAttention(
+            cfg.input_dim, cfg.num_att_heads, dropout=cfg.att_weights_dropout, batch_first=True
+        )
+        self.key_padding_mask = cfg.key_padding_mask
+        self.dropout = torch.nn.Dropout(cfg.dropout)
+
+    def forward(self, input: torch.Tensor):
+
+        # layer norm, Multi-head self attention with dropout and residual connection
```
Atticus1806, line 37:
```suggestion
        # layer norm, Multi-head self attention with dropout
```

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,42 @@
...
+from dataclasses import dataclass
+from typing import Optional, Callable
+import torch
+
+from i6_models.config import ModelConfiguration
+
+
+@dataclass
+class ConformerMHSAV1Config(ModelConfiguration):
+    input_dim: int  # input dim and total dimension for query/key and value projections, should be dividable by `num_att_heads`
```
Atticus1806, line 11: In [torch](https://pytorch.org/docs/stable/_modules/torch/nn/modules/activation.html#MultiheadAttention) they hard check for this divisibilty with 
`self.head_dim = embed_dim // num_heads`
`assert self.head_dim * num_heads == self.embed_dim, "embed_dim must be divisible by num_heads"`.
I think it would be nice to have this check in the config already, this should be doable through the `post_init` similar to typing. 
Maybe we could extend this even further and introduce something like `value_check()` which in the base config is just a `pass`, but can be overwritten in each config (if there are checks to make ofc) and then get automatically called by the `__post_init` of the `ModelConfiguration` base class. 

Even though this is resolved now, in the Conformer part this `assert (kernel_size - 1) % 2 == 0`  could have been resolved as well (I know its not required anymore just as an example). This way we can keep the error where the values are actually set and since the config is closely tied to the module this should work in my opinion. 

Any opinions?

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,42 @@
...
+        super().__init__()
+
+        self.layernorm = torch.nn.LayerNorm(cfg.input_dim)
+        self.mhsa = torch.nn.MultiheadAttention(
+            cfg.input_dim, cfg.num_att_heads, dropout=cfg.att_weights_dropout, batch_first=True
+        )
+        self.key_padding_mask = cfg.key_padding_mask
+        self.dropout = torch.nn.Dropout(cfg.dropout)
+
+    def forward(self, input: torch.Tensor):
```
Atticus1806, line 35: I am not sure about the name `input`. So in general this should not be a problem, but it shadows the python function `input`, which is at least not clean in my opinion. But I guess this should not be too big of a deal.

tests/test_conformer.py
```diff
@@ -0,0 +1,26 @@
...
+    input_shape = [3, 10, 20]  # B,T,F
+    cfg = ConformerMHSAV1Config(20, 4)
+    assert get_output_shape(input_shape, cfg) == [3, 10, 20]
+
+    # all parameters
+    input_shape = [4, 15, 32]  # B,T,F
+    cfg = ConformerMHSAV1Config(
+        32, 8, key_padding_mask=torch.randint(0, 2, input_shape[:2]) > 0, att_weights_dropout=0.2, dropout=0.3
+    )
+
```
Atticus1806, line 25:
```suggestion
```

Atticus1806 (changes requested):
Please see comments. Some minor things, but also some more fundamental. 
Regarding the failing test I think you need to add `torch` to the `requirements.txt` since its not done in the main yet.

i6_models/parts/conformer/mhsa.py
kuacakuaca: i think this kind of check will be quite common, so adding this value_check() sounds quite reasonable.

i6_models/parts/conformer/mhsa.py
kuacakuaca: i'll change it to `input_tensor`.

i6_models/parts/conformer/mhsa.py
Atticus1806: C.f. #12

michelwi (commented):
Only one comment left from my side

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,45 @@
...
+from i6_models.config import ModelConfiguration
+
+
+@dataclass
+class ConformerMHSAV1Config(ModelConfiguration):
+    input_dim: int
+    """input dim and total dimension for query/key and value projections, should be dividable by `num_att_heads`"""
+    num_att_heads: int
+    """number of attention heads"""
+    att_weights_dropout: float = 0.1
```
curufinwe, line 15: See discussion about defaults in other PRs: https://github.com/rwth-i6/i6_models/pull/4#discussion_r1189133116 and https://github.com/rwth-i6/i6_models/pull/6#discussion_r1196077265

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,45 @@
...
+        super().__init__()
+
+        self.layernorm = torch.nn.LayerNorm(cfg.input_dim)
+        self.mhsa = torch.nn.MultiheadAttention(
+            cfg.input_dim, cfg.num_att_heads, dropout=cfg.att_weights_dropout, batch_first=True
+        )
+        self.dropout = torch.nn.Dropout(cfg.dropout)
+
+    def forward(self, input_tensor: torch.Tensor, key_padding_mask: Optional[torch.Tensor] = None):
+
```
Atticus1806, line 37: Can you please add a docstring explaining the inputs and the return of this function? Esp. for the mask I think this is important to see what it is for.
C.f. the forward in #4 #5

Atticus1806 (approved):

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,55 @@
...
+        super().__init__()
+
+        self.layernorm = torch.nn.LayerNorm(cfg.input_dim)
+        self.mhsa = torch.nn.MultiheadAttention(
+            cfg.input_dim, cfg.num_att_heads, dropout=cfg.att_weights_dropout, batch_first=True
+        )
+        self.dropout = cfg.dropout
+
+    def forward(self, input_tensor: torch.Tensor, key_padding_mask: Optional[torch.Tensor] = None) -> torch.Tensor:
+
```
michelwi, line 41:
```suggestion
```

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import Optional, Callable
+import torch
+
+from i6_models.config import ModelConfiguration
+
+
+@dataclass
+class ConformerMHSAV1Config(ModelConfiguration):
+    input_dim: int
+    """input dim and total dimension for query/key and value projections, should be dividable by `num_att_heads`"""
```
michelwi, line 12:
```suggestion
    """input dim and total dimension for query/key and value projections, should be divisible by `num_att_heads`"""
```

michelwi (approved):

i6_models/parts/conformer/mhsa.py
```diff
@@ -0,0 +1,55 @@
...
+    num_att_heads: int
+    """number of attention heads"""
+    att_weights_dropout: float
+    """attention weights dropout"""
+    dropout: float
+    """multi-headed self attention output dropout"""
+
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        assert self.input_dim % self.num_att_heads == 0, "input_dim must be dividable by num_att_heads"
```
SimBe195, line 22:
```suggestion
        assert self.input_dim % self.num_att_heads == 0, "input_dim must be divisible by num_att_heads"
```

PR #8, opened 2023-05-11
Conformer Assembly

Atticus1806:
This PR implements ConformerEncoderV1, parts of ConvFrontendV1 and ConformerBlockV1.

For now this is WIP since this is dependent on a few things from the three other PRs. This is why there are the dummy configurations for now, which are going to be replaced by the real configurations once they are in the respective PR. 

What is not implemented yet is SpecAugment as I am not 100% sure how to properly do it. 

EDIT to avoid confusion: The only "new" file compared to the other PRs is assemblies/conformer/conformer.py 
Since the other ones are not merged yet and I had to branch from somewhere unfortunately the other files are marked new too.


i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,179 @@
...
+
+@dataclass
+class ConformerBlockV1Config(ModelConfiguration):
+    # hyperparameters
+    norm_dim: int
+
+    # nested configurations
+    ff_cfg: ConformerFeedForwardV1Config  # Configuration for ConformerPositionwiseFeedForwardV1
+    mhsa_cfg: ConformerMHSAV1Config  # Configuration for ConformerMHSAV1
+    conv_cfg: ConformerConvV1Config  # Configuration for ConformerConvolutionV1
```
vieting, line 39: Why do the names of the configs differ from the names of the classes they are used for (at least according to the comments)?

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+@dataclass
+class ConformerFrontendV1Config(ModelConfiguration):
+
```
vieting, line 44:
```suggestion
```

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,179 @@
...
+        x = self.subsampling(x)  # [B, F, T']
+        x = x.transpose(2, 1)  # [B, T', F]
+        x = self.linear(x)  # [B, T', F']
+        x = self.dropout(x)  # [B, T', F']
+
+        return x
+
+
+class ConformerEncoderV1(nn.Module):
+    def __init__(self, cfg: ConformerV1Config):
```
vieting, line 155:
```suggestion
    def __init__(self, cfg: ConformerEncoderV1Config):
```
Also here, I'd suggest to keep names for config and class the same.

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+@dataclass
+class ConformerMHSAV1Config(ModelConfiguration):
+    pass
+
+
+@dataclass
+class ConformerConvV1Config(ModelConfiguration):
+    pass
```
vieting, line 28: I think it would be nicer to define these in the files where the respective classes are defined rather than here.

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def __init__(self, cfg: ConformerBlockV1Config):
+        """
+        :param cfg: Conformer Block Configuration with subunits for the different Conformer parts
+        """
+        super().__init__()
+
+        self.ff_config = cfg.ff_cfg
+        self.mhsa_config = cfg.mhsa_cfg
+        self.conv_config = cfg.conv_cfg
```
vieting, line 81: Do we really need these as class attributes?

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,179 @@
...
+        self.ff_2 = ConformerPositionwiseFeedForwardV1(cfg=self.ff_config)
+        self.final_layer_norm = torch.nn.LayerNorm(self.ff_config.dim)
+
+    def forward(self, tensor: torch.Tensor):
+        """
+        :param tensor: input tensor of shape [B, T, F]
+        :return: torch.Tensor of shape [B, T, F]
+        """
+        residual = tensor  #  [B, T, F]
+        x = self.ff_1(tensor)  #  [B, T, F]
```
vieting, line 95: The feature dim might change here, right? Should we account for that in the commented shape or do we only care about the order of axes?

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,179 @@
...
+        :param cfg: Conformer Frontend Configuration
+        """
+        super().__init__()
+
+        self.feature_dim = cfg.feature_dim
+        self.hidden_dim = cfg.hidden_dim
+        self.dropout_value = cfg.dropout
+        self.conv_stride = cfg.conv_stride
+        self.conv_kernel = cfg.conv_kernel
+        self.conv_padding = cfg.conv_padding
```
vieting, line 123: Do we really need these as class attributes?

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,179 @@
...
+
+        self.spec_aug = ...  # TODO
```
vieting, line 125: If we decide to go forward with a PR that still has open TODOs (fine for me), it should at least not break the code.

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,179 @@
...
+    def __init__(self, cfg: ConformerV1Config):
+        """
+        :param cfg: Conformer Encoder Configuration with subunits for Frontend and Conformer blocks
+        """
+        super().__init__()
+
+        self.num_layers = cfg.num_layers
+
+        self.front_cfg = cfg.front_cfg
+        self.block_cfg = cfg.block_cfg
```
vieting, line 164: Do we really need these as class attributes?

vieting (changes requested):

i6_models/assemblies/conformer/conformer.py
Atticus1806: I can change that for now, but this will change anyways and depend on the name given in the other PRs.

i6_models/assemblies/conformer/conformer.py
SimBe195: IMO this should be indicated in the docstring, e.g. input `[B, T, F]`, output `[B, T, F']`.

i6_models/assemblies/conformer/conformer.py
Atticus1806: we could also either keep the whole `cfg` as class attribute or none of it. I don't have a strong opinion on this.

i6_models/assemblies/conformer/conformer.py
Atticus1806: According to doc of the FF module it should not change.

i6_models/assemblies/conformer/conformer.py
SimBe195: I would say not to set attributes that are never used anywhere else.

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,179 @@
...
+    norm_dim: int
+
+    # nested configurations
+    ff_cfg: ConformerFeedForwardV1Config  # Configuration for ConformerPositionwiseFeedForwardV1
+    mhsa_cfg: ConformerMHSAV1Config  # Configuration for ConformerMHSAV1
+    conv_cfg: ConformerConvV1Config  # Configuration for ConformerConvolutionV1
+
+
+@dataclass
+class ConformerFrontendV1Config(ModelConfiguration):
```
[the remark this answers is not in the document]
Atticus1806: Since they also hold the inputs this was my way of thinking yes.

i6_models/assemblies/conformer/conformer.py
Atticus1806: so not even `self.cfg` then?

i6_models/assemblies/conformer/conformer.py
Atticus1806: I agree. They should be added in the corresponding PR imho, I am just adding them here as dummies until they exist.

i6_models/assemblies/conformer/conformer.py
Atticus1806: Tbh. I just thought it would be nice to have it available without further thinking, but I will remove it for now and we can add it later if required.

Atticus1806:
Since @JackTemaki and I want to test some stuff today I took the liberty of actually implementing the config classes. 
@mmz33 @Judyxujj @kuacakuaca  feel free to copy them over, but they are by no means meant to be complete code, so if you prefer to define them in another way, please go ahead. This was only a quick write to be able to work with them now.

i6_models/assemblies/conformer/conformer.py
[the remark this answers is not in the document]
albertz: No. You need to format them differently. There currently is not really a well defined standard way, but the most accepted way in the community, which also partly works in PyCharm, is to use attribute docstrings, which is a string right after the attribute. Like:
```python
feature_dim: int
"""Feature dimension of the input data"""
```
See [here](https://youtrack.jetbrains.com/issue/PY-28549) and [here](https://stackoverflow.com/questions/64360563/how-to-auto-generate-docstrings-in-pycharm-for-dataclasses) for some discussion, and [PEP 257](https://peps.python.org/pep-0257/).

Atticus1806:
This PR should be ready now. It relies on #14 which should be merged first. 
Regarding SpecAugment we had the discussion that there is no clear full implementation in `torch` this is why this is on TODO here atm. 
Since some people said this might even come from outside the Conformer or not be used I would vouch for making it `Optional` and leave this TODO for a PR @JackTemaki and I will tackle next week. Or we say this is 100% required than this PR would be delayed until this is implemented. Would like to hear opinions on this. 

Otherwise feel free to review.

JackTemaki (approved):

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+
+    # nested configurations
+    ff_cfg: ConformerPositionwiseFeedForwardV1Config
+    mhsa_cfg: ConformerMHSAV1Config
+    conv_cfg: ConformerConvolutionV1Config
+
+
+@dataclass
+class ConformerEncoderV1Config(ModelConfiguration):
```
vieting, line 37: Do we have a convention on whether configs should be at the top or right before the corresponding module? I would prefer the latter.

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,110 @@
...
+        residual = x + residual  # [B, T, F]
+        x = self.conv(residual)  #  [B, T, F]
+        residual = x + residual  # [B, T, F]
+        x = self.ff_2(residual)  #  [B, T, F]
+        x = 0.5 * x + residual  #  [B, T, F]
+        x = self.final_layer_norm(x)  #  [B, T, F]
+        return x
+
+
+class ConformerEncoderV1(nn.Module):
```
vieting, line 85: Other modules tend to have a docstring here. I would also add a link to the conformer paper somewhere. Maybe here would be a good place.

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,110 @@
...
+        self.frontend = ConformerFrontendV1(cfg=cfg.front_cfg)
+        self.module_list = torch.nn.ModuleList([ConformerBlockV1(cfg.block_cfg) for _ in range(cfg.num_layers)])
+
+    def forward(self, data_tensor: torch.Tensor, sequence_mask: torch.Tensor):
+        """
+        :param data_tensor: input tensor of shape [B, T', F]
+        :param sequence_mask: mask tensor where 0 defines positions within the sequence and 1 outside [B, T']
+        :return: torch.Tensor of shape [B, T, F']
+
+        F: feature dim after feature extraction, F': internal model feature dim
```
vieting, line 101:
```suggestion
        F: input feature dim, F': internal and output feature dim
```

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+        self.linear = nn.Linear(in_features=cfg.feature_dim, out_features=cfg.hidden_dim)
+        self.dropout = nn.Dropout(p=cfg.dropout)
+
+    def forward(self, data_tensor: torch.Tensor):
+        """
+        :param data_tensor: input tensor of shape [B, T, F]
+        :return: torch.Tensor of shape [B, T', F']
+
+        F: feature dim after feature extraction, F': internal model feature dim
```
vieting, line 65:
```suggestion
        F: input feature dim, F': internal and output feature dim
```

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        :param cfg: conformer frontend configuration
+        """
+        super().__init__()
+        if cfg.spec_aug_cfg:
+            # TODO: This still needs to be implemented
+            raise NotImplementedError
+        else:
+            self.spec_aug = None
+        self.subsampling = nn.Conv1d(
```
vieting, line 50: I would rename `self.subsampling` to `self.conv` or `self.conv_subsampling` or something similar to make clear that there is a convolution involved and not only subsampling by dropping frames or so.

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,77 @@
...
+            in_channels=cfg.feature_dim,
+            out_channels=cfg.feature_dim,
```
vieting, line 52: Did you check in other implementations if this is always done like this? The paper does not tell anything about it, but we might check other implementations for reference.

Atticus1806: I am not 100% sure here. It seems like most frameworks handle the part before the first Conformer encoder independently and don't see it as part of the Conformer (e.g. torchaudio does it like [this](https://pytorch.org/audio/main/_modules/torchaudio/models/conformer.html#Conformer)) In one of the meetings we decided to have this as part of the conformer. we can change this and only have the "stack" as a model here and leave the sub-sampling to the user, but I would not make this change in the PR without other opinions since after all we agreed on a variant in the first place. 

From what I found we have e.g: 
https://espnet.github.io/espnet/_modules/espnet2/asr_transducer/encoder/blocks/conv_input.html#ConvInput
https://speechbrain.readthedocs.io/en/latest/_modules/speechbrain/lobes/models/ESPnetVGG.html#ESPnetVGG

In PaddleSpeech they have it as a BaseEncoder class from which the Conformer then inherits which defines the subsampling in different variations.
https://github.com/PaddlePaddle/PaddleSpeech/blob/5f53e902e1c85a7ec6c1645d61a26701d171428a/paddlespeech/s2t/modules/encoder.py#L59

So as a conclusion I think this all suggests that subsampling is not seen as part of the Conformer. So how we handle this here? Any opinions?

vieting: I know the decision was to stay close to the paper. However, just having one layer here differs from all other implementations (the ones you linked and our own ones) and is most likely suboptimal. The question is whether we want to ignore this for now to have a working version as soon as possible and everything else will be "v2" or if we follow another implementation or if we should make it flexible enough to allow overwriting this.

i6_models/parts/conformer/mhsa.py
```diff
@@ -10,14 +10,17 @@
...
 class ConformerMHSAV1Config(ModelConfiguration):
+    """
+    :param input_dim: input dim and total dimension for query/key and value projections, should be divisible by `num_att_heads`
+    :param num_att_heads: number of attention heads
+    :param att_weights_dropout: attention weights dropout
+    :param dropout: multi-headed self attention output dropout
+    """
+
     input_dim: int
-    """input dim and total dimension for query/key and value projections, should be divisible by `num_att_heads`"""
```
mmz33, line 14: Why this is changed now? i think the proper documentation for dataclass attributes is either this way or something like this:
```
input_dim: int  # ...
```
or did we agree somehwere that we want to use this?

Atticus1806: >As far as I have seen, PyCharm supports normal Class-based param inits now also for dataclasses, so there is no need anymore to have this specific docstring style. But this can be adapted later if wanted.

Is what @JackTemaki said. And from what I saw the doc string came up this way in PyCharm.

albertz: It does not for me?

image

However, the docstring works for this standard variant we had before here.

mmz33: Same for me.

vieting: I'd prefer the variant that @Atticus1806 proposed.

Atticus1806: In my opinion this makes the code more messy because it inreases the lines the code block has. Hovering over the attribute in the code still gives you the doc string.
In the end I don't think this is important, we should just take a decision I can implement so this PR does not take too much longer.

Atticus1806:
@JackTemaki so you successfully tested this now right?

i6_models/config.py
```diff
@@ -51,3 +53,18 @@ def _validate_types(self) -> None:
...
+
+
+@dataclass
+class SubassemblyWithOptions:
+    """
+    Dataclass for a combination of an Subassembly/Part and the corresponding configuration.
+    Also provides a function to construct the corresponding object through this dataclass
+    """
+
+    module_class: Callable[[ModelConfiguration], nn.Module]
```
SimBe195, line 65:
```suggestion
    module_class: type
```
In my opinion expecting a callable here makes this class a little bit too complicated to use. I would expect it to work simply with something like
`frontend = SubassemblyWithOptions(ConformerVGGFrontendV1, frontend_cfg)`.
and not have to introduce some lambda function or whatever to get a callable.
You could get fancy and introduce some abstract base class that all parts/assemblies derive from to make sure that the constructor actually works with an instance of `ModelConfiguration`, but I think simply annotating with `type` is fine.

SimBe195: Hm yes, it seems that you are right. On my side Pylance raised the issue
`Argument of type "Type[ConformerVGGFrontendV1]" cannot be assigned to parameter "module_class" of type "(ModelConfiguration) -> Module" in function "__init__"` but perhaps this is related to something else.

albertz: I did not check but maybe:

`ConformerVGGFrontendV1` actually is like `Callable[[ModelConfiguration], ConformerVGGFrontendV1]`. I'm not sure if it complains because of this difference. You could try `T = TypeVar("T", bound=nn.Module)` and then use `Callable[[ModelConfiguration], T]` as the type here. Or maybe also the same for the cfg, like: `T2 = TypeVar("T2", bound=ModelConfiguration)`, and then `Callable[[T2], T]`. Maybe actually only this second case on cfg.

i6_models/config.py
```diff
@@ -51,3 +53,18 @@ def _validate_types(self) -> None:
...
+    cfg: ModelConfiguration
+
+    def construct(self) -> nn.Module:
+        """Constructs an instance of the given module class"""
+        return self.module_class(self.cfg)
```
SimBe195, line 70:
```suggestion
ConfigType = TypeVar("ConfigType", bound=ModelConfiguration)
ModuleType = TypeVar("ModuleType", bound=nn.Module)


@dataclass
class SubassemblyWithOptions(Generic[ConfigType, ModuleType]):
    """
    Dataclass for a combination of an Subassembly/Part and the corresponding configuration.
    Also provides a function to construct the corresponding object through this dataclass
    """

    module_class: Callable[[ConfigType], ModuleType]
    cfg: ConfigType

    def construct(self) -> ModuleType:
        """Constructs an instance of the given module class"""
        return self.module_class(self.cfg)
```

The original version raises typing issues because concrete instances of module classes only accept subclasses of `ModelConfiguration` and not `ModelConfiguration` itself. This should fix typing issues.

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,115 @@
...
+        :return: torch.Tensor of shape [B, T, F']
+
+        F: input feature dim, F': internal and output feature dim
+        T': data time dim, T: down-sampled time dim (internal time dim)
+        """
+        x = self.frontend(data_tensor)  # [B, T, F']
+        for module in self.module_list:
+            x = module(x, sequence_mask)  # [B, T, F']
+
+        return x
```
curufinwe, line 115:
```suggestion
        x, sequence_mask = self.frontend(data_tensor, sequence_mask)  # [B, T, F']
        for module in self.module_list:
            x = module(x, sequence_mask)  # [B, T, F']

        return x, sequence_mask
```

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,115 @@
...
+        :param cfg: conformer block configuration with subunits for the different conformer parts
+        """
+        super().__init__()
+        self.ff_1 = ConformerPositionwiseFeedForwardV1(cfg=cfg.ff_cfg)
+        self.mhsa = ConformerMHSAV1(cfg=cfg.mhsa_cfg)
+        self.conv = ConformerConvolutionV1(model_cfg=cfg.conv_cfg)
+        self.ff_2 = ConformerPositionwiseFeedForwardV1(cfg=cfg.ff_cfg)
+        self.final_layer_norm = torch.nn.LayerNorm(cfg.ff_cfg.input_dim)
+
+    def forward(self, tensor: torch.Tensor, sequence_mask: torch.Tensor):
```
albertz, line 51:
```suggestion
    def forward(self, tensor: torch.Tensor, sequence_mask: torch.Tensor) -> torch.Tensor:
```

i6_models/config.py
```diff
@@ -51,3 +53,22 @@ def _validate_types(self) -> None:
...
+
+
+@dataclass
+class SubassemblyWithOptions:
+    """
+    Dataclass for a combination of a Subassembly/Part and the corresponding configuration.
+    Also provides a function to construct the corresponding object through this dataclass
+    """
+
+    module_class: Callable[[ConfigType], ModuleType]
```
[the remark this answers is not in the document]
albertz: Also to potentially perform `isinstance(obj, opts.module_class)` checks somewhere.

i6_models/assemblies/conformer/conformer.py
albertz, line 1: If this is all about V1, would you put any V2 into a separate file? Then maybe the file should be renamed to `conformer_v1.py`?

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,115 @@
...
+    Attributes:
+        ff_cfg: Configuration for ConformerPositionwiseFeedForwardV1
+        mhsa_cfg: Configuration for ConformerMHSAV1
+        conv_cfg: Configuration for ConformerConvolutionV1
+    """
+
+    # nested configurations
+    ff_cfg: ConformerPositionwiseFeedForwardV1Config
+    mhsa_cfg: ConformerMHSAV1Config
+    conv_cfg: ConformerConvolutionV1Config
```
albertz, line 32: Do we really want the `_cfg` postfix always? I think it is somewhat obsolete. Whatever we do, I think it should be consistent. See also my other comment.

i6_models/config.py
```diff
@@ -51,3 +53,22 @@ def _validate_types(self) -> None:
 
     def __post_init__(self) -> None:
         self._validate_types()
+
+
+ConfigType = TypeVar("ConfigType", bound=ModelConfiguration)
+ModuleType = TypeVar("ModuleType", bound=nn.Module)
```
albertz, line 59: We might need to extend `ModuleType` even further, as a `Protocol` or so, to define that we expect an `__init__` with one single argument `cfg: ConfigType`.

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,115 @@
...
+    def __init__(self, cfg: ConformerEncoderV1Config):
+        """
+        :param cfg: conformer encoder configuration with subunits for frontend and conformer blocks
+        """
+        super().__init__()
+
+        self.frontend = cfg.frontend.construct()
+        self.module_list = torch.nn.ModuleList([ConformerBlockV1(cfg.block_cfg) for _ in range(cfg.num_layers)])
+
+    def forward(self, data_tensor: torch.Tensor, sequence_mask: torch.Tensor):
```
albertz, line 102:
```suggestion
    def forward(self, data_tensor: torch.Tensor, sequence_mask: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
```

i6_models/assemblies/conformer/conformer.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    Conformer block module
+    """
+
+    def __init__(self, cfg: ConformerBlockV1Config):
+        """
+        :param cfg: conformer block configuration with subunits for the different conformer parts
+        """
+        super().__init__()
+        self.ff_1 = ConformerPositionwiseFeedForwardV1(cfg=cfg.ff_cfg)
```
albertz, line 45: I think it is more consistent (standard) to use `ff1` instead of `ff_1`.
```suggestion
        self.ff1 = ConformerPositionwiseFeedForwardV1(cfg=cfg.ff_cfg)
```
Similarly for `ff2` below.

i6_models/config.py
```diff
@@ -51,3 +53,22 @@ def _validate_types(self) -> None:
...
+
+
+@dataclass
+class SubassemblyWithOptions:
```
albertz, line 63: Shouldn't this derive from `ModelConfiguration`?

albertz, line 63: Regarding naming, this is now `Options` but elsewhere we always use `Config`. Shouldn't we make this more consistent?

i6_models/config.py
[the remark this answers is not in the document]
Atticus1806: with this change pycharm complains though that the type does not match when doing something like:
`frontend = SubassemblyWithConfig(module_class=ConformerVGGFrontendV1, cfg=frontend_cfg)`
`Expected type 'ModuleType', got 'Type[ConformerVGGFrontendV1]' instead `

i6_models/config.py
```diff
@@ -51,3 +53,22 @@ def _validate_types(self) -> None:
...
+    """
+    Dataclass for a combination of a Subassembly/Part and the corresponding configuration.
+    Also provides a function to construct the corresponding object through this dataclass
+    """
+
+    module_class: Callable[[ConfigType], ModuleType]
+    cfg: ConfigType
```
albertz, line 70: I wonder if we really need to have this attrib, so that the cfg is nested here.

Originally, I thought that the config type itself would have the `module_class` attrib.

E.g. `ConformerConvolutionV1Config` below would just additionally have the `module_class = ConformerConvolutionV1`.

Or the `ConformerEncoderV1Config` would just have `module_class = ConformerEncoderV1`.

When setting config options in the encoder for the frontend, currently it would be like:
```python
conformer_cfg.frontend = SubassemblyWithOptions(module_class=MyFrontend, cfg=MyFrontendCfg(...))
```
But with what I suggest, it would just be:
```python
conformer_cfg.frontend = MyFrontendCfg(...)
```

i6_models/config.py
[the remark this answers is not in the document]
albertz: Sorry, yes, it should be:
```suggestion
    module_class: Type[ModuleType]
```

i6_models/config.py
Atticus1806: This is an option too ofc., I think @vieting proposed smth. similar. But i would not implement it in this here, because this requires a full new discussion. Here for now I would just go with how it was proposed in the meeting back when we laid out the structure of the conformer.

albertz: Was this really proposed this way? Actually my understanding of what we discussed in the meeting was what I wrote here, not your variant.

i6_models/config.py
Atticus1806: I think this should be a follow up PR too.

i6_models/config.py
Atticus1806:
```
  conformer_cfg = ConformerOptions()
  conformer_cfg.frontend = SubassemblyWithOptions(repo.assemblies.VGGFrontend, vgg_opts)

```

This is in the meeting notes from 25.4.

i6_models/config.py
albertz: If you do not fix this here, I think you might see a warning in PyCharm, mypy and other tools about wrong typing somewhere. Not sure.

i6_models/config.py
albertz: Ah ok, then I did not paid too much attention during that discussion. I would have voted for the easier variant otherwise. I would in general always vote for the easier solution. But ok then.

i6_models/config.py
Atticus1806: With this the example I posted above does not recieve a PyTorch warning.

i6_models/config.py
Atticus1806: We can change this of course. But I would not do this in this PR since after all this is about the conformer. 
And I would def. prefer a follow-up PR for this (or even one right now I can merge into this).

i6_models/config.py
albertz: I mean in the `construct` method. Or maybe in other places. Not for the creation of the cfg/opts object.

i6_models/config.py
albertz: Also, I think @curufinwe always wants to avoid any further discussion on things which have already been discussed once. Even if there were maybe misunderstandings. E.g. I still think we did some bad choices which make it all too complicated and difficult, not really usable, we could have made it much easier, more modular, more usable, and I'm not sure if everybody really understands this, esp the aspect on code complexity, but we don't want to discuss this anymore.

So basically, once there was some decision, this will not be changed anymore, at least for V1.

So it means, we would now stick with this.

i6_models/assemblies/conformer/conformer.py
albertz, line 1: You want to remove this file, right? You wanted to rename it to `conformer_v1.py`, not copy it.

Atticus1806: yes this was a small mistake i just made with pushing from pycharm :D

i6_models/config.py
Atticus1806: Agreed. In general I am not against rediscussing things that might have been bad decisions, here I don't want to discuss it though because this PR is somewhat unrelated.

> I don't expect that this will be changed now once we have it in master.

Why? If you think something is not good as it is you can open a PR/Issue and then we can rediscuss/change it.

i6_models/assemblies/conformer/conformer.py
Atticus1806: I am not 100% sure about this. So I agree here this is somehwat redundant, so my first instinct would be to remove it too. But on the other hand these configs are also meant as blueprints for other configs where maybe other ff parameters might be relevant (im making up params now but e.g. ff_count or ff_ordering). And then I am not sure if we want to have one "ff" param and the other two ones with some postfix. 
But I am not really decided on this.

i6_models/config.py
albertz: > > I don't expect that this will be changed now once we have it in master.
> 
> Why?

Just from experience. Also that was my understanding on the decisions I'm unhappy with (complicated, difficult, complex, non-modular code design). I'm happy to discuss discuss any of that again.

Maybe the situation in my issue is a bit different, because those things are versionized (`ConformerConvolutionV1`), and as I understand, we would want to just have `ConformerConvolutionV2` for any future easier cleaner code. I just wondered why not directly have the easier and cleaner code from the beginning.

And here in this case, this is not versionized. So any change on this would affect other code which uses it.
Or maybe it should be versionized as well? `SubassemblyWithOptionsV1`? But maybe not.

Maybe we can stop the discussion then here in this PR on this.

i6_models/assemblies/conformer/conformer.py
albertz: I'm not really sure I understand your argument pro having the `_cfg` postfix. Or rather, I don't really see why that is a problem.

christophmluscher (approved):

i6_models/assemblies/conformer/conformer_v1.py
```diff
@@ -0,0 +1,118 @@
...
+    Attributes:
+        ff_cfg: Configuration for ConformerPositionwiseFeedForwardV1
+        mhsa_cfg: Configuration for ConformerMHSAV1
+        conv_cfg: Configuration for ConformerConvolutionV1
+    """
+
+    # nested configurations
+    ff_cfg: ConformerPositionwiseFeedForwardV1Config
+    mhsa_cfg: ConformerMHSAV1Config
+    conv_cfg: ConformerConvolutionV1Config
```
albertz, line 33:
```suggestion
    ff: ConformerPositionwiseFeedForwardV1Config
    mhsa: ConformerMHSAV1Config
    conv: ConformerConvolutionV1Config
```
I thought this was already discussed somewhere but I currently don't find it anymore. Isn't it nicer like this?

i6_models/assemblies/conformer/conformer_v1.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        residual = tensor  #  [B, T, F]
+        x = self.ff1(residual)  #  [B, T, F]
+        residual = 0.5 * x + residual  #  [B, T, F]
+        x = self.mhsa(residual, sequence_mask)  #  [B, T, F]
+        residual = x + residual  # [B, T, F]
+        x = self.conv(residual)  #  [B, T, F]
+        residual = x + residual  # [B, T, F]
+        x = self.ff2(residual)  #  [B, T, F]
+        x = 0.5 * x + residual  #  [B, T, F]
```
albertz, line 66: I find the mix of `tensor`, `x` and `residual` somewhat confusing here. Esp you also sometimes use `residual` as input, sometimes `x`.

Why not directly name the input arg `x`, then you don't have `tensor` here at all. You might also want to require it to be a positional arg.

And you can avoid the `residual` completely maybe. E.g. like:

```suggestion
        x = 0.5 * self.ff1(x) + x  #  [B, T, F]
        x = self.mhsa(x, sequence_mask) + x  #  [B, T, F]
        x = self.conv(x) + x  #  [B, T, F]
        x = 0.5 * self.ff2(x) + x  #  [B, T, F]
```
I personally find this more readable. And it's shorter.

i6_models/assemblies/conformer/conformer_v1.py
Atticus1806: I would say no. For me `ff` would imply that this is the FF module or some factory (like for Norm in the Convolution).
If more people prefer the shorter variation we can change it, otherwise I would leave it for now.

SimBe195 (approved):

vieting (approved):

i6_models/assemblies/conformer/conformer_v1.py
albertz: My argument is not just to have it shorter, but also for consistency. The `_cfg` postfix describes the type of the attrib. For other attribs, we don't have that (e.g. `num_layers_int`).

Atticus1806: I would argue here that for `num_layers` the analogy is that you put `num` and not just `layers`, since in our case we do not pass the `ff` layer but a config for the `ff` layer. But I would go with the majority here, I think both variants are okay.

christophmluscher: I would also prefer `cfg`. it does not add too much and makes a clear distinction between layer/module and config

JackTemaki (approved):

i6_models/assemblies/conformer/conformer_v1.py
curufinwe: I also prefer `_cfg`

Issue #10, opened 2023-05-15
Issue with typegaurd type checking


mmz33:
There is an issue when using python version > 3.8 here: https://github.com/rwth-i6/i6_models/blob/main/i6_models/config.py#L32

This check will fail because `field.type` will be a string instead of actual type because of some behaviour change. See here: https://github.com/cython/cython/issues/5010

The tests done automatically for the PRs uses python 3.7 and this is why they passed.

Example of error message:
```
    def test_conformer_convolution_output_shape():
        def get_output_shape(
            batch, time, features, norm="batch_norm", kernel_size=31, dropout=0.1, activation=nn.functional.silu
        ):
            x = torch.randn(batch, time, features)
            cfg = ConformerConvolutionV1Config(
                channels=features, kernel_size=kernel_size, dropout=dropout, norm=norm, activation=activation
            )
            conformer_conv_part = ConformerConvolutionV1(cfg)
            y = conformer_conv_part(x)
            return y.shape
    
>       assert get_output_shape(10, 50, 250) == (10, 50, 250)

tests/test_conformer.py:18: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/test_conformer.py:11: in get_output_shape
    cfg = ConformerConvolutionV1Config(
<string>:8: in __init__
    ???
i6_models/config.py:38: in __post_init__
    self._validate_types()
i6_models/config.py:33: in _validate_types
    typeguard.check_type(getattr(self, field.name), field.type)
/work/asr4/zeineldeen/py_envs/py_3.10_tf_2.9/lib/python3.10/site-packages/typeguard/_functions.py:74: in check_type
    check_type_internal(value, expected_type, memo)
[...]

PR #11, opened 2023-05-16
Bump python to 3.8

Atticus1806:
Since `i6_core` was upgraded to 3.8 I would suggest we do the same here. 
This will most likely crash the typing as stated in #10, but I think this is good because otherwise we get severe mismatches since most people I know don't locally use 3.7 anymore but rather either 3.8 or even 3.10.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -13,7 +13,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
       with:
-        python-version: 3.7
+        python-version: 3.8
         cache: 'pip'
     - run: pip install black==22.3.0
     - run: black --diff --check .
--- a/.github/workflows/model_tests.yml
+++ b/.github/workflows/model_tests.yml
@@ -16,7 +16,7 @@ jobs:
           path: ""
       - uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.8
           cache: 'pip'
       - run: |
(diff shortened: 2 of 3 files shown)


Atticus1806:
Okay I now see, I misunderstood, this seems to only start happening in 3.9, not 3.8. 
I think this bump should still happen, but we need to discuss about the typing then I guess.

michelwi (approved):
ok, from my side

christophmluscher (approved):

albertz (approved):
From my side very much preferred (even maybe later versions).

But this is maybe mostly a question for @curufinwe who has best overview on what older versions we want to support currently.

albertz:
Did you check with @curufinwe?

Atticus1806:
No but 3.8 was already approved for i6_core, so this should not be a problem.

albertz:
> No but 3.8 was already approved for i6_core, so this should not be a problem.

I don't see that @curufinwe has checked it for i6_core either.

Atticus1806:
AppTek Tests were run on 3.8 already all the time, even before the bump. Thats why the conclusion was that there is no problem iirc.

albertz:
But also, i6_core is maybe separate from RETURNN and i6_models. i6_core is for all preprocessing and training pipelines. Recognition could also run standalone, and might have more restrictions. In any case, we should just check with @curufinwe.

curufinwe:
For posterity: yes, it's ok to update to python 3.8.

PR #12, opened 2023-05-16
Value check for config

Atticus1806:
This adds the option to check some internal values of the config at the creation of the config. C.f. the test for some example. This could be useful if some parameters are to be in relation to each other.

### DIFF
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -33,5 +33,12 @@ class ModelConfiguration:
             except typeguard.TypeCheckError as exc:
                 raise typeguard.TypeCheckError(f'In field "{field.name}" of "{type(self)}": {exc}')
 
+    def _value_check(self) -> None:
+        """
+        Base case for the function supposed to hold the value check of a config.
+        """
+        pass
+
     def __post_init__(self) -> None:
         self._validate_types()
+        self._value_check()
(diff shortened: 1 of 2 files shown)


albertz:
Is this really needed? The user could simply overwrite `__post_init__` directly.

michelwi (approved):

Atticus1806:
Then they need to remember to either call `self._validate_types()` or call the super `__post_init` I think which is also prone to errors.

albertz:
> Then they need to remember to either call `self._validate_types()` or call the super `__post_init` I think which is also prone to errors.

You are always supposed to call `super()...`. The same argument applies for most other Python code, esp also `__init__`, but it's not really a problem.

Atticus1806:
I don't have a  strong opinion on this, so both is fine for me. Most important thing is to have people use and implement it no matter which way we chose. I will leave this up for a bit so people can comment, but in the end we can also just close this after there is no new discussion / no interest.

albertz:
> Most important thing is to have people use and implement it no matter which way we chose.

Actually I prefer that such checks would just be done directly in the module `__init__`, or wherever they are actually relevant. I think that makes the code simpler, and then also easier to understand why some condition is required.

Atticus1806:
Still lets discuss today where to put this.

PR #13, opened 2023-05-16
Fix typecheck

Atticus1806:
Attempt on fixing the typecheck problems in #10 . @mmz33 can you maybe test if this fixes the problems also for you? At least in my setup it fixed raised warnings and failed tests. 
Also raised this here, maybe there is a better way https://github.com/agronholm/typeguard/issues/358.

### DIFF
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -30,9 +30,11 @@ class ModelConfiguration:
         for field in fields(type(self)):
             try:
                 if isinstance(field.type, str):
-                    # there field.type most of the time is a str object which needs to be resolved to the type itself
-                    # at the moment is does not look like there is typeguard support for it, thus manual fixes
-                    # c.f. https://github.com/agronholm/typeguard/issues/358
+                    # Later Python versions use forward references,
+                    # and then field.type is a str object.
+                    # This needs to be resolved to the type itself.
+                    # At the moment is does not look like there is typeguard support for it, thus manual fixes.
+                    # C.f. https://github.com/agronholm/typeguard/issues/358
                     from typeguard._functions import check_type_internal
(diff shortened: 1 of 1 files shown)


Atticus1806:
As an additional discussion point: Do we want to fix the typeguard version in the requirements? Maybe to 4.0.0? This is what I tested my code with right now.

i6_models/config.py
```diff
@@ -29,7 +29,18 @@ class ModelConfiguration:
     def _validate_types(self) -> None:
         for field in fields(type(self)):
             try:
-                typeguard.check_type(getattr(self, field.name), field.type)
+                if isinstance(field.type, str):
```
albertz, line 32: You should add some comment. Maybe referring to the issue you should open on the typeguard repo about how to handle this.

i6_models/config.py
```diff
@@ -30,6 +30,9 @@ def _validate_types(self) -> None:
         for field in fields(type(self)):
             try:
                 if isinstance(field.type, str):
+                    # there field.type most of the time is a str object which needs to be resolved to the type itself
+                    # at the moment is does not look like there is typeguard support for it, thus manual fixes
+                    # c.f. https://github.com/agronholm/typeguard/issues/358
```
albertz, line 35:
```suggestion
                    # Later Python versions use forward references,
                    # and then field.type is a str object.
                    # This needs to be resolved to the type itself.
                    # At the moment is does not look like there is typeguard support for it, thus manual fixes.
                    # C.f. https://github.com/agronholm/typeguard/issues/358
```

mmz33:
I tested this and it fixes the issue described in #10.

albertz (approved):

mmz33 (approved):

Issue #9, opened 2023-05-12
Folder Structure


Atticus1806:
When trying to use the first Conformer today, @JackTemaki brought up that if we now treat `i6_models` as recipe folder, than the depth of 2 of `i6_models` folders is problematic. 
Should we change this now? After the current PRs are through? Do we not change it? Maybe @JackTemaki can give more details.

Atticus1806:
Bumping on this one. I think we should come up with a decision asap, since the more we implement the more difficult moving files will be. I guess if no one replies I guess Nick and I will just take a decision.

curufinwe:
i6_models currently has the "normal" python project structure. The idea is to have this as a "normal" python library. My recommendation thus would be to work around the issue, either by symlinking, adding to PYTHONPATH or installing via pip (once we have a setup.py defined). Having the code all in the top-level makes the code less standalone.

PR #14, opened 2023-05-19
Add __all__ and __init__ files

Atticus1806:
I think it is good to be able to import from e.g. `parts.conformer` and not have to think about the location of every single import.

I was even thinking to add something like `i6_models/model_imports/conformer.py` where all this file does is import all conformer init files from the 3 module types and have them bundled in one file. Opinions?

### DIFF
--- a/i6_models/parts/conformer/__init__.py
+++ b/i6_models/parts/conformer/__init__.py
@@ -1,5 +1,4 @@
 from .convolution import *
 from .feedforward import *
-from .frontend import *
 from .mhsa import *
 from .norm import *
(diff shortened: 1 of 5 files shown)


Atticus1806:
Also PyCharm marked this `Callable` in MHSA typing as unused, removed it out of habit, but I did not bother to create a full new PR for this small change, so thats why this is in here aswell.

JackTemaki (approved):
Accept, but I would also add imports in the init of higher package levels up to i6_models itself. That way we encourage people to import from the root level (so always only `from i6_models import Something`), and we can freely move the location of all the code if necessary.

Atticus1806:
Actually `__init__.py` on highest level kind of breaks pycharm. So i postpone this until #9 is solved.

curufinwe (approved):

PR #15, opened 2023-05-21
Add contributing guidelines

curufinwe:
Same as the meeting notes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,10 @@
+# Code conventions
+
+* use black to lint the code
+* default values for Part/Assembly config:
+  * defaults for common hyper-parameters (e.g. number of layers): no
+  * defaults for rarely/never tuned hyper-parameters (e.g. optimizer constants): maybe
+  * defaults for structural elements (parts / assemblies / pyTorch modules/functions): yes
+        if a PR contains a default (not covered by iii) there should be a good reason for it (in the PR / code)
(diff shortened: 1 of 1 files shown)


Atticus1806 (approved):
I will approve this, but maybe we should change the penultimate line and remove the `for torch.nn.Sequential`. I am not sure if there are any other modules which requires the module instead of a functional but I would assume we would handle them similarly, so we could already state this here. 
E.g. `prefer to use torch.nn.functional over equivalent module unless it’s required for functionality of a certain module (e.g. torch.nn.Sequential)`

mmz33 (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,10 @@
+# Code conventions
+
+* use black to lint the code
+* default values for Part/Assembly config:
+  * defaults for common hyper-parameters (e.g. number of layers): no
+  * defaults for rarely/never tuned hyper-parameters (e.g. optimizer constants): maybe
+  * defaults for structural elements (parts / assemblies / pyTorch modules/functions): yes
```
vieting, line 7:
```suggestion
  * defaults for structural elements (parts / assemblies / PyTorch modules/functions): yes
```
I think this is the convention regarding capitalization, see e.g. the [official website](https://pytorch.org/).

JackTemaki (approved):

vieting (approved):

christophmluscher (approved):

PR #16, opened 2023-05-26
Explicit padding for ConformerConvolutionV1

JackTemaki:
In case we want to use the model in pipelines for ONNX export, we can not use `padding="same"` for now:
https://github.com/pytorch/pytorch/issues/68880

While workarounds would be possible, I would strongly advise to not introduce export hacks already for our first model. Also while export might work, it is unclear if the ONNX runtime does.

### DIFF
--- a/i6_models/parts/conformer/convolution.py
+++ b/i6_models/parts/conformer/convolution.py
@@ -28,6 +28,9 @@ class ConformerConvolutionV1(nn.Module):
     """
     Conformer convolution module.
     see also: https://github.com/espnet/espnet/blob/713e784c0815ebba2053131307db5f00af5159ea/espnet/nets/pytorch_backend/conformer/convolution.py#L13
+
+    Uses explicit padding for ONNX exportability, see:
+    https://github.com/pytorch/pytorch/issues/68880
     """
 
     def __init__(self, model_cfg: ConformerConvolutionV1Config):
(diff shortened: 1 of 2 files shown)


mmz33 (approved):

vieting (approved):

i6_models/parts/conformer/convolution.py
```diff
@@ -28,20 +28,23 @@ class ConformerConvolutionV1(nn.Module):
...
+    https://github.com/pytorch/pytorch/issues/68880
     """
 
     def __init__(self, model_cfg: ConformerConvolutionV1Config):
         """
         :param model_cfg: model configuration for this module
         """
         super().__init__()
-
+        assert model_cfg.kernel_size % 2 == 1, "ConformerConvolutionV1 only supports odd kernel sizes"
```
albertz, line 41: I thought the decision was that this check is in the cfg and not here? (Even though I disagreed with this decision.)

JackTemaki: Good point, why not something like this now?

i6_models/parts/conformer/convolution.py
```diff
@@ -23,25 +23,35 @@ class ConformerConvolutionV1Config(ModelConfiguration):
...
+    https://github.com/pytorch/pytorch/issues/68880
     """
 
     def __init__(self, model_cfg: ConformerConvolutionV1Config):
         """
         :param model_cfg: model configuration for this module
         """
         super().__init__()
-
+        model_cfg.check_valid()
```
Atticus1806, line 48: Now you call `check_valid()` twice, is this really required?

JackTemaki: This seems unavoidable if we want to have a check at creation but also a final check at runtime.

Atticus1806 (approved):

PR #17, opened 2023-05-31
Implements two front-ends for acoustic encoders

christophmluscher:
This is not specific for the Conformer so maybe `parts/conformer` is not the right place?


Atticus1806:
Some tests regarding the downsampling shapes (similar to #4) would be nice :)

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,282 @@
...
+        :return: torch.Tensor of shape [B,T',F]
+        """
+        # conv 2d layers expect shape [B,F,T,C] so we have to transpose here
+        tensor = torch.transpose(tensor, 1, 2)  # [B,F,T]
+        # and add a dim
+        tensor = tensor[:, :, :, None]  # [B,F,T,C]
+        tensor = self.conv1(tensor)
+        tensor = self.conv2(tensor)
+        tensor = torch.transpose(tensor, 1, 2)  # transpose back to [B,T,F,C]
+        tensor = torch.squeeze(tensor)  # [B,T,F]
```
kuacakuaca, line 126: `torch.squeeze` will only remove size 1 dimensions, `torch.reshape` maybe? but don't you just need that operation once at the end of the frontend module as the covolution and pooling layers are all 2-dim.?

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,282 @@
...
+        """
+        T might be reduced to T' depending on ...
+
+        :param tensor: input tensor of shape [B,T,F]
+        :return: torch.Tensor of shape [B,T',F]
+        """
+        # conv 2d layers expect shape [B,F,T,C] so we have to transpose here
+        tensor = torch.transpose(tensor, 1, 2)  # [B,F,T]
+        # and add a dim
+        tensor = tensor[:, :, :, None]  # [B,F,T,C]
```
albertz, line 122: The comment on the shape is wrong. You are not adding the channel dim here. The feature dim *is* the channel dim (the second dim always). You are adding another spatial dim here.
```suggestion
        tensor = tensor[:, :, :, None]  # [B,F,T,H]
```
But this makes me realize: I think you are using the VGG in the wrong way. Actually, I think usually the VGG operates on the time and feature dim, and for the first 2d-conv layer, it adds a dummy channel dim. So it means, above you have:
```python
self.conv1 = nn.Conv2d(1, conv1_channels, ...)
```
And the `transpose` above is not needed. So `tensor` is still [B,T,F]. And here you would do:
```suggestion
        tensor = tensor[:, None]  # [B,C=1,T,F]
```
In this comment, I keep "F" as it was the name of the original dim, but "F" here is now used as a spatial dim for the convolution.

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,282 @@
...
+        super().__init__()
+
+        model_cfg.check_valid()
+
+        conv_padding = self._get_padding(model_cfg.conv_kernel_size)
+        pool_1_padding = self._get_padding(model_cfg.pool_1_kernel_size)
+        pool_2_padding = self._get_padding(model_cfg.pool_2_kernel_size)
+
+        self.conv1 = nn.Conv2d(
+            in_channels=model_cfg.features,
```
albertz, line 77: I think this is wrong. See my comment below. I think the VGG operates on time and feature as the spatial dims. And you would add a dummy channel dim. So it means here:
```suggestion
            in_channels=1,
```

i6_models/parts/conformer/frontend.py
albertz, line 126: I don't understand, why do you squeeze the dim here? The `pooling1` requires a 4D tensor, because it is 2D pooling, or not?

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,282 @@
...
+        :param tensor: input tensor of shape [B,T,F]
+        :return: torch.Tensor of shape [B,T',F]
+        """
+        # conv 2d layers expect shape [B,F,T,C] so we have to transpose here
+        tensor = torch.transpose(tensor, 1, 2)  # [B,F,T]
+        # and add a dim
+        tensor = tensor[:, :, :, None]  # [B,F,T,C]
+        tensor = self.conv1(tensor)
+        tensor = self.conv2(tensor)
+        tensor = torch.transpose(tensor, 1, 2)  # transpose back to [B,T,F,C]
```
albertz, line 125: Why do you do that? I don't think this is right. See my other comments.

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,282 @@
...
+        tensor = torch.squeeze(tensor)  # [B,T,F]
+
+        tensor = self.activation(tensor)
+        tensor = self.pooling1(tensor)
+
+        # conv 2d layers expect shape [B,F,T,C] so we have to transpose here
+        tensor = torch.transpose(tensor, 1, 2)  # [B,F,T]
+        tensor = tensor[:, :, :, None]  # [B,F,T,C]
```
albertz, line 133: Again see my other comments. I think the comment is wrong, but also the `transpose`, and adding the dim here. All this should not be needed when you don't use the `squeeze` above, which is also wrong, I think.
```suggestion
```

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,282 @@
...
+        tensor = self.conv4(tensor)
+        tensor = torch.transpose(tensor, 1, 2)  # transpose back to [B,T,F,C]
+        tensor = torch.squeeze(tensor)  # [B,T,F]
+
+        return tensor
+
+    @staticmethod
+    def _get_padding(input_size: Union[int, Tuple[int, ...]]) -> int:
```
albertz, line 270: Why do you really need this function? Why not directly call `get_padding`?

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,282 @@
...
+        return get_padding(input_size)
+
+
+def get_padding(input_size: Union[int, Tuple[int, ...]]) -> int:
```
albertz, line 274: Maybe it should be private:
```suggestion
def _get_padding(input_size: Union[int, Tuple[int, ...]]) -> int:
```

albertz (changes requested):
I think the main problem is that the VGG is applied in the wrong way. Or at least inconsistent to how it usually is done. I don't think this is intended this way.

tests/test_conformer.py
```diff
@@ -9,6 +9,12 @@
     ConformerPositionwiseFeedForwardV1,
     ConformerPositionwiseFeedForwardV1Config,
 )
+from i6_models.parts.conformer.frontend import (
+    ConformerVGGFrontendV1,
+    ConformerVGGFrontendV1Config,
+    ConformerVGGFrontendV2Config,
+    ConformerVGGFrontendV2,
```
albertz, line 16:
```suggestion
    ConformerVGGFrontendV2,
    ConformerVGGFrontendV2Config,
```
Just for consistency.

i6_models/parts/conformer/frontend.py
albertz, line 1: We might want to have a generic interface (base class) for frontend modules, such that we can easily exchange them later on.

It might be a protocol or a base class.

Currently we have here just `(tensor: torch.Tensor) -> torch.Tensor`. However, more generic would be if we have masking or seq lens as another argument. E.g. for our BLSTM-based frontend, we likely would want to handle padding correctly. Or maybe we would want a self-attention-based frontend later, or whatever. In most cases, we want to handle the padding correctly. So more generic would be `(tensor: torch.Tensor, seq_lens: torch.Tensor) -> torch.Tensor`.

```python
class FrontendIface(Protocol, nn.Module):
  def forward(self, tensor: torch.Tensor, seq_lens: torch.Tensor) -> torch.Tensor:
    ...
```

tests/test_conformer.py
```diff
@@ -84,3 +90,73 @@ def get_output(shape, norm):
...
+    def get_output_shape(
+        batch,
+        time,
+        features,
+        conv1_channels,
+        conv2_channels,
+        conv3_channels,
+        conv4_channels,
+        pool_red_1,
+        pool_red_2,
```
albertz, line 107:
```suggestion
        pool1_red,
        pool2_red,
```
Consistency.

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,279 @@
...
+        tensor = torch.squeeze(tensor)  # [B,T,F]
+
+        return tensor
+
+
+def _get_padding(input_size: Union[int, Tuple[int, ...]]) -> int:
+    if isinstance(input_size, int):
+        out = (input_size - 1) // 2
+    elif isinstance(input_size, tuple):
+        out = min(input_size) // 2
```
albertz, line 275: This seems to be a bit arbitrary to me.

And also inconsistent to the case above (no -1 here?).

Why not:
```suggestion
        out = tuple((s - 1) // 2 for s in input_size)
```

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,279 @@
...
+        tensor = self.conv4(tensor)
+        tensor = torch.transpose(tensor, 1, 2)  # transpose back to [B,T,F,C]
+        tensor = torch.squeeze(tensor)  # [B,T,F]
+
+        return tensor
+
+
+def _get_padding(input_size: Union[int, Tuple[int, ...]]) -> int:
+    if isinstance(input_size, int):
+        out = (input_size - 1) // 2
```
albertz, line 273:
```suggestion
        return (input_size - 1) // 2
```
This would be slightly shorter code (thus easier to read).

i6_models/parts/conformer/frontend.py
albertz: Or:
```suggestion
        return tuple((s - 1) // 2 for s in input_size)
```

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,279 @@
...
+    elif isinstance(input_size, tuple):
+        out = min(input_size) // 2
+    else:
+        raise NotImplementedError
```
albertz, line 277: This is the wrong exception. `NotImplementedError` is only if there is an expected possible case here which is not implemented. But you are not really expecting anything else here, so the right exception is `TypeError`. Sth like:
```suggestion
        raise TypeError(f"unexpected size type {type(input_size)}")
```

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,279 @@
...
+        tensor = torch.transpose(tensor, 1, 2)  # [B,F,T]
+        tensor = tensor[:, :, :, None]  # [B,F,T,C]
+        tensor = self.conv4(tensor)
+        tensor = torch.transpose(tensor, 1, 2)  # transpose back to [B,T,F,C]
+        tensor = torch.squeeze(tensor)  # [B,T,F]
+
+        return tensor
+
+
+def _get_padding(input_size: Union[int, Tuple[int, ...]]) -> int:
```
albertz, line 271: I wonder a bit why you use the padding at all. Is this intentional? I think e.g. in ESPnet, they don't add any padding in the frontend. I think in most other implementations, they would not add padding. But I think in our TF-based setups, we add padding, but probably only because people never really thought about it and it was the default for our `ConvLayer`. Maybe you also want to have it configurable?

Further, maybe you might want padding only in the time dimension, but not in the other dimension?

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,279 @@
...
+            assert self.pool2_kernel_size % 2 == 1, "ConformerVGGFrontendV1 only supports odd kernel sizes"
+
+    def __post__init__(self):
+        super().__post_init__()
+        self.check_valid()
+
+
+class ConformerVGGFrontendV1(nn.Module):
+    """
+    Convolutional Front-End
```
albertz, line 63: As said above, you should describe with words (very shortly, maybe 5-10 words) what this does (so that the reader can easily grasp the differences to other frontends).

Also, where do you have this from? Some references would be nice. Did we use that in some other papers? Referencing those papers would also be helpful.

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,279 @@
...
+            assert self.pool_kernel_size % 2 == 1, "ConformerVGGFrontendV2 only supports odd kernel sizes"
+
+    def __post__init__(self):
+        super().__post_init__()
+        self.check_valid()
+
+
+class ConformerVGGFrontendV2(nn.Module):
+    """
+    Convolutional Front-End
```
albertz, line 181: Same comment here, docstring should be extended.

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,279 @@
...
+    """
+    Attributes:
+        in_features: feature dimension of input
+        conv1_channels: number of channels for first conv layers
+        conv2_channels: number of channels for second conv layers
+        conv3_channels: number of channels for third conv layers
+        conv4_channels: number of channels for fourth dconv layers
+        conv_kernel_size: kernel size of conv layers
+        pool1_kernel_size: kernel size of first pooling layer
+        pool1_strides: strides of first pooling layer
```
Atticus1806, line 30: Is plural for `stride` in the doc intended?

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,279 @@
...
+    def __init__(self, model_cfg: ConformerVGGFrontendV1Config):
+        """
+        :param model_cfg: model configuration for this module
+        """
+        super().__init__()
+
+        model_cfg.check_valid()
+
+        conv_padding = self._get_padding(model_cfg.conv_kernel_size)
+        pool1_padding = self._get_padding(model_cfg.pool1_kernel_size)
```
Atticus1806, line 78: Is the `self` call correct here?

i6_models/parts/conformer/frontend.py
Atticus1806, line 271: Can you add a doc so we can understand the way of thinking here?

tests/test_conformer.py
```diff
@@ -84,3 +90,73 @@ def get_output(shape, norm):
...
+            in_features=features,
+            conv1_channels=conv1_channels,
+            conv2_channels=conv2_channels,
+            conv3_channels=conv3_channels,
+            conv4_channels=conv4_channels,
+            conv_kernel_size=(3, 3),
+            pool1_kernel_size=(pool_red_1, 1),
+            pool1_strides=(pool_red_1, 1),
+            pool2_kernel_size=(pool_red_2, 1),
+            pool2_strides=(pool_red_2, 1),
```
Atticus1806, line 121:
```suggestion
            pool1_kernel_size=(pool1_red, 1),
            pool1_strides=(pool1_red, 1),
            pool2_kernel_size=(pool2_red, 1),
            pool2_strides=(pool2_red, 1),
```

i6_models/parts/conformer/frontend.py
```diff
@@ -0,0 +1,282 @@
+from __future__ import annotations
+
+__all__ = [
+    "ConformerVGGFrontendV1Config",
+    "ConformerVGGFrontendV1",
+    "ConformerVGGFrontendV2Config",
+    "ConformerVGGFrontendV2",
```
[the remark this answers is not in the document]
christophmluscher: > Both have 4 conv layers? And by intention, this is fixed, right?

yes

> Why do you have all channels individually configurable in one case but not the other case? This seems inconsistent to me. I would prefer to have this consistent.

what would you prefer? configurable or not?

i6_models/parts/conformer/frontend.py
christophmluscher: > Uses explicit padding for ONNX exportability, see: https://github.com/pytorch/pytorch/issues/68880

from Nick, I think the explicit padding is then here also necessary

albertz: No, what you link is about `padding="same"` vs `padding=<some non-zero int>`. Or `padding="valid"` vs `padding=0`.

When I say, why padding at all, I mean, why not `padding=0`?

i6_models/parts/conformer/frontend.py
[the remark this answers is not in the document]
albertz: I like to have it individually configurable.

i6_models/parts/conformer/frontend.py
christophmluscher: the idea was so that the time dimension does not change. without padding it seems that we loose a few frames.. I made this more configurable, so that the user has more influence on this

i6_models/parts/conformer/frontend.py
christophmluscher: This would be a discussion for the meeting IMHO

@curufinwe @JackTemaki @michelwi @Atticus1806 @mmz33 @kuacakuaca @Judyxujj @vieting

i6_models/parts/conformer/frontend.py
christophmluscher: I am using `torch.flatten` now. for `torch.reshape` do I need to know the output tensor shape?? so batch and time dim or so? or can I get that somehow elegantly?

i6_models/parts/conformer/frontend.py
christophmluscher: @albertz why would you make this a `Protocol` and not an `ABC`?

- usage easier for the user?
- run time speed up?

other things?

albertz: `Protocol` would not require that frontends need to derive from this class. Frontends would simply derive from `nn.Module`. So it potentially also matches to external written frontend modules.

albertz: For completeness, I think in the last meeting, we agreed on the interface:
```python
def forward(self, tensor: torch.Tensor, seq_mask: torch.Tensor) -> torch.Tensor:
```
Right? I.e. with the seq mask, not the seq len.

Another aspect which I realized via the discussion in https://github.com/rwth-i6/i6_models/pull/19:

Do we require that `tensor` and `seq_mask` are on the GPU device here? (Or maybe more generally, whatever device the parameters are on.)

For any frontend which requires seq lens, this can be potentially slow: It's not only that we need to perform a `sum` to get the seq lens (this was already clear), but also, for many ops which use seq lens, they require the seq lens on CPU, so it involves a GPU -> CPU copy, which can be very slow.

Judyxujj:
I have two questions here,
1. the param `stride` should be added to the Config, since usually we set stride in Conv2d to apply the time subsampling
2. after the front end convolution models, we need to apply one additional linear layer to project the dimension back to the model size. Should the linear layer be included in the front end, or should it be included in the ConformerBlock? I would prefer the former

christophmluscher:
> I have two questions here,
> 
> 1. the param `stride` should be added to the Config, since usually we set stride in Conv2d to apply the time subsampling

already added but not pushed yet :)

> 2. after the front end convolution models, we need to apply one additional linear layer to project the dimension back to the model size. Should the linear layer be included in the front end, or should it be included in the ConformerBlock? I would prefer the former

I was thinking about putting this as an option into the frontend: with param `int N` -> add linear layer with `N` outputs, None -> no linear layer

@Atticus1806

christophmluscher:
tests fail due to the seq mask update missing. I am not 100% sure how to perform the update in a clean fashion. Is there a PyTorch way to do this?

albertz:
> tests fail due to the seq mask update missing. I am not 100% sure how to perform the update in a clean fashion. Is there a PyTorch way to do this?

I'm not exactly sure what you refer to. What update do you mean? You mean how to compute the seq mask which is supposed to be returned by the frontend? As I mentioned, you could apply maxpooling with the same striding and kernel size just as the other operations.

christophmluscher:
> you could apply maxpooling with the same striding and kernel size just as the other operations.

this would also work for conv layers?

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,354 @@
...
+        if self.include_linear_layer:
+            tensor = self.linear(tensor)
+
+        return tensor, sequence_mask
+
+
+def _get_padding(input_size: Union[int, Tuple[int, ...]]) -> Union[int, Tuple[int, ...]]:
+    """
+    get padding in order to not reduce the time dimension
+    :param input_size:
```
albertz, line 346: I think there always should be an empty line between the description and the params.
```suggestion

    :param input_size:
```

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,354 @@
...
+
+        :param tensor: input tensor of shape [B,T,F]
+        :param sequence_mask: masking tensor of shape [B,T], contains length information of the sequences
+        :return: torch.Tensor of shape [B,T',F']
+        """
+        tensor = tensor[:, None, :, :]  # [B,C=1,T,F]
+
+        tensor = self.conv1(tensor)
+        tensor = self.activation(tensor)
+        tensor = self.pool(tensor)  # [B,C,T',F]
```
albertz, line 324: The comment is wrong, right? The F might potentially also change.
```suggestion
        tensor = self.pool(tensor)  # [B,C,T',F']
```

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,354 @@
...
+
+        tensor = self.conv2(tensor)  # [B,C,T",F]
```
albertz, line 326:
```suggestion
        tensor = self.conv2(tensor)  # [B,C,T",F"]
```

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,354 @@
...
+        tensor = self.activation(tensor)
+
+        tensor = self.conv3(tensor)
+        tensor = self.activation(tensor)
+
+        tensor = self.conv4(tensor)
+
```
albertz, line 333:
```suggestion
        sequence_mask = _mask_pool(
            sequence_mask, self.conv1.kernel_size[0], self.conv1.stride[0], self.conv1.padding[0]
        )
        sequence_mask = _mask_pool(
            sequence_mask, self.pool.kernel_size[0], self.pool.stride[0], self.pool.padding[0]
        )
        sequence_mask = _mask_pool(
            sequence_mask, self.conv2.kernel_size[0], self.conv2.stride[0], self.conv2.padding[0]
        )
        sequence_mask = _mask_pool(
            sequence_mask, self.conv3.kernel_size[0], self.conv3.stride[0], self.conv3.padding[0]
        )
        sequence_mask = _mask_pool(
            sequence_mask, self.conv4.kernel_size[0], self.conv4.stride[0], self.conv4.padding[0]
        )

```

`_mask_pool` is defined below.

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,354 @@
...
+    get padding in order to not reduce the time dimension
+    :param input_size:
+    :return:
+    """
+    if isinstance(input_size, int):
+        return (input_size - 1) // 2
+    elif isinstance(input_size, tuple):
+        return tuple((s - 1) // 2 for s in input_size)
+    else:
+        raise TypeError(f"unexpected size type {type(input_size)}")
```
albertz, line 354:
```suggestion
        raise TypeError(f"unexpected size type {type(input_size)}")


def _mask_pool(seq_mask, kernel_size, stride, padding):
    """
    :param seq_mask: [B,T]
    :param kernel_size:
    :param stride:
    :param padding:
    :return: [B,T'] using maxpool
    """
    seq_mask = torch.unsqueeze(seq_mask, 1)  # [B,1,T]
    seq_mask = nn.functional.max_pool1d(seq_mask, kernel_size, stride, padding)  # [B,1,T']
    seq_mask = torch.squeeze(seq_mask, 1)  # [B,T']
    return seq_mask
```

tests/test_conformer.py
```diff
@@ -84,3 +90,234 @@ def get_output(shape, norm):
...
+        in_dim,
+        out_dim,
+    ):
+        data_input = torch.randn(batch, time, features)
+        data_input = torch.cat((data_input, torch.zeros(batch, time_padding, features)), dim=1)
+        data_input = torch.cat((data_input, torch.randn(batch, time + time_padding, features)), dim=0)
+
+        sequence_mask = torch.zeros(batch, time)
+        sequence_mask = torch.cat((sequence_mask, torch.ones(batch, time_padding)), dim=1)
+        sequence_mask = torch.cat((sequence_mask, torch.zeros(batch, time + time_padding)), dim=0)
```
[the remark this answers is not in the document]
christophmluscher: my idea was that for the frontend the input sequence defines the sequence mask. but maybe that is not general enough?!

i6_models/parts/frontend/vgg.py
Atticus1806: Does this expect the seq_mask to be integer? Because in my variant I am using a boolean mask and get the following error:
`RuntimeError: "max_pool2d_with_indices_out_cuda_frame" not implemented for 'Bool'` 
(yes 2d even though the function where this happens is the `max_pool1d` call

tests/test_conformer.py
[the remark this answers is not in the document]
albertz: The input has potentially multiple sequences of different lengths, so it has shape [B,T,F], and the seq mask describes the seq len of each of the sequences. The input itself does not define the seq mask or the seq len. The seq mask or seq len is some additional information you need to pass. Via the seq mask, you can derive the seq len, and via the seq len, you can derive the seq mask. Although the seq mask is a bit more generic than the seq len.

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,382 @@
...
+
+def _mask_pool(seq_mask, kernel_size, stride, padding):
+    """
+    :param seq_mask: [B,T]
+    :param kernel_size:
+    :param stride:
+    :param padding:
+    :return: [B,T'] using maxpool
+    """
+    seq_mask = torch.unsqueeze(seq_mask, 1)  # [B,1,T]
```
albertz, line 379: As mentioned in another commend (which you already clicked as resolved, although it is not resolved?), you miss a cast to int here, and then later a cast back to bool.
```suggestion
    seq_mask = torch.unsqueeze(seq_mask, 1)  # [B,1,T]
    seq_mask = seq_mask.int()  # max_pool needs integer
```

tests/test_conformer.py
albertz, line 187: Btw, another thing here: seq mask should be a bool tensor, I guess, but it is not here.

i6_models/parts/frontend/protocol.py
```diff
@@ -0,0 +1,26 @@
...
+        """
+        :param tensor: input tensor of shape [B,T,F]
+        :param sequence_mask: masking tensor of shape [B,T], contains length information of the sequences
+        :return: torch.Tensor of shape [B,T',F']
+        """
+        raise NotImplementedError
+
+
+class FrontendInterface(BaseFrontendInterface, nn.Module):
+    @abstractmethod
```
albertz, line 19: Same.
```suggestion
```

i6_models/parts/frontend/protocol.py
```diff
@@ -0,0 +1,26 @@
+import torch
+import torch.nn as nn
+from abc import abstractmethod
+from typing import Protocol, Tuple
+
+
+class BaseFrontendInterface(Protocol):
+    @abstractmethod
```
albertz, line 8: It's not supposed to be an abstract method?
```suggestion
```

i6_models/parts/frontend/protocol.py
```diff
@@ -0,0 +1,26 @@
+import torch
+import torch.nn as nn
+from abc import abstractmethod
+from typing import Protocol, Tuple
+
+
+class BaseFrontendInterface(Protocol):
```
albertz, line 7: Should this also be versionized?
```suggestion
class BaseFrontendInterfaceV1(Protocol):
```
?

i6_models/parts/frontend/protocol.py
albertz, line 1: Maybe the filename should be versionized?
`protocol_v1.py`?

i6_models/parts/frontend/vgg.py
albertz, line 1: Maybe the filename should be versionized? `vgg_v1.py`?

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,384 @@
...
+    conv_kernel_size: IntTupleIntType
+    conv_padding: Optional[IntTupleIntType]
+    pool1_kernel_size: IntTupleIntType
+    pool1_stride: Optional[IntTupleIntType]
+    pool1_padding: Optional[IntTupleIntType]
+    pool2_kernel_size: IntTupleIntType
+    pool2_stride: Optional[IntTupleIntType]
+    pool2_padding: Optional[IntTupleIntType]
+    activation: Union[nn.Module, Callable[[torch.Tensor], torch.Tensor]]
+    linear_input_dim: Optional[int]
```
albertz, line 56: Does it really make sense to have this configurable? There is only one single valid value for it, which is via the flattening of the layer before.

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,384 @@
...
+    pool_padding: Optional[IntTupleIntType]
+    conv2_channels: int
+    conv2_stride: IntTupleIntType
+    conv3_channels: int
+    conv3_stride: IntTupleIntType
+    conv4_channels: int
+    conv_kernel_size: IntTupleIntType
+    conv_padding: Optional[IntTupleIntType]
+    activation: Union[nn.Module, Callable[[torch.Tensor], torch.Tensor]]
+    linear_input_dim: Optional[int]
```
albertz, line 226: Same here.

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,384 @@
...
+        """
+        # and add a dim
+        tensor = tensor[:, None, :, :]  # [B,C=1,T,F]
+
+        tensor = self.conv1(tensor)
+        tensor = self.conv2(tensor)
+
+        tensor = self.activation(tensor)
+        tensor = self.pool1(tensor)  # [B,C,T',F']
+        sequence_mask = _mask_pool(
```
albertz, line 168: This is not enough. You need to do the same for `self.conv1` and `self.conv2`, or not? This is what I wrote in my earlier comment.

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,384 @@
...
+            _get_int_tuple_int(self.pool1.stride, 0),
+            _get_int_tuple_int(self.pool1.padding, 0),
+        )
+
+        tensor = self.conv3(tensor)
+        tensor = self.conv4(tensor)
+
+        tensor = self.activation(tensor)
+        tensor = self.pool2(tensor)  # [B,C,T",F"]
+        sequence_mask = _mask_pool(
```
albertz, line 180: Again, `_mask_pool` missing for `conv3` and `conv4`.

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,384 @@
...
+        :return: torch.Tensor of shape [B,T',F']
+        """
+        tensor = tensor[:, None, :, :]  # [B,C=1,T,F]
+
+        tensor = self.conv1(tensor)
+        tensor = self.activation(tensor)
+        tensor = self.pool(tensor)  # [B,C,T,F']
+
+        tensor = self.conv2(tensor)  # [B,C,T',F']
+        if self.conv2.stride[0] > 0:
```
albertz, line 330: Why is this `if` needed? You can move such logic to `_mask_pool`. But it's not really strictly necessary anyway to have this.

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,384 @@
...
+            sequence_mask = _mask_pool(
+                sequence_mask, self.conv2.kernel_size[0], self.conv2.stride[0], self.conv2.padding[0]
+            )
+        tensor = self.activation(tensor)
+
+        tensor = self.conv3(tensor)  # [B,C,T",F']
+        if self.conv3.stride[0] > 0:
```
albertz, line 337: Same.

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,384 @@
...
+
+        :param tensor: input tensor of shape [B,T,F]
+        :param sequence_mask: masking tensor of shape [B,T], contains length information of the sequences
+        :return: torch.Tensor of shape [B,T',F']
+        """
+        tensor = tensor[:, None, :, :]  # [B,C=1,T,F]
+
+        tensor = self.conv1(tensor)
+        tensor = self.activation(tensor)
+        tensor = self.pool(tensor)  # [B,C,T,F']
```
albertz, line 327: `_mask_pool` missing for `conv1` and `pool`.

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,384 @@
...
+        tensor = self.activation(tensor)
+
+        tensor = self.conv3(tensor)  # [B,C,T",F']
+        if self.conv3.stride[0] > 0:
+            sequence_mask = _mask_pool(
+                sequence_mask, self.conv3.kernel_size[0], self.conv3.stride[0], self.conv3.padding[0]
+            )
+        tensor = self.activation(tensor)
+
+        tensor = self.conv4(tensor)
```
albertz, line 343: `_mask_pool` missing here.

i6_models/parts/frontend/vgg.py
```diff
@@ -0,0 +1,384 @@
...
+    :return: [B,T'] using maxpool
+    """
+    seq_mask = torch.unsqueeze(seq_mask, 1)  # [B,1,T]
+    seq_mask = nn.functional.max_pool1d(seq_mask, kernel_size, stride, padding)  # [B,1,T']
+    seq_mask = torch.squeeze(seq_mask, 1)  # [B,T']
+    return seq_mask
+
+
+def _get_int_tuple_int(variable: IntTupleIntType, index: int) -> int:
+    return variable[index] if isinstance(variable, Tuple) else variable
```
albertz, line 384:
```suggestion
    return variable[index] if isinstance(variable, tuple) else variable
```

christophmluscher:
I will move the other frontend to another PR.

@Atticus1806 tested this within the conformer so should be ok now

everything else I would like to move to follow up PRs so that we can merge and start using this

i6_models/parts/frontend/common.py
```diff
@@ -0,0 +1,43 @@
...
+def _mask_pool(seq_mask: torch.Tensor, kernel_size: int, stride: int, padding: int) -> torch.Tensor:
+    """
+    :param seq_mask: [B,T]
+    :param kernel_size:
+    :param stride:
+    :param padding:
+    :return: [B,T'] using maxpool
+    """
+    seq_mask = torch.unsqueeze(seq_mask, 1)  # [B,1,T]
+    print("mask")
```
Atticus1806, line 34: I think these were for testing right?

i6_models/parts/frontend/vgg_act.py
```diff
@@ -0,0 +1,192 @@
...
+
+        model_cfg.check_valid()
+
+        conv_padding = (
+            model_cfg.conv_padding if model_cfg.conv_padding is not None else _get_padding(model_cfg.conv_kernel_size)
+        )
+        pool1_padding = model_cfg.pool1_padding if model_cfg.pool1_padding is not None else 0
+        pool2_padding = model_cfg.pool2_padding if model_cfg.pool2_padding is not None else 0
+
+        self.include_linear_layer = True if model_cfg.linear_output_dim is not None else False
```
Atticus1806, line 105: I think here we also want an assert to check that `linear_input_dim` is also set.

i6_models/parts/frontend/protocol.py
```diff
@@ -0,0 +1,23 @@
...
+    def forward(self, tensor: torch.Tensor, sequence_mask: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        :param tensor: input tensor of shape [B,T,F]
+        :param sequence_mask: masking tensor of shape [B,T], contains length information of the sequences
+        :return: torch.Tensor of shape [B,T',F']
+        """
+        raise NotImplementedError
+
+
+class FrontendInterfaceV1(BaseFrontendInterfaceV1, nn.Module):
```
albertz, line 16: I'm not sure it works that you can combine a `Protocol` and the condition that sth is derived from `nn.Module`. I think this does not work.

We had a similar discussion here:
https://github.com/rwth-i6/i6_models/pull/21#discussion_r1245025446
https://github.com/rwth-i6/i6_models/pull/21#discussion_r1246450812

I'm not sure though what would be the best option here. I think we can only have one: either the `Protocol`, or saying it must derive from `nn.Module`.

Maybe the protocol, the interface can just be informal, i.e. we define it somewhere in the documentation, but otherwise it is not enforced. Although you would get an error obviously when you pass some incorrect module.

i6_models/parts/frontend/protocol.py
```diff
@@ -0,0 +1,23 @@
...
+import torch.nn as nn
+from typing import Protocol, Tuple
+
+
+class BaseFrontendInterfaceV1(Protocol):
+    def forward(self, tensor: torch.Tensor, sequence_mask: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        :param tensor: input tensor of shape [B,T,F]
+        :param sequence_mask: masking tensor of shape [B,T], contains length information of the sequences
+        :return: torch.Tensor of shape [B,T',F']
```
albertz, line 11: The doc is incomplete, it is also the new seq mask.

i6_models/parts/frontend/vgg_act.py
```diff
@@ -0,0 +1,197 @@
...
+        # and add a dim
+        tensor = tensor[:, None, :, :]  # [B,C=1,T,F]
+
+        tensor = self.conv1(tensor)
+        tensor = self.conv2(tensor)
+
+        tensor = self.activation(tensor)
+        tensor = self.pool1(tensor)  # [B,C,T',F']
+        if sequence_mask is not None:
+            sequence_mask = sequence_mask.float()
```
albertz, line 169: I think it would be better if this cast would happen inside `_mask_pool` and it always stays a bool mask in this function.

i6_models/parts/frontend/vgg_act.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        # and add a dim
+        tensor = tensor[:, None, :, :]  # [B,C=1,T,F]
+
+        tensor = self.conv1(tensor)
+        tensor = self.conv2(tensor)
+
+        tensor = self.activation(tensor)
+        tensor = self.pool1(tensor)  # [B,C,T',F']
+        if sequence_mask is not None:
```
albertz, line 168: I think after the discussion today we said that we don't want to have this optional, right?

i6_models/parts/frontend/vgg_act.py
```diff
@@ -0,0 +1,197 @@
...
+        )
+        self.activation = model_cfg.activation
+        if self.include_linear_layer:
+            self.linear = nn.Linear(
+                in_features=model_cfg.linear_input_dim,
+                out_features=model_cfg.linear_output_dim,
+                bias=True,
+            )
+
+    def forward(self, tensor: torch.Tensor, sequence_mask: Optional[torch.Tensor]) -> Tuple[torch.Tensor, torch.Tensor]:
```
albertz, line 152:
```suggestion
    def forward(self, tensor: torch.Tensor, sequence_mask: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
```
Not optional, see also my other comment.

Or if you keep it optional, then the return type would be wrong.
```suggestion
    def forward(self, tensor: torch.Tensor, sequence_mask: Optional[torch.Tensor]) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
```

i6_models/parts/frontend/vgg_act.py
```diff
@@ -0,0 +1,197 @@
...
+        tensor = tensor[:, None, :, :]  # [B,C=1,T,F]
+
+        tensor = self.conv1(tensor)
+        tensor = self.conv2(tensor)
+
+        tensor = self.activation(tensor)
+        tensor = self.pool1(tensor)  # [B,C,T',F']
+        if sequence_mask is not None:
+            sequence_mask = sequence_mask.float()
+            sequence_mask = _mask_pool(
```
albertz, line 170: Before you do the mask pooling for `pool1`, you need to do the mask pooling for `conv1` and `conv2`. See also my earlier (yet unresolved) comments on this. There are also more mask poolings missing here.

tests/test_frontend_vgg.py
```diff
@@ -0,0 +1,129 @@
...
+        out_dim,
+    ):
+        data_input = torch.randn(batch, time, features)
+        data_input = torch.cat((data_input, torch.zeros(batch, time_padding, features)), dim=1)
+        data_input = torch.cat((data_input, torch.randn(batch, time + time_padding, features)), dim=0)
+
+        sequence_mask = torch.ones(batch, time)
+        sequence_mask = torch.cat((sequence_mask, torch.zeros(batch, time_padding)), dim=1)
+        sequence_mask = torch.cat((sequence_mask, torch.ones(batch, time + time_padding)), dim=0)
+        sequence_mask = sequence_mask.bool()
```
albertz, line 36: I don't really understand this arbitrary creation of the seq mask. Is this on purpose? Why don't you pass it as an argument, just like the `data_input`?

tests/test_frontend_vgg.py
```diff
@@ -0,0 +1,129 @@
...
+)
+
+
+def test_conformer_vgg_layer_act_frontend_v1():
+    torch.manual_seed(42)
+
+    def get_output_shape(
+        batch,
+        time,
+        time_padding,
```
albertz, line 18: What is this `time_padding` used for? I don't think you need that.

tests/test_frontend_vgg.py
```diff
@@ -0,0 +1,129 @@
...
+        in_dim,
+        out_dim,
+    ):
+        data_input = torch.randn(batch, time, features)
+        data_input = torch.cat((data_input, torch.zeros(batch, time_padding, features)), dim=1)
+        data_input = torch.cat((data_input, torch.randn(batch, time + time_padding, features)), dim=0)
+
+        sequence_mask = torch.ones(batch, time)
+        sequence_mask = torch.cat((sequence_mask, torch.zeros(batch, time_padding)), dim=1)
+        sequence_mask = torch.cat((sequence_mask, torch.ones(batch, time + time_padding)), dim=0)
```
albertz, line 35: This `cat` does not make sense to me. Why are you concatenating in the batch dimension? So the output tensor has shape [batch * 2, time + time_padding]. That looks wrong.

i6_models/parts/frontend/vgg_act.py
```diff
@@ -0,0 +1,197 @@
...
+    conv_kernel_size: IntTupleIntType
+    conv_padding: Optional[IntTupleIntType]
+    pool1_kernel_size: IntTupleIntType
+    pool1_stride: Optional[IntTupleIntType]
+    pool1_padding: Optional[IntTupleIntType]
+    pool2_kernel_size: IntTupleIntType
+    pool2_stride: Optional[IntTupleIntType]
+    pool2_padding: Optional[IntTupleIntType]
+    activation: Union[nn.Module, Callable[[torch.Tensor], torch.Tensor]]
+    linear_input_dim: Optional[int]
```
albertz, line 53: Why is `linear_input_dim` configurable? This is sth which is implied by the other settings, via the input dimension (F), or not? I think I rather would configure the `in_features` (F) for this module, instead of this.

i6_models/parts/frontend/vgg.py
christophmluscher: this is supposed to be an implicit user check.. this is what the user expects as output size from the front end. I prefer to leave it in.

i6_models/parts/frontend/vgg.py
christophmluscher: the stride controls the mask size if I understand correctly.. so I should only update the mask when I allow setting the stride. since I am following our implementation I do not allow this everywhere.

i6_models/parts/frontend/vgg_act.py
christophmluscher: from our meeting notes, I took that we use the export flag to ignore the mask for export. this was in reference to this?!

tests/test_frontend_vgg.py
christophmluscher: I am not sure why this matters for the tests here. Can you elaborate?

tests/test_frontend_vgg.py
christophmluscher: idea was to have a sequence with padding and one without... so that I can check if the padding somehow messes something up

tests/test_frontend_vgg.py
christophmluscher: in order to simulate having different seq lengths within one batch.. maybe you know a cleaner way to create this?

i6_models/parts/frontend/vgg_act.py
christophmluscher: I think these are two points.. 
why would you make the `in_features` configurable.. for our use case this should normally not be necessary, right?
`linear_input_dim` is for the implicit user check.. necessary to specify what is expected, if there is a mismatch that points to a bug.

i6_models/parts/frontend/vgg_act.py
albertz: There might have been some confusion around that. I'm not sure if we really came to a final conclusion.

I think it was still unclear whether it is possible or not. But actually when I tested it, there was not really a problem, so the only argument for not having it falls away.

See also the discussion here: https://github.com/rwth-i6/i6_models/pull/22#discussion_r1250523347

tests/test_frontend_vgg.py
albertz: But you have 2 * batch dim here? Already the shape looks wrong.

But if you just want to have random data, why not doing that more directly, using `numpy.random` or so?

E.g. see this function: https://github.com/rwth-i6/returnn/blob/3d148398e2f2177b896a03b9b6eafb1ad9863f16/returnn/tensor/utils.py#L35

Or this: https://github.com/rwth-i6/returnn/blob/3d148398e2f2177b896a03b9b6eafb1ad9863f16/tests/test_TFNetworkLayer.py#L34

tests/test_frontend_vgg.py
albertz: The code looks like you are trying to create a specific sequence mask with some meaning. But I fail to understand the meaning, the logic of the code. But now if I get you right, there is no real meaning at all, it is supposed to be just random. But if that is the case, I would just more explicitly make it random. I would maybe create random seq lens first and then create the seq mask based on that, such that the seq mask is reasonable as we would usually have it.

tests/test_frontend_vgg.py
albertz: I don't think you need that as an argument. You would simply create a minibatch where there are different seqs of different lengths. Then you have some with padding, maybe some without.

i6_models/parts/frontend/vgg_act.py
christophmluscher: with which PT version are you testing? I think we have some limitations that we stick to PT1.13. @Atticus1806 @JackTemaki can you confirm?
if you are using a later version for example PT2, this might be already solved...

tests/test_frontend_vgg.py
christophmluscher: yes, there is no deeper meaning.
ok. I can do that

tests/test_frontend_vgg.py
christophmluscher: I ll check those

i6_models/parts/frontend/vgg_act.py
Atticus1806: So what @albertz posted worked for me in my setup as a fix so it might be that we will be able to also have a mask in inference. The reason this is still "up for discussion" is that i want to wait for @JackTemaki to confirm that this also works for him, since he tested more than me.

i6_models/parts/frontend/vgg_act.py
christophmluscher: the stride controls the mask size if I understand correctly.. so I should only update the mask when I allow setting the stride. since I am following our implementation I do not allow this everywhere.

i6_models/parts/frontend/protocol.py
christophmluscher: I would say for now no..

christophmluscher: we can still later extend this to be a dir with an `__init__`, if necessary without breaking imports.

i6_models/parts/frontend/vgg_act.py
christophmluscher: I prefer keeping this implicit check.

i6_models/parts/frontend/vgg_act.py
Atticus1806: The fix seems to work. So we can remove the optionality.

i6_models/parts/frontend/vgg_act.py
albertz: If the stride is one and same padding, the `_mask_pool` does anyway not have an effect. I would prefer writing code in a way that reduces the amount if implicit assumptions on other parts. Write it in a way that it will just always work. Or at least make a comment here explaining exactly what you say, so that a reader does not need to know it. When reading some code, you should be able to understand the code just by looking at it, without knowing the surrounding code, so the logic should alway be as local as possible, with the least amount of dependencies and assumptions, or any implicit assumptions should be very clearly mentioned, otherwise the code looks wrong.

i6_models/parts/frontend/vgg_act.py
albertz: So I think the opinion of others would be good here.

I think this way is inconsistent to all other PyTorch modules, where you specify the input and output feature dimension (classical example is `nn.Linear`).

christophmluscher: The module has the input and output feature dimension. This would be additionally.

albertz: But output dimensions are usually called `out_features`, but here called `linear_output_dim`.
And no, you do not have any option for the input dimension here?

christophmluscher: I have `in_features`, `linear_input_dim, and `out_features`, so I am confused what you mean?

i6_models/parts/frontend/vgg_act.py
```diff
@@ -0,0 +1,200 @@
...
+        pool1_padding = model_cfg.pool1_padding if model_cfg.pool1_padding is not None else 0
+        pool2_padding = model_cfg.pool2_padding if model_cfg.pool2_padding is not None else 0
+
+        self.include_linear_layer = True if model_cfg.out_features is not None else False
+        assert not (
+            (model_cfg.linear_input_dim is not None) != (model_cfg.out_features is not None)
+        ), "please set input and output dim of the linear layer"
+
+        self.conv1 = nn.Conv2d(
+            in_channels=model_cfg.in_features,
```
albertz, line 113: This is wrong, right?
```suggestion
            in_channels=1,
```

i6_models/parts/frontend/vgg_act.py
albertz: Ah, sorry, I was not looking at the latest version. Yes, you have that.

So then the remaining thing is just `linear_input_dim`. I think it would be good to get the opinions of some others. Whether people actually realize this is nothing to configure but just to perform a sanity check.

i6_models/parts/frontend/vgg_act.py
albertz: `_mask_pool(..., self.conv1.stride, ...)` will just always work, no matter what striding (or no striding) you have in `self.conv1`.

i6_models/parts/frontend/vgg_act.py
```diff
@@ -0,0 +1,200 @@
...
+        T might be reduced to T' or T'' depending on stride of the layers
+
+        stride is only allowed for the pool1 and pool2 operation.
+        other ops do not have stride configurable -> no update of mask sequence required
+
+        :param tensor: input tensor of shape [B,T,F]
+        :param sequence_mask: the sequence mask for the tensor
+        :return: torch.Tensor of shape [B,T",F'] and the shape of the sequence mask
+        """
+        # and add a dim
```
albertz, line 167: I think we should add a check here that the input feature dim is as configured, sth like:
```suggestion
        assert tensor.shape[-1] == self.cfg.in_features
        # and add a dim
```

i6_models/parts/frontend/vgg_act.py
```diff
@@ -0,0 +1,200 @@
...
+        if self.include_linear_layer:
+            self.linear = nn.Linear(
+                in_features=model_cfg.linear_input_dim,
+                out_features=model_cfg.out_features,
+                bias=True,
+            )
+
+    def forward(
+        self, tensor: torch.Tensor, sequence_mask: Optional[torch.Tensor]
+    ) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
```
albertz, line 156:
```suggestion
        self, tensor: torch.Tensor, sequence_mask: torch.Tensor
    ) -> Tuple[torch.Tensor, torch.Tensor]:
```

i6_models/parts/frontend/common.py
```diff
@@ -0,0 +1,42 @@
+from typing import Tuple, Union
+
+import torch
+from torch import nn
+
+
+IntTupleIntType = Union[int, Tuple[int, int]]
+
+
+def _get_padding(input_size: Union[int, Tuple[int, ...]]) -> Union[int, Tuple[int, ...]]:
```
albertz, line 10: I don't think it should be private here.
```suggestion
def get_padding(input_size: Union[int, Tuple[int, ...]]) -> Union[int, Tuple[int, ...]]:
```
But actually, I would try to make the name more expressive.
```suggestion
def get_same_padding(input_size: Union[int, Tuple[int, ...]]) -> Union[int, Tuple[int, ...]]:
```

i6_models/parts/frontend/common.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    if isinstance(input_size, int):
+        return (input_size - 1) // 2
+    elif isinstance(input_size, tuple):
+        return tuple((s - 1) // 2 for s in input_size)
+    else:
+        raise TypeError(f"unexpected size type {type(input_size)}")
+
+
+def _mask_pool(seq_mask: torch.Tensor, kernel_size: int, stride: int, padding: int) -> torch.Tensor:
```
albertz, line 25:
```suggestion
def mask_pool(seq_mask: torch.Tensor, kernel_size: int, stride: int, padding: int) -> torch.Tensor:
```

i6_models/parts/frontend/common.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    seq_mask = seq_mask.float()
+    seq_mask = torch.unsqueeze(seq_mask, 1)  # [B,1,T]
+    seq_mask = nn.functional.max_pool1d(seq_mask, kernel_size, stride, padding)  # [B,1,T']
+    seq_mask = torch.squeeze(seq_mask, 1)  # [B,T']
+    seq_mask = seq_mask.bool()
+    return seq_mask
+
+
+def _get_int_tuple_int(variable: IntTupleIntType, index: int) -> int:
```
albertz, line 41:
```suggestion
def get_int_tuple_int(variable: IntTupleIntType, index: int) -> int:
```

i6_models/parts/frontend/vgg_act.py
christophmluscher: just for my understanding :)
from the code, as is there should be no striding.. but you could set it via `model.conv1.stride` from the outside or so from the outside, correct? that would cover that edge case, correct? are there any other edge cases that you are thinking of?

albertz: I'm not speaking at all about any such edge cases. I mean for a reader of the code, who does not know anything about how `conv1.stride` is configured, the code looks wrong. So to understand this code, you need to have such external knowledge. This is usually a bad idea. You want to reduce interdependencies, to be able to understand code locally, without needing to know anything else. It just makes code easier to understand, and also less buggy overall. E.g. now assume you change the other code, or maybe you copy over this module and take it as a base for a V2 variant, where the `conv1.stride` is configurable and not fixed. You can easily forget then about that you also need to change this other code, which depends on this specific assumption. It's always good to remove interdependencies.

christophmluscher: I see your point.
follow up: what is external knowledge for you? what is locally? everything not in the function? because the config and module should always go together or not?!

albertz: Everything in the function. And there, the closer in number-of-lines-difference, the better. It's a continuum. Going to the other extreme, yes, in the same file is still better than across files. You cannot always avoid it. There always need to be compromises. Sometimes it would have an effect on efficiency and you don't want that.

Btw, here, for efficiency, in your `_mask_pool` function, you might want to add an extra check if `stride == 1` and same padding that then you don't need to do anything.

i6_models/parts/frontend/README.md
```diff
@@ -0,0 +1,5 @@
+# Different front-ends for acoustic encoders
+
+- Two classes are required
+- `Config` class inherits from `ModelConfiguration`
+- Actual Model inherits from `nn.Module` from `torch`
```
albertz, line 5: I don't exactly understand what this readme is telling me.

Two classes for what? For the user? Which two classes?

Why is the comment for `Config` relevant?

Why is the comment for `nn.Module` relevant?

i6_models/parts/frontend/__init__.py
```diff
@@ -0,0 +1 @@
+from .vgg_act import *
```
albertz, line 1: Is this a good idea? I think I would rather not have this.
```suggestion
```

i6_models/parts/frontend/vgg_act.py
```diff
@@ -0,0 +1,234 @@
...
+        out_dim = calculate_output_dim(
+            in_dim=self.cfg.in_features,
+            filter_size=get_int_tuple_int(self.pool1.kernel_size, 1),
+            stride=get_int_tuple_int(self.pool1.stride, 1),
+        )
+        out_dim = calculate_output_dim(
+            in_dim=out_dim,
+            filter_size=get_int_tuple_int(self.pool2.kernel_size, 1),
+            stride=get_int_tuple_int(self.pool2.stride, 1),
+        )
```
albertz, line 232: Here you again have the implicit assumption that there is no striding in the convolutions. My same arguments as earlier would apply, that it is better to avoid as much implicit assumptions as possible.

i6_models/parts/frontend/README.md
christophmluscher: for now this is just stubs. I ll extend

i6_models/parts/frontend/__init__.py
christophmluscher: Why is this not good?

albertz: Why would you want that? There could be other frontends. Do you really want to put all possible frontends into the `i6_models.parts.frontend` namespace?

For the same reason, we don't have `i6_models.assemblies.ConformerEncoderV1`, you need to import it from `i6_models.assemblies.conformer.ConformerEncoderV1`.

christophmluscher: Flexibility for the user.

christophmluscher: But I have no strong feelings on this

i6_models/parts/frontend/common.py
```diff
@@ -0,0 +1,60 @@
...
+    seq_mask = seq_mask.bool()
+    return seq_mask
+
+
+def get_int_tuple_int(variable: IntTupleIntType, index: int) -> int:
+    return variable[index] if isinstance(variable, tuple) else variable
+
+
+def calculate_output_dim(in_dim: int, filter_size: int, stride: int) -> int:
+    if stride == 1:
```
albertz, line 52: You don't really need that. You even do the same check below inside `ceildiv`. But even there, I don't think you need this.

i6_models/parts/frontend/common.py
```diff
@@ -0,0 +1,60 @@
...
+        return in_dim
+
```
albertz, line 54:
```suggestion
```

i6_models/parts/frontend/common.py
```diff
@@ -0,0 +1,60 @@
...
+    def ceildiv(a: int, b: int):
+        if isinstance(b, int) and b == 1:
+            return a
```
albertz, line 57: This was here in case `a` was a dimension tag, to avoid unnecessary logic (creating a new dimension tag).

You don't have that here. So this can simply be removed.

```suggestion
```

christophmluscher:
@JackTemaki @albertz

albertz:
> I would change the code to only support tuples instead of both tuples and integer, this makes the code more readable and more consistent.

I'm not sure I agree. PyTorch itself also supports both. And the more common use case it that the user provides just a single int.

i6_models/parts/frontend/vgg_act.py
```diff
@@ -0,0 +1,260 @@
...
+        activation: activation function at the end
+        out_features: output size of the final linear layer
+    """
+
+    in_features: int
+    conv1_channels: int
+    conv2_channels: int
+    conv3_channels: int
+    conv4_channels: int
+    conv_kernel_size: Tuple
```
JackTemaki, line 45: You could do `Tuple[int, int]` everywhere so that it is clear to the user a two-sized tuple is expected.

i6_models/parts/frontend/common.py
```diff
@@ -0,0 +1,51 @@
...
+    seq_mask = nn.functional.max_pool1d(seq_mask, kernel_size, stride, padding)  # [B,1,T']
+    seq_mask = torch.squeeze(seq_mask, 1)  # [B,T']
+    seq_mask = seq_mask.bool()
+    return seq_mask
+
+
+def calculate_output_dim(in_dim: int, filter_size: int, stride: int) -> int:
+    if stride == 1:
+        return in_dim
+
```
albertz, line 47:
```suggestion
```

i6_models/parts/frontend/common.py
```diff
@@ -0,0 +1,51 @@
...
+    seq_mask = seq_mask.float()
+    seq_mask = torch.unsqueeze(seq_mask, 1)  # [B,1,T]
+    seq_mask = nn.functional.max_pool1d(seq_mask, kernel_size, stride, padding)  # [B,1,T']
+    seq_mask = torch.squeeze(seq_mask, 1)  # [B,T']
+    seq_mask = seq_mask.bool()
+    return seq_mask
+
+
+def calculate_output_dim(in_dim: int, filter_size: int, stride: int) -> int:
```
albertz, line 44: I realize now, you do not have `padding` as an argument here. Why? The output dim depends on the padding as well. What you implemented now is specifically the case for `padding=0` or `padding="valid"`. But I see that you can have potentially other paddings. Then is is wrong here.

i6_models/parts/frontend/vgg_act.py
albertz: Yes, not just "could", `Tuple` should always be specified.

But then, I still would allow `Union[int, Tuple[int, int]]` here, just as PyTorch does for `Conv2d` etc.

i6_models/parts/frontend/common.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    apply strides to the masking
+
+    :param seq_mask: [B,T]
+    :param kernel_size:
+    :param stride:
+    :param padding:
+    :return: [B,T'] using maxpool
+    """
+    if stride == 1 and padding == 1:
```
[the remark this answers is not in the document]
christophmluscher: I dont think so.. the test fails for this... but maybe I have something wrong in the test..
I ll leave this for now..

i6_models/parts/frontend/vgg_act.py
christophmluscher: But then I could just revert to `IntTupleIntType`

On what should we agree? Otherwise, I'll just leave as is :)

i6_models/parts/frontend/common.py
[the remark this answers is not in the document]
albertz: `padding == 1` is wrong in any case. It must depend on the kernel size. E.g. with `kernel_size == 1`, we must have `padding == 0`.

i6_models/parts/frontend/common.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    if isinstance(input_size, int):
+        return (input_size - 1) // 2
+    elif isinstance(input_size, tuple):
+        return tuple((s - 1) // 2 for s in input_size)
+    else:
+        raise TypeError(f"unexpected size type {type(input_size)}")
+
+
+def mask_pool(seq_mask: torch.Tensor, kernel_size: int, stride: int, padding: int) -> torch.Tensor:
```
albertz, line 23: Should use kwargs for this.
```suggestion
def mask_pool(seq_mask: torch.Tensor, *, kernel_size: int, stride: int, padding: int) -> torch.Tensor:
```

i6_models/parts/frontend/common.py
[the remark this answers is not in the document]
albertz: Ah, `padding` is added both left and right? Then it should be:
```suggestion
    if stride == 1 and padding * 2 == kernel_size - 1:
```

JackTemaki (approved):

i6_models/parts/frontend/vgg_act.py
albertz: So what is the state here?

christophmluscher: Only allowing tuple makes the code more readable and allowing int and tuple is more consistent with PT. Since we are not adding any expressive by allowing int, I would leave as is. if this is something that becomes necessary in the future we can make a new PR.

albertz: Yes I know, and I argued for allowing simply `int`, for consistency with PyTorch, and for simplicity for the user. It's very uncommon that you would pass a tuple here. So if you want simplicity, maybe just allow an `int`? But my question was, what do we do now? Was this discussed in our meeting?

christophmluscher: No, it was not discussed. The people present did not have strong feelings in either direction. So I decided to leave it as is.

christophmluscher: see meeting notes sent by Willi

albertz: Ok then.

albertz (approved):

michelwi:
> > I have two questions here,
> > 
> > 1. the param `stride` should be added to the Config, since usually we set stride in Conv2d to apply the time subsampling
> 
> already added but not pushed yet :)

Was this added in the final version of the PR, I don't seem to find it..

PR #18, opened 2023-06-14
Python Packaging

christophmluscher:


pyproject.toml
```diff
@@ -1,3 +1,33 @@
+[build-system]
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
+[project]
+name = "rwth_i6_hlt_pytorch_models"
+version = "0.0.1"
+authors = [
+{ name="Christopy Lüscher", email="luescher@cs.rwth-aachen.de" },
```
michelwi, line 8:
```suggestion
{ name="Christoph Lüscher", email="luescher@cs.rwth-aachen.de" },
```

pyproject.toml
```diff
@@ -1,3 +1,33 @@
...
+{ name="Jingjing Xu", email="jxu@cs.rwth-aachen.de" },
+{ name="Mohammad Zeineldeen", email="zeineldeen@cs.rwth-aachen.de" },
+{ name="Albert Zeyer", email="zeyer@cs.rwth-aachen.de" },
+{ name="Peter Vieting", email="vieting@cs.rwth-aachen.de" },
+{ name="Simon Berger", email="berger@cs.rwth-aachen.de" },
+{ name="Eugen Beck", email="ebeck@apptek.com" },
+{ name="Ping Zheng", email="pzheng@apptek.com" },
+{ name="Wilfried Michel", email="wmichel@apptek.com" },
+]
+description = "A collection of PyTorch NN models and parts. TO be used with the RETURNN framework."
```
michelwi, line 20:
```suggestion
description = "A collection of PyTorch NN models and parts. To be used with the RETURNN framework."
```
I would remove the reference to returnn, i think these are generic pytorch model definitions that can be used with any pytorch based tool, right?

pyproject.toml
```diff
@@ -1,3 +1,33 @@
...
+version = "0.0.1"
+authors = [
+{ name="Christopy Lüscher", email="luescher@cs.rwth-aachen.de" },
+{ name="Nick Rossenbach", email="rossenbach@cs.rwth-aachen.de" },
+{ name="Benedikt Hilmes", email="hilmes@cs.rwth-aachen.de" },
+{ name="Jingjing Xu", email="jxu@cs.rwth-aachen.de" },
+{ name="Mohammad Zeineldeen", email="zeineldeen@cs.rwth-aachen.de" },
+{ name="Albert Zeyer", email="zeyer@cs.rwth-aachen.de" },
+{ name="Peter Vieting", email="vieting@cs.rwth-aachen.de" },
+{ name="Simon Berger", email="berger@cs.rwth-aachen.de" },
```
SimBe195, line 15:
```suggestion
{ name="Simon Berger", email="sberger@cs.rwth-aachen.de" },
```

pyproject.toml
christophmluscher: thx :D

michelwi (approved):

SimBe195 (approved):

Atticus1806 (approved):

vieting (approved):

pyproject.toml
```diff
@@ -1,3 +1,33 @@
+[build-system]
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
+[project]
+name = "rwth_i6_hlt_pytorch_models"
```
albertz, line 5: I think we did not talk about the name yet. This would be the PIP package name then, right? Ideally the PIP package name should be the same as the import name. So in setups, you would do `import rwth_i6_hlt_pytorch_models`.

I think the name is too complicated. I would suggest to just use `i6_models` here, which is also not taken by any other project on PyPI yet. Then we can stay with the original `import i6_models`.

```suggestion
name = "i6_models"
```

christophmluscher: yes I agree

JackTemaki (approved):

PR #19, opened 2023-06-19
Add BlstmEncoder

JackTemaki:
- Adds a BlstmEncoder part for multi-layered BLSTMs with dropout
 - Adds tests for batching/scripting verification and ONNX export

### DIFF
--- a/i6_models/parts/blstm.py
+++ b/i6_models/parts/blstm.py
@@ -50,7 +50,7 @@ class BlstmEncoderV1(torch.nn.Module):
             hidden_size=config.hidden_dim,
             bidirectional=True,
             num_layers=config.num_layers,
-            batch_first=False,
+            batch_first=True,
             dropout=self.dropout,
         )
 
--- a/i6_models/parts/blstm.py
+++ b/i6_models/parts/blstm.py
@@ -1,0 +1,76 @@
+from dataclasses import dataclass
+import torch
+from torch import nn
+
+from i6_models.config import ModelConfiguration
+
+
+@dataclass()
+class BlstmEncoderConfig(ModelConfiguration):
+    num_layers: int
+    """number of bi-directional LSTM layers, minimum 2"""
+    input_dimension: int
+    """input dimension size"""
+    hidden_dimension: int
+    """hidden dimension of one direction of LSTM, the total output size is twice of this"""
+    dropout: float
+    """nn.LSTM supports an internal Dropout layer"""
+    enforce_sorted: bool = True
+    """
+        keep activated for ONNX-Export, requires that the lengths are sorted decreasing from longest
+
+        Sorting can performed using something like:    
+
+            audio_features_len, indices = torch.sort(audio_features_len, descending=True)
+            audio_features = audio_features[indices, :, :]
+            labels = labels[indices, :]
+            labels_len = labels_len[indices]
+    """
+
+
+class BlstmEncoder(torch.nn.Module):
+    """
+    Simple multi-layer BLSTM model including dropout, batch-first variant,
+    hardcoded to use B,T,F input
+
+    supports: TorchScript, ONNX-export
+    """
+
+    def __init__(self, config: BlstmEncoderConfig):
+        """
+        :param config: configuration object
+        """
+        super().__init__()
+        self.dropout = config.dropout
+        self.enforce_sorted = config.enforce_sorted
+        self.blstm_stack = nn.LSTM(
+            input_size=config.input_dimension,
+            hidden_size=config.hidden_dimension,
+            bidirectional=True,
+            num_layers=config.num_layers,
+            batch_first=False,
+            dropout=self.dropout,
+        )
+
+    def forward(self, x: torch.Tensor, seq_len: torch.Tensor) -> torch.Tensor:
+        """
+        :param x: [B, T, input_dimension]
+        :param seq_len: [B], should be on CPU for Script/Trace mode
+        :return [B, T, 2 * hidden_dimension]
+        """
+        if not torch.jit.is_scripting() and not torch.jit.is_tracing():
+            # during graph mode we have to assume all Tensors are on the correct device,
+            # otherwise move lengths to the CPU if they are on GPU
+            if seq_len.get_device() >= 0:
+                seq_len = seq_len.cpu()
+
+        blstm_packed_in = nn.utils.rnn.pack_padded_sequence(
+            input=x,
+            lengths=seq_len,
+            enforce_sorted=self.enforce_sorted,
+            batch_first=True,
+        )
+        blstm_out, _ = self.blstm_stack(blstm_packed_in)
+        blstm_out, _ = nn.utils.rnn.pad_packed_sequence(blstm_out, padding_value=0.0, batch_first=True)
+
+        return blstm_out
--- a/i6_models/parts/blstm.py
+++ b/i6_models/parts/blstm.py
@@ -7,25 +7,27 @@ from i6_models.config import ModelConfiguration
 
 @dataclass()
 class BlstmEncoderConfig(ModelConfiguration):
+    """
+    Attributes:
+        num_layers: number of bi-directional LSTM layers, minimum 2
+        input_dim: input dimension size
+        hidden_dim: hidden dimension of one direction of LSTM, the total output size is twice of this
+        dropout: nn.LSTM supports internal Dropout applied between each layer of BLSTM (but not on input/output)
+        enforce_sorted: keep activated for ONNX-Export, requires that the lengths are sorted decreasing from longest
+
+            Sorting can for example be performed independent of the ONNX export in e.g. train_step:
+
+                audio_features_len, indices = torch.sort(audio_features_len, descending=True)
+                audio_features = audio_features[indices, :, :]
+                labels = labels[indices, :]
+                labels_len = labels_len[indices]
+    """
+
     num_layers: int
-    """number of bi-directional LSTM layers, minimum 2"""
-    input_dimension: int
-    """input dimension size"""
-    hidden_dimension: int
-    """hidden dimension of one direction of LSTM, the total output size is twice of this"""
+    input_dim: int
+    hidden_dim: int
     dropout: float
-    """nn.LSTM supports an internal Dropout layer"""
     enforce_sorted: bool = True
-    """
-        keep activated for ONNX-Export, requires that the lengths are sorted decreasing from longest
-
-        Sorting can performed using something like:    
-
-            audio_features_len, indices = torch.sort(audio_features_len, descending=True)
-            audio_features = audio_features[indices, :, :]
-            labels = labels[indices, :]
-            labels_len = labels_len[indices]
-    """
 
 
 class BlstmEncoder(torch.nn.Module):
@@ -44,8 +46,8 @@ class BlstmEncoder(torch.nn.Module):
         self.dropout = config.dropout
         self.enforce_sorted = config.enforce_sorted
         self.blstm_stack = nn.LSTM(
-            input_size=config.input_dimension,
-            hidden_size=config.hidden_dimension,
+            input_size=config.input_dim,
+            hidden_size=config.hidden_dim,
             bidirectional=True,
             num_layers=config.num_layers,
             batch_first=False,
@@ -54,9 +56,9 @@ class BlstmEncoder(torch.nn.Module):
 
     def forward(self, x: torch.Tensor, seq_len: torch.Tensor) -> torch.Tensor:
         """
-        :param x: [B, T, input_dimension]
+        :param x: [B, T, input_dim]
         :param seq_len: [B], should be on CPU for Script/Trace mode
-        :return [B, T, 2 * hidden_dimension]
+        :return [B, T, 2 * hidden_dim]
         """
         if not torch.jit.is_scripting() and not torch.jit.is_tracing():
             # during graph mode we have to assume all Tensors are on the correct device,
--- a/i6_models/parts/blstm.py
+++ b/i6_models/parts/blstm.py
@@ -6,7 +6,7 @@ from i6_models.config import ModelConfiguration
 
 
 @dataclass()
-class BlstmEncoderConfig(ModelConfiguration):
+class BlstmEncoderV1Config(ModelConfiguration):
     """
     Attributes:
         num_layers: number of bi-directional LSTM layers, minimum 2
@@ -30,7 +30,7 @@ class BlstmEncoderConfig(ModelConfiguration):
     enforce_sorted: bool = True
 
 
-class BlstmEncoder(torch.nn.Module):
+class BlstmEncoderV1(torch.nn.Module):
     """
     Simple multi-layer BLSTM model including dropout, batch-first variant,
     hardcoded to use B,T,F input
@@ -38,7 +38,7 @@ class BlstmEncoder(torch.nn.Module):
     supports: TorchScript, ONNX-export
     """
 
-    def __init__(self, config: BlstmEncoderConfig):
+    def __init__(self, config: BlstmEncoderV1Config):
         """
         :param config: configuration object
         """
--- a/i6_models/parts/blstm.py
+++ b/i6_models/parts/blstm.py
@@ -5,7 +5,7 @@ from torch import nn
 from i6_models.config import ModelConfiguration
 
 
-@dataclass()
+@dataclass
 class BlstmEncoderV1Config(ModelConfiguration):
     """
     Attributes:
@@ -16,7 +16,7 @@ class BlstmEncoderV1Config(ModelConfiguration):
         enforce_sorted:
             True: expects that sequences are sorted by sequence length in decreasing order.
                 Will not do any sorting.
-                This is required for ONNX-Export.
+                This is required for ONNX-Export, and thus the recommended setting.
             False: no expectation.
                 It will internally enforce that they are sorted
                 and undo the reordering at the output.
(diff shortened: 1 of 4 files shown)


i6_models/parts/blstm.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        :param x: [B, T, input_dimension]
+        :param seq_len: [B], should be on CPU for Script/Trace mode
+        :return [B, T, 2 * hidden_dimension]
+        """
+        if not torch.jit.is_scripting() and not torch.jit.is_tracing():
+            # during graph mode we have to assume all Tensors are on the correct device,
+            # otherwise move lengths to the CPU if they are on GPU
+            if seq_len.get_device() >= 0:
+                seq_len = seq_len.cpu()
```
albertz, line 65: Aren't we anyway always expecting this?

i6_models/parts/blstm.py
```diff
@@ -0,0 +1,76 @@
...
+from torch import nn
+
+from i6_models.config import ModelConfiguration
+
+
+@dataclass()
+class BlstmEncoderConfig(ModelConfiguration):
+    num_layers: int
+    """number of bi-directional LSTM layers, minimum 2"""
+    input_dimension: int
```
albertz, line 12: We have this somewhat inconsistent throughout other parts in the code. E.g. in `ConformerPositionwiseFeedForwardV1Config` we have `input_dim` and `hidden_dim`. Shouldn't we try to make it consistent?

i6_models/parts/blstm.py
```diff
@@ -0,0 +1,76 @@
...
+    """input dimension size"""
+    hidden_dimension: int
+    """hidden dimension of one direction of LSTM, the total output size is twice of this"""
+    dropout: float
+    """nn.LSTM supports an internal Dropout layer"""
+    enforce_sorted: bool = True
```
albertz, line 18: I think a more descriptive name for this would be better. E.g. what this does, what his is about (seq lengths).

i6_models/parts/blstm.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+        keep activated for ONNX-Export, requires that the lengths are sorted decreasing from longest
+
+        Sorting can performed using something like:    
+
+            audio_features_len, indices = torch.sort(audio_features_len, descending=True)
+            audio_features = audio_features[indices, :, :]
+            labels = labels[indices, :]
+            labels_len = labels_len[indices]
```
albertz, line 27: The comment is a bit misleading. Maybe also because `enforce_sorted` in `pack_padded_sequence` is a bit misleading. Actually, what it means:

`enforce_sorted=True`: It does nothing. It expects that the input is sorted. But actually, I think this is also not really true anymore? E.g. see https://github.com/pytorch/pytorch/issues/23079. The question is, what code is really needing this? `pack_padded_sequence` or `LSTM` or sth else?

`enforce_sorted=False`: This will actually automatically sort the sequence internally in `pack_padded_sequence`. So it will make sure (it enforces) that the sequences are sorted. I find the name of this option a bit misleading. The sorting code is basically exactly what you post here, using `torch.sort`.

When you say, `enforce_sorted=True` is needed for ONNX, what this implies is that the sorting code using `torch.sort` is not supported by ONNX? So it means that your code here would also not work. Or vice-versa, if this code works, you could simply also do `enforce_sorted=False`, which does exactly the same.

JackTemaki: The comment is not misleading at all to me, but I can reformulate it if it is unclear to you.

`keep activated for ONNX-Export` -> you need to set it to true if you want to export this model via ONNX, otherwise it will not work (I tested it)

`requires that the lengths are sorted decreasing from longest` -> if true, you have to care about the sorting yourself, which is also clear and a true statement.

The  example code can be part of the train_step. That way you have a consistent model for training and export, and you also save operations to sort your data only once (for the model and the loss computation) and not twice.

i6_models/parts/blstm.py
```diff
@@ -0,0 +1,76 @@
...
+import torch
+from torch import nn
+
+from i6_models.config import ModelConfiguration
+
+
+@dataclass()
+class BlstmEncoderConfig(ModelConfiguration):
+    num_layers: int
+    """number of bi-directional LSTM layers, minimum 2"""
```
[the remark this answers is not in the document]
JackTemaki: Benedikt said the same, I was not sure as it did not become part of the contributions. Will change.

i6_models/parts/blstm.py
JackTemaki: In RETURNN yes, but I did not want to have that expectation here (I am also using the model in a different context myself)

i6_models/parts/blstm.py
albertz: As explained, the example code is exactly what happens when you do `enforce_sorted=False`. So, if you say that `enforce_sorted=False` does not work, then your example code also will not work. And that is what I mean by misleading. It sounds like the example code would be a solution here. But it is exactly the code which is used when you do `enforce_sorted=False`, which apparently does not work, as you say.

albertz: Or do you mean that this example code would not be part of the export?

albertz: Also, another thing what I mean by misleading: According to https://github.com/pytorch/pytorch/issues/23079, the sorting should not be needed at all.

i6_models/parts/blstm.py
albertz: But this is somewhat relevant, if we maybe want to have that expectation always for all modules here in i6_models. 

That would simplify the code in many places, basically all places which involves seq lens, because as far as I understood, most (all?) typical code would always require seq lens on CPU. So if you don't have this expectation, you would need such code in many places.

And this can potentially be slow. If the seq len is really on GPU, it involves a GPU->CPU copy, which can be very slow.

i6_models/parts/blstm.py
JackTemaki: I would rather keep the actual name of the parameter.

i6_models/parts/blstm.py
JackTemaki: > Or do you mean that this example code would not be part of the export?

Correct, I made the comment more explicit about that.

>  the sorting should not be needed at all.

Feel free to debug and fix it, but I think this is unnecessary effort and so I will not participate in this.

i6_models/parts/blstm.py
JackTemaki: I have the seq_len always also on GPU, so that I can always assume all my tensors are on the target device. I also have setups that perform operations on the seq_len, e.g. within feature extraction or the conversion from seq_len to a mask, and I want to do this on GPU.

Still we can discuss to have the seq_len on CPU assumptions here in i6_models, I can find other solutions for me to deal with that.

i6_models/parts/blstm.py
```diff
@@ -0,0 +1,78 @@
...
+                seq_len = seq_len.cpu()
+
+        blstm_packed_in = nn.utils.rnn.pack_padded_sequence(
+            input=x,
+            lengths=seq_len,
+            enforce_sorted=self.enforce_sorted,
+            batch_first=True,
+        )
+        blstm_out, _ = self.blstm_stack(blstm_packed_in)
+        blstm_out, _ = nn.utils.rnn.pad_packed_sequence(blstm_out, padding_value=0.0, batch_first=True)
```
Atticus1806, line 76: Would it make sense to make the padding value configurable?

tests/test_blstm.py
```diff
@@ -0,0 +1,60 @@
+import onnxruntime as ort
+import tempfile
+import torch
+from torch.onnx import export as export_onnx
+
+from i6_models.parts.blstm import BlstmEncoder, BlstmEncoderConfig
+
+
+def test_blstm_onnx_export():
```
Atticus1806, line 9: This makes me wonder: Do we either want some example export somewhere? Or maybe even a Blueprint for copy pasting? Or do we just use this function (or part of it) as blueprint and maybe document it?

i6_models/parts/blstm.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+
+    num_layers: int
+    input_dim: int
+    hidden_dim: int
+    dropout: float
+    enforce_sorted: bool = True
+
+
+class BlstmEncoder(torch.nn.Module):
```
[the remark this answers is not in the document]
albertz: I thought that we just always want to have everything versionized?

i6_models/parts/blstm.py
albertz: I don't think this is necessary. This is for the output. The padding value ideally should not be used at all by any following code.

i6_models/parts/blstm.py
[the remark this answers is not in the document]
JackTemaki: Yes, will fix.

tests/test_blstm.py
JackTemaki: I would say not for now, we can unify this later on.

Atticus1806 (approved):

christophmluscher (approved):

i6_models/parts/blstm.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        super().__init__()
+        self.dropout = config.dropout
+        self.enforce_sorted = config.enforce_sorted
+        self.blstm_stack = nn.LSTM(
+            input_size=config.input_dim,
+            hidden_size=config.hidden_dim,
+            bidirectional=True,
+            num_layers=config.num_layers,
+            batch_first=False,
```
Atticus1806, line 53: I think  this needs to be True.

SimBe195 (approved):

i6_models/parts/blstm.py
```diff
@@ -0,0 +1,78 @@
...
+
+@dataclass()
+class BlstmEncoderV1Config(ModelConfiguration):
+    """
+    Attributes:
+        num_layers: number of bi-directional LSTM layers, minimum 2
+        input_dim: input dimension size
+        hidden_dim: hidden dimension of one direction of LSTM, the total output size is twice of this
+        dropout: nn.LSTM supports internal Dropout applied between each layer of BLSTM (but not on input/output)
+        enforce_sorted: keep activated for ONNX-Export, requires that the lengths are sorted decreasing from longest
```
albertz, line 16: I still think this description is a bit misleading or incomplete. `enforce_sorted=True` actually means that the model expects that the sequences are sorted, `enforce_sorted=False` means that it will enforce internally that sequences are sorted and it is not needed that they are sorted beforehand.

```suggestion
        enforce_sorted:
            True: expects that sequences are sorted by sequence length in decreasing order.
                Will not do any sorting.
                This is required for ONNX-Export.
            False: no expectation.
                It will internally enforce that they are sorted
                and undo the reordering at the output.        
```

i6_models/parts/blstm.py
```diff
@@ -0,0 +1,78 @@
...
+                audio_features = audio_features[indices, :, :]
+                labels = labels[indices, :]
+                labels_len = labels_len[indices]
+    """
+
+    num_layers: int
+    input_dim: int
+    hidden_dim: int
+    dropout: float
+    enforce_sorted: bool = True
```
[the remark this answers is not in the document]
JackTemaki: Yes, this is a good default. Any model should be ONNX exportable per default. Also I still think it is better (perfomancewise) to sort the whole batch in the beginning, instead of doing it separately for the RNN stack and the loss computation.

And what does this have to do with RETURNN itself? This is purely on the side of the user if you want to do sorting by hand or not.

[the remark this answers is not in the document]
albertz: As a user, I would expect that I could just use default settings and don't care about it. This would not be the case here.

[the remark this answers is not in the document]
JackTemaki: > Also, I don't think that ONNX exportability is such a high priority. ONNX will probably be replaced by sth else in the future anyway. And ONNX export is not needed for most cases except of RASR. And this sounds like it is really just a workaround for a current version of ONNX which might already be fixed next week or so. The other argument about performance is more relevant here.

This model is for doing experiments next week, not at some undefined future, at which we might decide to drop BLSTM in favor of other models or have a different implementations. I understood my task as: I should push a PyTorch BLSTM model that works with RASR in its current form, and was tested to give accurate WERs compared to the TF BLSTM within a RASR setup. And for this the only currently accepted way is via ONNX (making it the top priority).

As the users are from our chair (or affiliates to that), they expect that if they write a config with this model it can be exported for RASR.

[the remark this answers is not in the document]
albertz: When you want to push sth here in this repo, I think it is relevant that it is in good shape, and not just useful for the experiments you do next week.

What happens when you have `enforce_sorted=True` but the seqs are actually not sorted?

Would it throw an error then? If that is the case, then this still maybe ok. But I would still argue that this is unexpected behavior for the common user, who likely would expect to be able to use sth like a `BlstmEncoder` just as-is.

Or would it perform some incorrect computation without throwing an error? If that is the case, then definitely `enforce_sorted=True` cannot be the default. That would be very dangerous, lead to hard-to-debug behavior.

If you don't know, this is sth important which should be checked first.

Or we just make the default as `enforce_sorted=False`. I think this is a more reasonable default. The average user can just use it as-is and don't care about details. And in case of ONNX export, the user would get an error and thus notice about the ONNX specific problem.

Or we would just not have a default if we cannot agree here.

I think I made my opinion clear here and probably cannot add much more to that. So the input of some other people here would be nice. @michelwi @christophmluscher @Marvin84 or so?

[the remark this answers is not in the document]
JackTemaki: > When you want to push sth here in this repo, I think it is relevant that it is in good shape, and not just useful for the experiments you do next week.

Our discussion is not about good shape, but about usage preference. I am just assuming that more than 50% of the cases somebody uses this encoder it will be used for RASR. 

> Would it throw an error then? 

Yes, it throws an error:

```
RuntimeError: `lengths` array must be sorted in decreasing order when `enforce_sorted` is True. You can pass `enforce_sorted=False` to pack_padded_sequence and/or pack_sequence to sidestep this requirement if you do not need ONNX exportability.
```

> Or we would just not have a default if we cannot agree here.

This is a valid option.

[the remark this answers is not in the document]
Atticus1806: I think this is something we should discuss today, since I think this trade-off will be a discussion in future .
On the other side: What is the use-case for this model? For now I know of Hybrid Baseline but do we actually use it for other experiments? Hybrid requires this export from my understanding so I would vote pro `True` default, but if there are other models which do not use it I would prefer `False` or nothing.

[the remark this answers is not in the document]
SimBe195: > Or we would just not have a default if we cannot agree here.
Yes, I believe we decided that we want to only want to apply defaults very sparsely when there is a good reason and it's universally agreed upon. So if this one is controversial I would just not set a default value.

i6_models/parts/blstm.py
```diff
@@ -0,0 +1,84 @@
...
```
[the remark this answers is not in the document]
Atticus1806: I am not sure about this. The parameter internally is named enforce_sorted. This might cause confusion.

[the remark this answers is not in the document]
JackTemaki: While the name is certainly not optimal, I would still stick to it because of what Benedikt says, but also because the RuntimeError you get if you do it wrong clearly mentions `enforce_sorted` as parameter name.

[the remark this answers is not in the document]
SimBe195: Okay, that's a good reason, then I'm fine with the name.

i6_models/parts/blstm.py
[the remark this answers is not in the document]
curufinwe: Result of offline discussion: don't set default, but write in the docstring that the recommended value is True.

i6_models/parts/blstm.py
```diff
@@ -0,0 +1,84 @@
+from dataclasses import dataclass
+import torch
+from torch import nn
+
+from i6_models.config import ModelConfiguration
+
+
+@dataclass()
```
albertz, line 8: Consistency:
```suggestion
@dataclass
```

i6_models/parts/blstm.py
albertz: Was this discussed?

i6_models/parts/blstm.py
albertz, line 1: Shouldn't the filename be versionized? `blstm_v1.py`?

JackTemaki: No, also if needed it can be moved to blstm/blstm_v1.py later if needed without altering the import location.

i6_models/parts/blstm.py
```diff
@@ -0,0 +1,84 @@
...
+from torch import nn
+
+from i6_models.config import ModelConfiguration
+
+
+@dataclass
+class BlstmEncoderV1Config(ModelConfiguration):
+    """
+    Attributes:
+        num_layers: number of bi-directional LSTM layers, minimum 2
```
vieting, line 12: nitpick: you say minimum 2 here, but I don't see this being enforced anywhere and I'm also not sure if it couldn't just as well work with 1.

JackTemaki: Ops, that was from an older version.

i6_models/parts/blstm.py
albertz: Was this discussed now? What was the result of the discussion? I don't see this in the meeting notes.

PR #21, opened 2023-06-22
ModuleFactory

Atticus1806:
As discussed today moved out of #8

### DIFF
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -66,7 +66,7 @@ class ModuleFactoryV1(Generic[ConfigType, ModuleType]):
     Also provides a function to construct the corresponding object through this dataclass
     """
 
-    module_class: Callable[[ConfigType], ModuleType]
+    module_class: Type[ModuleType]
     cfg: ConfigType
 
     def __call__(self) -> ModuleType:
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -19,7 +19,7 @@ from dataclasses import dataclass, fields
 import typeguard
 from torch import nn
 from typing import Generic, TypeVar, Type
-from inspect import signature
+import inspect
 
 
 @dataclass
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -17,6 +17,8 @@ Similar approach as done in Fairseq: https://github.com/facebookresearch/fairseq
 from __future__ import annotations
 from dataclasses import dataclass, fields
 import typeguard
+from torch import nn
+from typing import TypeVar, Type
 
 
 @dataclass
@@ -51,3 +53,22 @@ class ModelConfiguration:
 
     def __post_init__(self) -> None:
         self._validate_types()
+
+
+ConfigType = TypeVar("ConfigType", bound=ModelConfiguration)
+ModuleType = TypeVar("ModuleType", bound=nn.Module)
+
+
+@dataclass
+class SubassemblyFactory:
+    """
+    Dataclass for a combination of a Subassembly/Part and the corresponding configuration.
+    Also provides a function to construct the corresponding object through this dataclass
+    """
+
+    module_class: Type[ModuleType]
+    cfg: ConfigType
+
+    def construct(self) -> ModuleType:
+        """Constructs an instance of the given module class"""
+        return self.module_class(self.cfg)
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -69,6 +69,6 @@ class SubassemblyFactory:
     module_class: Type[ModuleType]
     cfg: ConfigType
 
-    def construct(self) -> ModuleType:
+    def __call__(self) -> ModuleType:
         """Constructs an instance of the given module class"""
         return self.module_class(self.cfg)
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -60,7 +60,7 @@ ModuleType = TypeVar("ModuleType", bound=nn.Module)
 
 
 @dataclass
-class SubassemblyFactory(Generic[ConfigType, ModuleType]):
+class ModuleFactory(Generic[ConfigType, ModuleType]):
     """
     Dataclass for a combination of a Subassembly/Part and the corresponding configuration.
     Also provides a function to construct the corresponding object through this dataclass
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -66,7 +66,7 @@ class SubassemblyFactory(Generic[ConfigType, ModuleType]):
     Also provides a function to construct the corresponding object through this dataclass
     """
 
-    module_class: Type[ModuleType]
+    module_class: Callable[[ConfigType], ModuleType]
     cfg: ConfigType
 
     def __call__(self) -> ModuleType:
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -60,7 +60,7 @@ ModuleType = TypeVar("ModuleType", bound=nn.Module)
 
 
 @dataclass
-class ModuleFactory(Generic[ConfigType, ModuleType]):
+class ModuleFactoryV1(Generic[ConfigType, ModuleType]):
     """
     Dataclass for a combination of a Subassembly/Part and the corresponding configuration.
     Also provides a function to construct the corresponding object through this dataclass
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -60,7 +60,7 @@ ModuleType = TypeVar("ModuleType", bound=nn.Module)
 
 
 @dataclass
-class SubassemblyFactory:
+class SubassemblyFactory(Generic[ConfigType, ModuleType]):
     """
     Dataclass for a combination of a Subassembly/Part and the corresponding configuration.
     Also provides a function to construct the corresponding object through this dataclass
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -75,4 +75,7 @@ class ModuleFactoryV1(Generic[ConfigType, ModuleType]):
         return self.module_class(self.cfg)
 
     def __post_init__(self) -> None:
-        assert signature(self.module_class).parameters["cfg"].annotation == type(self.cfg)
+        parameters = signature(self.module_class).parameters.values()
+        assert len(parameters) == 1
+        cfg_parameter = next(iter(parameters))
+        typeguard.check_type(self.cfg, cfg_parameter.annotation)
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -75,7 +75,16 @@ class ModuleFactoryV1(Generic[ConfigType, ModuleType]):
         return self.module_class(self.cfg)
 
     def __post_init__(self) -> None:
-        parameters = signature(self.module_class).parameters.values()
-        assert len(parameters) == 1
-        cfg_parameter = next(iter(parameters))
-        typeguard.check_type(self.cfg, cfg_parameter.annotation)
+        # Check typing of module_class and cfg, i.e. make sure that "self.module_class(self.cfg)" is a valid call.
+        parameters = inspect.signature(self.module_class).parameters.values()
+        assert len(parameters) >= 1
+        parameter_iter = iter(parameters)
+
+        # 1. Check that the first parameter is either not annotated or the annotation matches the type of self.cfg
+        cfg_parameter = next(parameter_iter)
+        if cfg_parameter.annotation is not inspect.Parameter.empty:
+            typeguard.check_type(self.cfg, cfg_parameter.annotation)
+            
+        # 2. Check that all other parameters have default values
+        for parameter in parameter_iter:
+            assert parameter.default is not inspect.Parameter.empty
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 from dataclasses import dataclass, fields
 import typeguard
 from torch import nn
-from typing import TypeVar, Type
+from typing import Callable, Generic, TypeVar
 
 
 @dataclass
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -84,7 +84,7 @@ class ModuleFactoryV1(Generic[ConfigType, ModuleType]):
         cfg_parameter = next(parameter_iter)
         if cfg_parameter.annotation is not inspect.Parameter.empty:
             typeguard.check_type(self.cfg, cfg_parameter.annotation)
-            
+
         # 2. Check that all other parameters have default values
         for parameter in parameter_iter:
             assert parameter.default is not inspect.Parameter.empty
--- a/i6_models/config.py
+++ b/i6_models/config.py
@@ -18,7 +18,8 @@ from __future__ import annotations
 from dataclasses import dataclass, fields
 import typeguard
 from torch import nn
-from typing import Callable, Generic, TypeVar
+from typing import Generic, TypeVar, Type
+from inspect import signature
 
 
 @dataclass
@@ -72,3 +73,6 @@ class ModuleFactoryV1(Generic[ConfigType, ModuleType]):
     def __call__(self) -> ModuleType:
         """Constructs an instance of the given module class"""
         return self.module_class(self.cfg)
+
+    def __post_init__(self) -> None:
+        assert signature(self.module_class).parameters["cfg"].annotation == type(self.cfg)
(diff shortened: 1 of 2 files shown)


i6_models/config.py
```diff
@@ -17,6 +17,8 @@ class ExampleModule(Module):
 from __future__ import annotations
 from dataclasses import dataclass, fields
 import typeguard
+from torch import nn
+from typing import TypeVar, Type
```
SimBe195, line 21:
```suggestion
from typing import Callable, Generic, TypeVar
```

i6_models/config.py
```diff
@@ -51,3 +53,22 @@ def _validate_types(self) -> None:
...
     def __post_init__(self) -> None:
         self._validate_types()
+
+
+ConfigType = TypeVar("ConfigType", bound=ModelConfiguration)
+ModuleType = TypeVar("ModuleType", bound=nn.Module)
+
+
+@dataclass
+class SubassemblyFactory:
```
SimBe195, line 63:
```suggestion
class SubassemblyFactory(Generic[ConfigType, ModuleType]):
```

i6_models/config.py
```diff
@@ -51,3 +53,22 @@ def _validate_types(self) -> None:
...
+    """
+    Dataclass for a combination of a Subassembly/Part and the corresponding configuration.
+    Also provides a function to construct the corresponding object through this dataclass
+    """
+
+    module_class: Type[ModuleType]
```
SimBe195, line 69:
```suggestion
    module_class: Callable[[ConfigType], ModuleType]
```

SimBe195 (changes requested):

i6_models/config.py
Atticus1806: I am not sure about this. I thought `Type` was meant for class objects.

i6_models/config.py
Atticus1806: We can add this, but I have to admit even after reading up on it I dont fully understand the benefits over just having the parameters typed. Also is the order within the generic relevant? Then the types would need to be swapped right?

i6_models/config.py
SimBe195: Yes but each class is also a callable due to its constructor. This typehint ensures that the constructor takes a single parameter of type `ConfigType` so that calling `self.module_class(self.cfg)` is valid.

i6_models/config.py
SimBe195: If you want to use generic types inside a class you usually declare it as a `Generic` depending on these generic types. Having this "connects" the generic types used inside. For example `ModuleType` used in `module_class` is the same as the `ModuleType` returned by the `__call__` function. So for example if you set `module_class = ConformerEncoderV1` somewhere type checkers will know that the `__call__` method will return a `ConformerEncoderV1` instance.

> Also is the order within the generic relevant?

No, I don't believe the order is relevant.

michelwi: After reading a bit myself, I think the advantage is that we could specify which sort of subassemblies we would like to allow. e.g.

```
# define a metaclass for Norms
@dataclass
class NormAssemblyConfiguration(ModelConfiguration):
  pass

class NormAssembly(torch.nn.Module):
  pass

# define some norms
@dataclass
class BatchNormAssemblyConfiguration(NormAssemblyConfiguration):
  pass

class BatchNormAssembly(NormAssembly):
  pass

@dataclass
class LayerNormAssemblyConfiguration(NormAssemblyConfiguration):
  pass

class LayerNormAssembly(NormAssembly):
  pass

# some non norm related Assembly
class UnrelatedAssemblyConfiguration(ModelConfiguration):
  pass

class UnrelatedAssembly(torch.nn.Module):
  pass

# define a custom AssemblyConfig, that should accept any norm, but only norms
@dataclass
class MyAssemblyConfig(ModelConfiguration):
   norm: SubassemblyFactory[NormAssemblyConfiguration, NormAssembly]

# test the new config class
c1 = MyAssemblyConfig(
    norm=SubassemblyFactory(
        module_class=BatchNormAssembly,
        cfg=BatchNormAssemblyConfiguration()
    )
)  # correct typing

c2 = MyAssemblyConfig(
    norm=SubassemblyFactory(
        module_class=UnrelatedAssembly,
        cfg=UnrelatedAssemblyConfiguration()
    )
)  # valid python, but should give typing error
```
see my test here
image


> Also is the order within the generic relevant?

you should match the order when typing your own configs. In the example above
```
norm: SubassemblyFactory[NormAssemblyConfiguration, NormAssembly]  # same order
norm: SubassemblyFactory[NormAssembly, NormAssemblyConfiguration]  # wrong order
```

> Then the types would need to be swapped right?

this is a matter of taste. I would agree, but there is no technical reason the order here should match the order of arguments in the SubassemblyFactory

michelwi: > If you want to use generic types inside a class you usually declare it as a Generic depending on these generic types. Having this "connects" the generic types used inside. 

I thought the fact that `ModuleType` is a `TypeVar` already did this, but apparently not.
image
(Type variable "ModuleType" has no meaning in this context)

However, if you move away from dataclasses, then you do not need to make the class a Generic to use TypeVars to enforce consistency:
image
("LayerNormAssemblyConfiguration" is incompatible with "BatchNormAssemblyConfiguration")

SimBe195: > However, if you move away from dataclasses, then you do not need to make the class a Generic to use TypeVars to enforce consistency

This example doesn't work however when you add `def __call__(self) -> ModuleType:` because the system doesn't know that the `ModuleType` from `__init__` is the same `ModuleType` as in `__call__` if you don't inherit from `Generic`.

michelwi: > This example doesn't work however when  [...]

Indeed it does not. 😅 
Thanks for the clarification and counterexample.

SimBe195 (approved):

michelwi (approved):

i6_models/config.py
```diff
@@ -51,3 +53,22 @@ def _validate_types(self) -> None:
...
+
+
+@dataclass
+class SubassemblyFactory(Generic[ConfigType, ModuleType]):
+    """
+    Dataclass for a combination of a Subassembly/Part and the corresponding configuration.
+    Also provides a function to construct the corresponding object through this dataclass
+    """
+
+    module_class: Callable[[ConfigType], ModuleType]
```
albertz, line 69: I think this should be:
```suggestion
    module_class: Type[ModuleType]
```
You want to be able to have code like:
```python
assert isinstance(obj, opt.module_class)
```
So it would be good to have that guarantee that this `module_class` is really the class, and not any random callable.

i6_models/config.py
```diff
@@ -51,3 +53,22 @@ def _validate_types(self) -> None:
...
     def __post_init__(self) -> None:
         self._validate_types()
+
+
+ConfigType = TypeVar("ConfigType", bound=ModelConfiguration)
+ModuleType = TypeVar("ModuleType", bound=nn.Module)
+
+
+@dataclass
+class SubassemblyFactory(Generic[ConfigType, ModuleType]):
```
albertz, line 63: Should this be versionized?
```suggestion
class SubassemblyFactoryV1(Generic[ConfigType, ModuleType]):
```

albertz, line 63: I don't like the name so much. This is not really just for subassemblies. This can (and probably will) be used for any of our modules which have this generic interface where `__init__` accepts the config object.
```suggestion
class ModuleFactoryV1(Generic[ConfigType, ModuleType]):
```
?

i6_models/config.py
albertz: Btw, to make type hinting happy about this, you might need to adapt the `ModuleType` definition, do define it in a way that says that its `__init__` is in this way.

Atticus1806: @SimBe195

i6_models/config.py
Atticus1806: I dont think so. This is no part/assembly/primitive

albertz: So only parts/assemblies/primitives should be versionized, and everything else not? Why this inconsistency?

Atticus1806: I don't have a strong opinion on this, but I think we don't expect this module to change, since its quite basic.

i6_models/config.py
SimBe195: > Btw, to make type hinting happy about this, you might need to adapt the ModuleType definition, do define it in a way that says that its __init__ is in this way.

And how would this work? Maybe by introducing a new protocol class? @albertz

albertz: I haven't really tried it but I assume so.

Where exactly does it fail anyway? We could also just ignore the type checking for that line where it fails.

I don't think we should define a wrong type here just to make type checking happy. We should define the correct type, and fix the type checking, or disable type checking where it does not work.

Before disabling type checking for that line, I would first try to define the type in the correct way though.

Atticus1806: So I just tested in my PyCharm and `Callable[[ModelConfiguration], ModuleType]` causes 
```
class Test(nn.Module):
    def __init__(self, cfg: TestConfig):
        super().__init__()
        self.x = nn.Linear(cfg.param, cfg.param)
```
to not be recognized as matching type while `Type[ModuleType]`.
When constructing in both cases I did not get a typeguard error.

SimBe195: I currently do not see a way to give typehints for how we expect the `__init__` of a class to look like. But maybe what we could do is at least to put an assertion into `__post_init__` that checks that the signature of `self.module_class.__init__` is correct.

SimBe195: > and Callable[[ModelConfiguration], ModuleType] causes [...] to not be recognized as matching type

It should be `Callable[[ConfigType], ModuleType]`. But anyway I believe `Type[ModuleType]` + a runtime assertion is preferable.

albertz: > So I just tested in my PyCharm and `Callable[[ModelConfiguration], ModuleType]` ...

But my suggestion was to actually not use this and use `Type[ModuleType]` instead? I'm a bit confused now.

> to not be recognized as matching type while `Type[ModuleType]`.

I think there is a typo? I'm not sure I understand the sentence correctly.

---

> It should be `Callable[[ConfigType], ModuleType]`.

No I think it really should be `Type[ModuleType]`. That is what we expect here.

> But anyway I believe `Type[ModuleType]` + a runtime assertion is preferable.

What runtime assertion do you mean? The code I posted was just for demonstration, which would require that it is `Type[ModuleType]`, not `Callable[[ConfigType], ModuleType]`. It was just an example to demonstrate why `Type[ModuleType]` is the correct thing here. 

---



> > Btw, to make type hinting happy about this, you might need to adapt the ModuleType definition, do define it in a way that says that its **init** is in this way.
> 
> And how would this work? Maybe by introducing a new protocol class? @albertz

Actually I tried this code now and I don't see any warning? As I asked before, where exactly do you get a warning when you use `Type[ModuleType]`? I don't see this. I tried with [this code](https://github.com/albertz/playground/blob/master/module_init_typing.py) and I don't see any warnings.

SimBe195: > But my suggestion was to actually not use this and use Type[ModuleType] instead? I'm a bit confused now.

There are basically two approaches:

1. `module_class: Callable[[ConfigType], ModuleType]` which I suggested initially to type-hint the constructor correctly but this has the disadvantage that you can't use things like `isinstance(obj, opt.module_class)` like you said.
2. `module_class: Type[ModuleType]` which is more how we want it but gives no typehints for the constructor, so calling `self.module_class(self.cfg)` may raise a type warning. @Atticus1806  just wanted to test both options in PyCharm.

> No I think it really should be Type[ModuleType]. That is what we expect here.

Yes I agree but only if it's still checked that the constructor of the given `module_class` has the right signature. This way it will already raise an error in sisyphus when you create a `ModelConfiguration` and not only later in the training job.

> What runtime assertion do you mean?

An assertion in `__post_init__` that checks the signature of `module_class.__init__` as I wrote in my [previous message](https://github.com/rwth-i6/i6_models/pull/21#discussion_r1242084866).

> Actually I tried this code now and I don't see any warning? As I asked before, where exactly do you get a warning when you use Type[ModuleType]? I don't see this. I tried with [this code](https://github.com/albertz/playground/blob/master/module_init_typing.py) and I don't see any warnings.

In VSCode (Pylance) do get a warning:

Atticus1806:
@SimBe195  @albertz  is this fine for you? Added a test and the assertion check that the type matches.

i6_models/config.py
```diff
@@ -51,3 +54,25 @@ def _validate_types(self) -> None:
...
+
+    module_class: Type[ModuleType]
+    cfg: ConfigType
+
+    def __call__(self) -> ModuleType:
+        """Constructs an instance of the given module class"""
+        return self.module_class(self.cfg)
+
+    def __post_init__(self) -> None:
+        assert signature(self.module_class).parameters["cfg"].annotation == type(self.cfg)
```
SimBe195, line 78: We shouldn't force the parameter to be called "cfg". Rather check that there is a single parameter and check the type of that one.
```suggestion
        parameters = signature(self.module_class).parameters.values()
        assert len(parameters) == 1
        cfg_parameter = next(iter(parameters))
        assert typeguard.check_type(self.cfg, cfg_parameter.annotation)
```

SimBe195 (approved):

i6_models/config.py
```diff
@@ -51,3 +54,28 @@ def _validate_types(self) -> None:
...
+
+    def __call__(self) -> ModuleType:
+        """Constructs an instance of the given module class"""
+        return self.module_class(self.cfg)
+
+    def __post_init__(self) -> None:
+        parameters = signature(self.module_class).parameters.values()
+        assert len(parameters) == 1
+        cfg_parameter = next(iter(parameters))
+        typeguard.check_type(self.cfg, cfg_parameter.annotation)
```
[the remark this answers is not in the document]
Atticus1806: Can you give an example where this "fails"?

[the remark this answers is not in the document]
albertz: I think in general using `signature` is not totally reliable and it is mostly intended to be used for inspection and debugging.

My examples were just examples. I can imagine there are more cases.

I just don't think such code is really necessary. I would keep it simple. But if you really want to have it, then keep it.

However, I think better than such a check would be to fix the typing, as I said before. Then PyCharm and other tools could also print you warnings.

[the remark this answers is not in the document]
albertz: I don't know. I haven't really tried.

I just asked the [question on StackOverflow](https://stackoverflow.com/questions/76572323/combine-protocol-with-base-class-requirement-type-intersection). If we don't get a good answer there, the [official `typing` package on GitHub](https://github.com/python/typing) might be another good place to ask, or the [official `mypy` package on GitHub](https://github.com/python/mypy).

i6_models/config.py
```diff
@@ -17,6 +17,9 @@ class ExampleModule(Module):
 from __future__ import annotations
 from dataclasses import dataclass, fields
 import typeguard
+from torch import nn
+from typing import Generic, TypeVar, Type
+from inspect import signature
```
SimBe195, line 22:
```suggestion
import inspect
```

i6_models/config.py
SimBe195, line 81: Result of offline discussion: Solving the issue via type-hinting seems to be either not possible or overly complicated. Instead have a more sophisticated type-check here that should cover all cases of how `module_class` can be defined.
```suggestion
        # Check typing of module_class and cfg, i.e. make sure that "self.module_class(self.cfg)" is a valid call.
        parameters = inspect.signature(self.module_class).parameters.values()
        assert len(parameters) >= 1
        parameter_iter = iter(parameters)

        # 1. Check that the first parameter is either not annotated or the annotation matches the type of self.cfg
        cfg_parameter = next(parameter_iter)
        if cfg_parameter.annotation is not inspect.Parameter.empty:
            typeguard.check_type(self.cfg, cfg_parameter.annotation)
            
        # 2. Check that all other parameters have default values
        for parameter in parameter_iter:
            assert parameter.default is not inspect.Parameter.empty
```

SimBe195 (changes requested):

SimBe195 (approved):

i6_models/config.py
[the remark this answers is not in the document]
SimBe195: Yes we decided that we want this check so that any error is detected early on. The check has been made a bit more general now so that all (or at least most) use-cases should be covered. If this breaks for some specific module class later we can adjust the check.

albertz (approved):

PR #22, opened 2023-06-22
Fix MHSA masking

Atticus1806:
This PR fixes the masking issue talked about today (that we usually want 1 to mark positions which are in the sequence, but MHSA expects the other way around)
Also the docstring indicates to me that for float mask weird things happen, so I added a bool conversion. 
```
        key_padding_mask: If specified, a mask of shape :math:`(N, S)` indicating which elements within ``key``
            to ignore for the purpose of attention (i.e. treat as "padding"). For unbatched `query`, shape should be :math:`(S)`.
            Binary and float masks are supported.
            For a binary mask, a ``True`` value indicates that the corresponding ``key`` value will be ignored for
            the purpose of attention. For a float mask, it will be directly added to the corresponding ``key`` value.
```

### DIFF
--- a/i6_models/parts/conformer/mhsa.py
+++ b/i6_models/parts/conformer/mhsa.py
@@ -39,17 +39,17 @@ class ConformerMHSAV1(torch.nn.Module):
         )
         self.dropout = cfg.dropout
 
-    def forward(self, input_tensor: torch.Tensor, key_padding_mask: Optional[torch.Tensor] = None) -> torch.Tensor:
+    def forward(self, input_tensor: torch.Tensor, sequence_mask: Optional[torch.Tensor] = None) -> torch.Tensor:
         """
         Apply layer norm and multi-head self attention and dropout
-        :param Optional[torch.Tensor] key_padding_mask: could be a binary or float mask of shape (B, T)
+        :param Optional[torch.Tensor] sequence_mask: could be a binary or float mask of shape (B, T), 1 signals within sequence, 0 outside, will be inverted to match the torch.nn.MultiheadAttention module
         which will be applied/added to dot product, used to mask padded key positions out
         """
-
+        inv_sequence_mask = 1 - sequence_mask
         output_tensor = self.layernorm(input_tensor)  # [B,T,F]
 
         output_tensor, _ = self.mhsa(
-            output_tensor, output_tensor, output_tensor, key_padding_mask=key_padding_mask, need_weights=False
+            output_tensor, output_tensor, output_tensor, key_padding_mask=inv_sequence_mask, need_weights=False
         )  # [B,T,F]
         output_tensor = torch.nn.functional.dropout(output_tensor, p=self.dropout, training=self.training)  # [B,T,F]
 
--- a/i6_models/parts/conformer/mhsa.py
+++ b/i6_models/parts/conformer/mhsa.py
@@ -45,7 +45,7 @@ class ConformerMHSAV1(torch.nn.Module):
         :param Optional[torch.Tensor] sequence_mask: could be a binary or float mask of shape (B, T), 1 signals within sequence, 0 outside, will be inverted to match the torch.nn.MultiheadAttention module
         which will be applied/added to dot product, used to mask padded key positions out
         """
-        inv_sequence_mask = 1 - sequence_mask
+        inv_sequence_mask = (1 - sequence_mask).bool()
         output_tensor = self.layernorm(input_tensor)  # [B,T,F]
 
         output_tensor, _ = self.mhsa(


JackTemaki (approved):

SimBe195 (approved):

michelwi (approved):

Atticus1806:
Actually now that I think about it, I think it would be better to also convert the mask to float before substraction. Since input might be bool.
@JackTemaki @SimBe195 @michelwi I will push, but I would reask for your approval.

i6_models/parts/conformer/mhsa.py
```diff
@@ -39,17 +39,20 @@ def __init__(self, cfg: ConformerMHSAV1Config):
...
+    def forward(self, input_tensor: torch.Tensor, sequence_mask: Optional[torch.Tensor] = None) -> torch.Tensor:
         """
         Apply layer norm and multi-head self attention and dropout
-        :param Optional[torch.Tensor] key_padding_mask: could be a binary or float mask of shape (B, T)
+        :param Optional[torch.Tensor] sequence_mask: could be a binary or float mask of shape (B, T), 1 signals within sequence, 0 outside, will be inverted to match the torch.nn.MultiheadAttention module
         which will be applied/added to dot product, used to mask padded key positions out
         """
-
+        if sequence_mask is not None:
+            inv_sequence_mask = (1 - sequence_mask.float()).bool()
```
michelwi, line 49: is it concerning that you subtract float mask from integer 1?

how about 
```suggestion
            inv_sequence_mask = torch.logical_not(sequence_mask)
```

Atticus1806: I thought about that too, but there was no clear agreement (yet?) whether the mask is always bool, that I am aware of. So here it might also be float (c.f. #17). But if you prefer logical_not I could always cast the mask to bool beforehand.

michelwi: according to the examples at https://pytorch.org/docs/stable/generated/torch.logical_not.html `torch.logical_not` also handles float and integer inputs (treating 0 and 0.0 as False and all other as True)

Atticus1806: Oh nice, I did not know that.

JackTemaki (approved):

tests/test_conformer.py
```diff
@@ -65,7 +65,11 @@ def get_output_shape(input_shape, cfg, **kwargs):
     # with key padding mask
     input_shape = [4, 15, 32]  # B,T,F
     cfg = ConformerMHSAV1Config(32, 8, 0.2, 0.3)
-    assert get_output_shape(input_shape, cfg, key_padding_mask=torch.randint(0, 2, input_shape[:2]) > 0) == [4, 15, 32]
+    assert get_output_shape(input_shape, cfg, sequence_mask=(torch.randint(0, 2, input_shape[:2]) > 0)) == [
+        4,
+        15,
+        32,
+    ]
```
michelwi, line 72:
```suggestion
    assert get_output_shape(input_shape, cfg, sequence_mask=(torch.randint(0, 2, input_shape[:2]) > 0)) == [4, 15, 32]
```
this line is even shorter than the original one.. black should not be overzealous here

michelwi (approved):
If you want, you may even add test cases for boolean, integer, and float masks

I'll reapprove, but the main part of the change is now my implementation anyway so 🤷

i6_models/parts/conformer/mhsa.py
```diff
@@ -39,17 +39,20 @@ def __init__(self, cfg: ConformerMHSAV1Config):
         )
         self.dropout = cfg.dropout
 
-    def forward(self, input_tensor: torch.Tensor, key_padding_mask: Optional[torch.Tensor] = None) -> torch.Tensor:
+    def forward(self, input_tensor: torch.Tensor, sequence_mask: Optional[torch.Tensor] = None) -> torch.Tensor:
         """
         Apply layer norm and multi-head self attention and dropout
-        :param Optional[torch.Tensor] key_padding_mask: could be a binary or float mask of shape (B, T)
+        :param Optional[torch.Tensor] sequence_mask: could be a binary or float mask of shape (B, T), 1 signals within sequence, 0 outside, will be inverted to match the torch.nn.MultiheadAttention module
```
albertz, line 45: While we change this, we can also fix the inconsistent docstring issue here, namely:
* There should be an empty line between the description and the first param.
* The type info not needed here (it's already typed).
```suggestion

        :param sequence_mask: could be a binary or float mask of shape (B, T), 1 signals within sequence, 0 outside, will be inverted to match the torch.nn.MultiheadAttention module
```

albertz (commented):
In PyTorch, where else is masking used? Is masking usually expected to be this way (inverted from what we thought)? Should we maybe change the meaning of our `sequence_mask` for all modules? Then we don't need the inversion here.

SimBe195 (approved):

Atticus1806:
Re-requesting review due to change. Now using xor, because this is exportable with the current setup.

JackTemaki (approved):

i6_models/parts/conformer/mhsa.py
```diff
@@ -39,17 +39,21 @@ def __init__(self, cfg: ConformerMHSAV1Config):
         )
         self.dropout = cfg.dropout
 
-    def forward(self, input_tensor: torch.Tensor, key_padding_mask: Optional[torch.Tensor] = None) -> torch.Tensor:
+    def forward(self, input_tensor: torch.Tensor, sequence_mask: Optional[torch.Tensor] = None) -> torch.Tensor:
```
albertz, line 42: Do we really want to allow to have it optional?

i6_models/parts/conformer/mhsa.py
```diff
@@ -39,17 +39,21 @@ def __init__(self, cfg: ConformerMHSAV1Config):
...
         """
         Apply layer norm and multi-head self attention and dropout
-        :param Optional[torch.Tensor] key_padding_mask: could be a binary or float mask of shape (B, T)
+
+        :param sequence_mask: could be a binary or float mask of shape (B, T), 1 signals within sequence, 0 outside, will be inverted to match the torch.nn.MultiheadAttention module
```
albertz, line 46: First, I would say bool mask, not binary mask.

Then: Do we really want to allow both cases here (bool or float)? We could also just say that we expect a bool mask in any case.

Then, what was the result of the discussion now w.r.t. what we actually want to have? Bool or float? Seq mask or seq len or some dataclass? Inv seq mask or seq mask? On device or CPU? I don't see that in the meeting notes. Or when I left the meeting, I think it was bool mask on GPU, but there was still ongoing discussion, maybe the dataclass was still a valid option?

i6_models/parts/conformer/mhsa.py
Atticus1806: We need this for export. During export, from our current knowledge it is almost impossible to generate a sequence mask (because all variants @JackTemaki and I found require `max_len` as int and/or torch.arange (which needs int input)). This is not exportable. One could write a custom op, but we decided against it for now because we don't do batch decoding anyway, this could be later added I guess if required. 
And then simply putting the mask to None in the initial Module is the solution we came up with.

i6_models/parts/conformer/mhsa.py
Atticus1806: Good catch on the doc, I missed that. Regarding the discussion I cannot say much so maybe @christophmluscher @SimBe195 or someone else could comment.

i6_models/parts/conformer/mhsa.py
albertz: What code did you try? I just tried `torch.arange(x.shape[0])` for some tensor `x` and that seems to work fine?

albertz: My testing code: https://github.com/albertz/playground/blob/master/torch-tracing-arange.py

Atticus1806: Hmm, you are right this seems to work. I have to admit most of the testing was done by @JackTemaki, but like this the export of the conformer works for me. 
I would wait for @JackTemaki s feedback here.

albertz: I extended the demo code for doing a seq mask, and that also works just fine: https://github.com/albertz/playground/blob/master/torch-tracing-seqmask.py

Atticus1806: So it seems @albertz variant is working. This means I will remove the support for optional mask here.

i6_models/parts/conformer/mhsa.py
```diff
@@ -39,17 +39,21 @@ def __init__(self, cfg: ConformerMHSAV1Config):
...
         which will be applied/added to dot product, used to mask padded key positions out
         """
-
+        if sequence_mask is not None:
+            inv_sequence_mask = torch.logical_xor(sequence_mask, torch.ones(sequence_mask.shape))
```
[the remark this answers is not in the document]
Atticus1806: But how would we track this? always assume something like `self.export_model` is set?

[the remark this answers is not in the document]
albertz: `torch.onnx.is_in_onnx_export()`

i6_models/parts/conformer/mhsa.py
```diff
@@ -39,18 +39,28 @@ def __init__(self, cfg: ConformerMHSAV1Config):
         )
         self.dropout = cfg.dropout
 
-    def forward(self, input_tensor: torch.Tensor, sequence_mask: Optional[torch.Tensor] = None) -> torch.Tensor:
+    @staticmethod
+    def _invert_sequence_mask(sequence_mask: torch.Tensor):
```
albertz, line 43: Maybe we want to move this function somewhere else, and call it maybe sth more expressive like `compat_logical_not`?

Atticus1806: we could do `util/compat.py` and then put `logical_not` in there?

i6_models/parts/conformer/mhsa.py
```diff
@@ -39,18 +39,28 @@ def __init__(self, cfg: ConformerMHSAV1Config):
...
+        """
+        Helper function to decide how to invert the sequence mask. For ONNX export use XOR with 1 since logical_not is not implemented.
+        Else logical_not is applied for efficiency reasons.
+
+        :param sequence_mask: bool mask of shape (B, T) to be inverted.
+        """
+        if torch.onnx.is_in_onnx_export():
+            return torch.logical_xor(sequence_mask, torch.ones(sequence_mask.shape, device=sequence_mask.device))
```
albertz, line 51:
```suggestion
            return torch.logical_xor(sequence_mask, torch.ones_like(sequence_mask))
```

Atticus1806: I think device is required. At least for me it crashed in my models when this was not set.

i6_models/util/compat.py
```diff
@@ -0,0 +1,14 @@
+import torch
+
+
+def _invert_sequence_mask(sequence_mask: torch.Tensor):
```
albertz, line 4:
```suggestion
def logical_not(tensor: torch.Tensor):
```

i6_models/parts/conformer/mhsa.py
```diff
@@ -2,10 +2,10 @@
 
 __all__ = ["ConformerMHSAV1", "ConformerMHSAV1Config"]
 from dataclasses import dataclass
-from typing import Optional
 import torch
 
 from i6_models.config import ModelConfiguration
+from i6_models.util.compat import logical_not
```
albertz, line 8: I would prefer to just import the module, like:
```suggestion
from i6_models.util import compat
```
And then below use `compat.logical_not`.

i6_models/util/compat.py
```diff
@@ -0,0 +1,18 @@
+"""
+Compatibility support for different functions. This could be for example for onnx export.
+"""
+
+import torch
+
+
+def logical_not(tensor: torch.Tensor, /):
```
albertz, line 8:
```suggestion
def logical_not(tensor: torch.Tensor, /) -> torch.Tensor:
```

albertz (approved):

christophmluscher (approved):

PR #26, opened 2023-07-14
add LogMelFeatureExtraction

JackTemaki:
Adds PyTorch-based log-mel feature extraction that is compatible to the librosa-based feature extraction in RETURNN.

The log-scaling and normalization are missing on purpose, as this should be an independent user decision.

Is a draft for now as test cases are still missing, but feel free to comment!

### DIFF
--- a/i6_models/primitives/feature_extraction.py
+++ b/i6_models/primitives/feature_extraction.py
@@ -1,0 +1,96 @@
+__all__ = ["LogMelFeatureExtractionV1", "LogMelFeatureExtractionV1Config"]
+
+from dataclasses import dataclass
+from typing import Tuple
+
+from librosa import filters
+import torch
+from torch import nn
+
+from i6_models.config import ModelConfiguration
+
+
+@dataclass
+class LogMelFeatureExtractionV1Config(ModelConfiguration):
+    """
+    Attributes:
+        sample_rate: audio sample rate in Hz
+        win_size: window size in seconds
+        hop_size: window shift in seconds
+        f_min: minimum mel filter frequency in Hz
+        f_max: maximum mel fitler frequency in Hz
+        min_amp: minimum amplitude for safe log
+        num_filters: number of mel windows
+        center: centered STFT with automatic padding
+    """
+    sample_rate: int
+    win_size: float
+    hop_size: float
+    f_min: int
+    f_max: int
+    min_amp: float
+    num_filters: int
+    center: bool
+
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        assert self.f_max <= self.sample_rate // 2, "f_max can not be larger than half of the sample rate"
+        assert self.f_min > 0 and self.f_max > 0 and self.sample_rate > 0, "frequencies need to be positive"
+        assert self.win_size > 0 and self.hop_size > 0, "window settings need to be positive"
+        assert self.num_filters > 0, "number of filters needs to be positive"
+        assert self.hop_size <= self.win_size, "using a larger hop size than window size does not make sense"
+
+
+class LogMelFeatureExtractionV1(nn.Module):
+    """
+        Librosa-compatible log-mel feature extraction using log10. Does not use torchaudio.
+
+        Using it wrapped with torch.no_grad() is recommended if no gradient is needed
+    """
+
+    def __init__(
+            self,
+            cfg: LogMelFeatureExtractionV1Config
+    ):
+        super().__init__()
+        self.register_buffer("n_fft", torch.tensor(int(cfg.win_size * cfg.sample_rate)))
+        self.register_buffer("hop_length", torch.tensor(int(cfg.hop_size * cfg.sample_rate)))
+        self.register_buffer("min_amp", torch.tensor(cfg.min_amp))
+        self.center = cfg.center
+        self.register_buffer("mel_basis", torch.tensor(filters.mel(
+            sr=cfg.sample_rate,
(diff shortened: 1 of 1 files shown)


i6_models/primitives/feature_extraction.py
```diff
@@ -0,0 +1,105 @@
...
+
+@dataclass
+class LogMelFeatureExtractionV1Config(ModelConfiguration):
+    """
+    Attributes:
+        sample_rate: audio sample rate in Hz
+        win_size: window size in seconds
+        hop_size: window shift in seconds
+        f_min: minimum mel filter frequency in Hz
+        f_max: maximum mel fitler frequency in Hz
```
vieting, line 21:
```suggestion
        f_max: maximum mel filter frequency in Hz
```

vieting: Also, I think I'd remove the "mel" from this line and the one above.

i6_models/primitives/feature_extraction.py
```diff
@@ -0,0 +1,105 @@
...
+        :param raw_audio: [B, T]
+        :param length in samples: [B]
+        :return features as [B,T,F] and length in frames [B]
+        """
+
+        S = (
+            torch.abs(
+                torch.stft(
+                    raw_audio,
+                    n_fft=self.n_fft,
```
vieting, line 83: This does not allow a window size different from the FFT size. I think we do use that in other setups.

i6_models/primitives/feature_extraction.py
```diff
@@ -0,0 +1,105 @@
...
+        self.register_buffer("window", torch.hann_window(int(cfg.win_size * cfg.sample_rate)))
+
+    def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        :param raw_audio: [B, T]
+        :param length in samples: [B]
+        :return features as [B,T,F] and length in frames [B]
+        """
+
+        S = (
```
vieting, line 79: Is there a reason for choosing this variable name?

vieting (commented):
Looks mostly good to me, just some small comments.

i6_models/primitives/feature_extraction.py
```diff
@@ -0,0 +1,105 @@
...
+
+@dataclass
+class LogMelFeatureExtractionV1Config(ModelConfiguration):
+    """
+    Attributes:
+        sample_rate: audio sample rate in Hz
+        win_size: window size in seconds
+        hop_size: window shift in seconds
+        f_min: minimum mel filter frequency in Hz
+        f_max: maximum mel filter frequency in Hz
```
JackTemaki, line 21:
```suggestion
        f_min: minimum filter frequency in Hz
        f_max: maximum filter frequency in Hz
```

i6_models/primitives/feature_extraction.py
JackTemaki: For RETURNN feature extraction, this was never possible! But sure, I can add that.

i6_models/primitives/feature_extraction.py
JackTemaki: will change

i6_models/primitives/feature_extraction.py
vieting: For log Mel features in the network, this was possible because you can specify it directly for the STFT.

JackTemaki: Yes sure, I was talking about the "old" librosa feature extraction that I am copying here.

tests/test_feature_extraction.py
```diff
@@ -0,0 +1,67 @@
...
+        min_amp=1e-10,
+        num_filters=80,
+        center=True,
+    )
+    fe_center = LogMelFeatureExtractionV1(cfg=fe_center_cfg)
+    fe_no_center_cfg = copy.deepcopy(fe_center_cfg)
+    fe_no_center_cfg.center = False
+    fe_no_center = LogMelFeatureExtractionV1(cfg=fe_no_center_cfg)
+    for i in range(10):
+        audio_length = int(numpy.random.randint(10000, 50000))
```
vieting, line 59: Should the random seed be fixed in order to make the test deterministic?

vieting (changes requested):
Some more small comments.

tests/test_feature_extraction.py
JackTemaki: There is no problem in having it non-deterministic, it could even catch an edge case this way (which should not exist though)

i6_models/primitives/feature_extraction.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        assert self.f_max <= self.sample_rate // 2, "f_max can not be larger than half of the sample rate"
+        assert self.f_min > 0 and self.f_max > 0 and self.sample_rate > 0, "frequencies need to be positive"
+        assert self.win_size > 0 and self.hop_size > 0, "window settings need to be positive"
+        assert self.num_filters > 0, "number of filters needs to be positive"
+        assert self.hop_size <= self.win_size, "using a larger hop size than window size does not make sense"
+        assert (
+            self.n_fft is None or self.n_fft >= self.win_size * self.sample_rate
```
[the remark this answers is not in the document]
JackTemaki: The int conversion is not needed here

vieting (approved):

JackTemaki:
@mmz33 as this is relevant for you and Atanas, can you review?

PR #28, opened 2023-07-26
Draft: Implements Encoder-Decoder Attention Model

mmz33:


JackTemaki:
I would recommend to build full setups before merging this to avoid the same problems we had with other code that we merged but never "used" before, I would help with this.

mmz33:
> I would recommend to build full setups before merging this to avoid the same problems we had with other code that we merged but never "used" before, I would help with this.

Yeah I agree.

JackTemaki (approved):

JackTemaki:
I completely forgot to test with this branch again after using it already for some time, did so now and it works normally.

i6_models/decoder/attention.py
```diff
@@ -0,0 +1,209 @@
...
+        self.output = nn.Linear(cfg.output_proj_dim // 2, cfg.vocab_size)
+        self.output_dropout = nn.Dropout(cfg.output_dropout)
+
+    def forward(
+        self,
+        encoder_outputs: torch.Tensor,
+        labels: torch.Tensor,
+        enc_seq_len: torch.Tensor,
+        state: Optional[Tuple[torch.Tensor, ...]] = None,
+        shift_embeddings=True,
```
Atticus1806, line 135:
```suggestion
        shift_embeddings: bool = True,
```

i6_models/decoder/zoneout_lstm.py
```diff
@@ -0,0 +1,47 @@
+import torch
+from torch import nn
+
+from typing import Tuple
+
+
+class ZoneoutLSTMCell(nn.Module):
```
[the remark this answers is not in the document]
JackTemaki: I do not like non-standard types in a config object, and nn.RNNCellBase would be one, thus no, I at least do not want one.

Atticus1806 (approved):

SimBe195 (changes requested):

i6_models/decoder/zoneout_lstm.py
```diff
@@ -0,0 +1,48 @@
...
+        super().__init__()
+        self.cell = cell
+        assert 0.0 <= zoneout_h <= 1.0 and 0.0 <= zoneout_c <= 1.0, "Zoneout drop probability must be in [0, 1]"
+        self.zoneout_h = zoneout_h
+        self.zoneout_c = zoneout_c
+
+    def forward(
+        self, inputs: torch.Tensor, state: Tuple[torch.Tensor, torch.Tensor]
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        with torch.autocast(device_type="cuda", enabled=False):
```
albertz, line 27: Why is this disabled here? That should be explained in the code, maybe with some ref.

SimBe195 (approved):

PR #29, opened 2023-07-27
Conformer ONNX export test case

JackTemaki:
This is a full test case to test that all our Conformer code is actually exportable.

Also adds a `tensor_length_to_mask` function, which is inspired by https://github.com/albertz/playground/blob/master/torch-tracing-seqmask.py

### DIFF
--- a/.github/workflows/model_tests.yml
+++ b/.github/workflows/model_tests.yml
@@ -20,8 +20,7 @@ jobs:
           cache: 'pip'
       - run: |
           pip install pytest
-          pip install -r requirements.txt
-          pip install -r requirements_dev.txt
+          pip install typeguard librosa torch==1.13.1 onnx=1.13.1 onnxruntime=1.13.1
       - name: Test Models
         run: |
           python -m pytest tests
--- a/.github/workflows/model_tests.yml
+++ b/.github/workflows/model_tests.yml
@@ -20,7 +20,7 @@ jobs:
           cache: 'pip'
       - run: |
           pip install pytest
-          pip install typeguard librosa torch==1.13.1 onnx=1.13.1 onnxruntime=1.13.1
+          pip install typeguard librosa torch==1.13.1 onnx==1.13.1 onnxruntime==1.13.1
       - name: Test Models
         run: |
           python -m pytest tests
--- a/i6_models/util/compat.py
+++ b/i6_models/util/compat.py
@@ -12,7 +12,7 @@ def logical_not(tensor: torch.Tensor, /) -> torch.Tensor:
 
     :param tensor: bool mask of shape (B, T) to be inverted.
     """
-    if torch.onnx.is_in_onnx_export():
+    if torch.jit.is_scripting() or torch.jit.is_tracing():
         return torch.logical_xor(tensor, torch.ones_like(tensor))
     else:
         return torch.logical_not(tensor)
--- a/i6_models/util/compat.py
+++ b/i6_models/util/compat.py
@@ -12,7 +12,7 @@ def logical_not(tensor: torch.Tensor, /) -> torch.Tensor:
 
     :param tensor: bool mask of shape (B, T) to be inverted.
     """
-    if torch.jit.is_scripting() or torch.jit.is_tracing():
+    if torch.onnx.is_in_onnx_export():
         return torch.logical_xor(tensor, torch.ones_like(tensor))
     else:
         return torch.logical_not(tensor)
--- a/i6_models/util/mask.py
+++ b/i6_models/util/mask.py
@@ -1,0 +1,16 @@
+import torch
+
+
+def tensor_mask_from_length(tensor: torch.Tensor, seq_len: torch.Tensor) -> torch.Tensor:
+    """
+    mask a tensor with a "positive" mask (boolean true means position is used)
+
+    This function is traceable.
+
+    :param tensor: [B,T,....]
+    :param seq_len: [B]
+    :return: [B,T]
+    """
+    r = torch.arange(tensor.shape[1], device=tensor.device)  # [T]
+    seq_mask = torch.less(r[None, :], seq_len[:, None])  # broadcast to [B,T]
+    return seq_mask
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 typeguard
-torch==1.13.1
+torch
 librosa
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 typeguard
-torch
+torch==1.13.1
 librosa
--- a/tests/test_conformer.py
+++ b/tests/test_conformer.py
@@ -155,7 +155,7 @@ def test_conformer_onnx_export():
         assert torch.allclose(outputs_normal, outputs_traced, atol=1e-5)
 
         export_onnx(
-            traced_model,
+            model,
             (dummy_data, dummy_data_len),
             f=f,
             verbose=True,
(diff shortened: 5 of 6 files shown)


i6_models/util/compat.py
```diff
@@ -12,7 +12,7 @@ def logical_not(tensor: torch.Tensor, /) -> torch.Tensor:
 
     :param tensor: bool mask of shape (B, T) to be inverted.
     """
-    if torch.onnx.is_in_onnx_export():
+    if torch.jit.is_scripting() or torch.jit.is_tracing():
```
Atticus1806, line 15: does this include all scenarios where `is_in_onnx_export` is `True`?

tests/test_conformer.py
```diff
@@ -79,3 +87,112 @@ def get_output(shape, norm):
...
+        outputs_onnx_other = torch.FloatTensor(
+            session.run(None, {"data": dummy_data.numpy(), "data_len": dummy_data_len_2.numpy()})[0]
+        )
+
+        # The default 1e-8 was slightly too strong
+        assert torch.allclose(outputs_normal, outputs_onnx, atol=1e-5)
+        # check that for different lengths we really get a different result
+        assert not torch.allclose(outputs_normal, outputs_onnx_other, atol=1e-5)
+
+        # in the future check with different batching and max size (20)
```
Atticus1806, line 184: is this TODO in this PR or in general?

i6_models/util/compat.py
JackTemaki: I thought it is, but going through the code I am actually not so sure anymore. It seems the internal tracing via the ONNX call is not exactly the same as the regular `jit.trace()`. I think just to be safe we might want to have all 3 checks....

tests/test_conformer.py
JackTemaki: In this PR we can not fix this, so this is a general warning that our model is not batch-size-safe, and to give an idea how to expand the tests once it is.

i6_models/util/compat.py
albertz: Why is this change needed for ONNX export?

The code below (`logical_not`) was specifically about missing (or buggy/broken) Torch-to-ONNX support, not about scripting or tracing.

JackTemaki: You might want to use e.g. scripting instead of tracing, but then this is called before the actual ONNX export, so `is_in_onnx_export` does not work.

requirements.txt
```diff
@@ -1,3 +1,3 @@
 typeguard
-torch
+torch==1.13.1
```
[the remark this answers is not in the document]
JackTemaki: > Specifically, we should always recommend to use the latest PyTorch version

As long as people do not actually expect this to work with the latest version to work, fine.

i6_models/util/compat.py
albertz: You might want to use scripting or tracing and not use ONNX, and then you would not want to have this, so this condition would be wrong then.

We should just make it more explicit then. We can have an own global flag `in_onnx_export`, which we can set in our torch-to-onnx-export script.

JackTemaki: I will just revert this change for now to not delay this further, scripting is anyway something only I used so far. Then it works as expected for the RETURNN script, and I will modify my own setup.

Atticus1806 (approved):

i6_models/util/mask.py
```diff
@@ -0,0 +1,16 @@
+import torch
+
+
+def tensor_mask_from_length(tensor: torch.Tensor, seq_len: torch.Tensor) -> torch.Tensor:
```
albertz, line 4: The function does not really depend on `tensor` at all, except for `tensor.shape[1]`. I would only use that as an explicit arg `max_len`.

The corresponding function in TensorFlow is [`tf.sequence_mask`](https://www.tensorflow.org/api_docs/python/tf/sequence_mask), which has a much better name and API. Why don't we just follow that (more or less)? Specifically, I would suggest:

```suggestion
def sequence_mask(lens: torch.Tensor, /, *, max_len: Optional[torch.Tensor] = None) -> torch.Tensor:
```

The `max_len` can be optional. We can just use `torch.max(lens)` if it is not explicitly given.

JackTemaki: I had `torch.max(lens)` in the original function, and this was not traceable, so I proposed to disable masks for the export. Then you made the suggestion to use tensor.shape, because that is traceable. Why do you want to go back now?

albertz: I did not suggest to change any logic here. I just say that `tensor` should not be the argument of the function, but `maxlen` (thus `tensor.shape[1]`) instead.

Also, I suggested to extend the logic by allowing `max_len=None` and having this default fallback logic.

And `torch.max` is traceable. It works just fine. Maybe you used `max` instead? This probably casts the `torch.Tensor` to a Python `int` and thus makes it non-traceable.

JackTemaki: Sorry for not being specific enough, I said using `torch.max` makes the whole function non-traceable, not that `torch.max` itself is not traceable. `torch.arange` only accepts an integer., so you would have to cast the result of torch.max to int, making it non-traceable. For some reason when using `tensor.shape` tracing works, although it is usually  treated like an integer, not a tensor.

JackTemaki: Hmm right, I wonder why the old tests failed. Yes, then the API change makes sense.

JackTemaki: What should I do for the device? Should we run the mask creation then on CPU and copy later, or pass device as extra argument?

albertz: Ah good question. `device` should be an extra argument. But we probably want to allow for some default behavior. Using the device of `seq_len` is maybe not a good idea, as in practice, `seq_len` often will be on CPU, but we want the seq mask usually on GPU.

In Torch >=2, it might make sense to rely on the Torch default device behavior (using `with torch.device(...)` context scope or `torch.set_default_device(...)`), which you get automatically when you pass `device=None` (or not at all) to `torch.arange`. So maybe we can just leave it as that, and users of outdated PyTorch versions must explicitly pass `device`.

.github/workflows/model_tests.yml
```diff
@@ -20,8 +20,7 @@ jobs:
           cache: 'pip'
       - run: |
           pip install pytest
-          pip install -r requirements.txt
-          pip install -r requirements_dev.txt
+          pip install typeguard librosa torch==1.13.1 onnx==1.13.1 onnxruntime==1.13.1
```
christophmluscher, line 23: I am just wondering why is this better than using the `requirements[_dev].txt`?

christophmluscher (approved):

albertz (changes requested):

tests/test_conformer.py
```diff
@@ -79,3 +87,112 @@ def get_output(shape, norm):
...
+            output_names=["classes"],
+            dynamic_axes={
+                "data": {0: "batch", 1: "time"},
+                "data_len": {0: "batch"},
+                "classes": {0: "batch", 1: "time"},
+            },
+        )
+
+        session = ort.InferenceSession(f.name)
+        outputs_onnx = torch.FloatTensor(
```
kuacakuaca, line 172: can you check if forwarding with new data which have a different time dimension works? I have been testing recognition with the exported onnx model which failed at MHSA reshape op. The requested size had a fixed time dim.

JackTemaki:
Can someone continue this PR? I will probably not find the time to look at this until the end of my internship...

JackTemaki:
This will be covered by Kaloyan in later PRs.

PR #34, opened 2023-09-06
Legacy-like zero-Specaugment implementation

JackTemaki:
Thanks to @mmz33 , here some specaugment compatible to old Returnn/TF setups. Code follows mostly the Tensorflow implementation.

### DIFF
--- a/i6_models/primitives/specaugment.py
+++ b/i6_models/primitives/specaugment.py
@@ -11,9 +11,9 @@ def _mask(tensor: torch.Tensor, batch_axis: int, axis: int, pos: torch.Tensor, m
     """
     batch_dim = tensor.shape[batch_axis]
     dim = tensor.shape[axis]
-    amount = torch.randint(low=1, high=max_len + 1, size=(batch_dim,), dtype=torch.int32).to(device=tensor.device)
-    pos2 = torch.min(pos + amount, torch.tensor([dim] * batch_dim).to(device=tensor.device))
-    idxs = torch.arange(0, dim).to(device=tensor.device).unsqueeze(0)  # [1,dim]
+    amount = torch.randint(low=1, high=max_len + 1, size=(batch_dim,), dtype=torch.int32, device=tensor.device)
+    pos2 = torch.min(pos + amount, torch.tensor([dim] * batch_dim, device=tensor.device))
+    idxs = torch.arange(0, dim, device=tensor.device).unsqueeze(0)  # [1,dim]
     pos_bc = pos.unsqueeze(1)  # [B,1]
     pos2_bc = pos2.unsqueeze(1)  # [B,1]
     cond = torch.logical_and(torch.greater_equal(idxs, pos_bc), torch.less(idxs, pos2_bc))  # [B,dim]
@@ -39,7 +39,7 @@ def _random_mask(tensor: torch.Tensor, batch_axis: int, axis: int, min_num: int,
     """
 
     batch_dim = tensor.shape[batch_axis]
-    num_masks = torch.randint(min_num, max_num, size=(batch_dim,)).to(device=tensor.device)  # [B]
+    num_masks = torch.randint(min_num, max_num, size=(batch_dim,), device=tensor.device)  # [B]
 
     z = -torch.log(-torch.log(torch.rand((batch_dim, tensor.shape[axis])).to(device=tensor.device)))  # [B,dim]
     _, indices = torch.topk(z, num_masks.max().item(), dim=1)
@@ -51,6 +51,7 @@ def _random_mask(tensor: torch.Tensor, batch_axis: int, axis: int, min_num: int,
 
 def zero_specaugment(
     audio_features: torch.Tensor,
+    *,
     time_min_num_masks: int,
     time_max_num_masks: int,
     time_mask_max_size: int,
@@ -85,6 +86,7 @@ def zero_specaugment(
 
 def zero_specaugment_by_length(
     audio_features: torch.Tensor,
+    *,
     time_mask_per_n_frames: int,
     time_min_num_masks: int,
     time_mask_max_size: int,
--- a/i6_models/primitives/specaugment.py
+++ b/i6_models/primitives/specaugment.py
@@ -39,19 +39,15 @@ def _random_mask(tensor: torch.Tensor, batch_axis: int, axis: int, min_num: int,
     """
 
     batch_dim = tensor.shape[batch_axis]
-    num_masks = torch.randint(min_num, max_num, size=(batch_dim,))  # [B]
+    num_masks = torch.randint(min_num, max_num, size=(batch_dim,), device="cpu")  # [B]
 
     max_num_masks = num_masks.max().item()
 
     z = torch.rand((batch_dim, tensor.shape[axis]), device=tensor.device)  # [B,dim]
-    _, indices = torch.topk(z, num_masks.max().item(), dim=1)
+    _, indices = torch.topk(z, max_num_masks, dim=1)
 
     # Make num_masks broadcastable to shape of tensor for torch.where.
-    for i in range(tensor.dim() - 1):
-        if i < batch_axis:
-            num_masks = num_masks.unsqueeze(0)
-        else:
-            num_masks = num_masks.unsqueeze(-1)
+    num_masks = torch.reshape(num_masks, [1] * batch_axis + [batch_dim] + [1] * (tensor.dim() - batch_axis - 1))
 
     num_masks = num_masks.to(device=tensor.device)
 
@@ -115,7 +111,7 @@ def specaugment_v1_by_length(
     freq_mask_max_size: int,
 ):
     """
-    Convenience wrapper around zero_specaugment with time-length adaptive number of masks.
+    Convenience wrapper around specaugment_v1 with time-length adaptive number of masks.
 
     :param audio_features: e.g. log-mel features as [B, T, F]
     :param time_max_mask_per_n_frames: used for the maximum number time masks,
--- a/i6_models/primitives/specaugment.py
+++ b/i6_models/primitives/specaugment.py
@@ -9,18 +9,18 @@ def _mask(tensor: torch.Tensor, batch_axis: int, axis: int, pos: torch.Tensor, m
     :param pos: at which positions along axis to start the mask (size [B])
     :param max_len: mask length drawn uniformly from [0, max_len]
     """
-    batch_dim = tensor.shape[batch_axis]
-    dim = tensor.shape[axis]
-    amount = torch.randint(low=1, high=max_len + 1, size=(batch_dim,), dtype=torch.int32, device=tensor.device)
-    pos2 = torch.min(pos + amount, torch.tensor([dim] * batch_dim, device=tensor.device))
-    idxs = torch.arange(0, dim, device=tensor.device).unsqueeze(0)  # [1,dim]
+    batch_dim_size = tensor.shape[batch_axis]
+    mask_dim_size = tensor.shape[axis]
+    mask_len = torch.randint(low=1, high=max_len + 1, size=(batch_dim_size,), dtype=torch.int32, device=tensor.device)
+    end_pos = torch.min(pos + mask_len, torch.tensor([mask_dim_size] * batch_dim_size, device=tensor.device))
+    idxs = torch.arange(0, mask_dim_size, device=tensor.device).unsqueeze(0)  # [1,dim]
     pos_bc = pos.unsqueeze(1)  # [B,1]
-    pos2_bc = pos2.unsqueeze(1)  # [B,1]
-    cond = torch.logical_and(torch.greater_equal(idxs, pos_bc), torch.less(idxs, pos2_bc))  # [B,dim]
+    end_pos_bc = end_pos.unsqueeze(1)  # [B,1]
+    mask = torch.logical_and(torch.greater_equal(idxs, pos_bc), torch.less(idxs, end_pos_bc))  # [B,dim]
     if batch_axis > axis:
-        cond = cond.transpose(0, 1)  # [dim,B]
-    cond = torch.reshape(cond, shape=[tensor.shape[i] if i in (batch_axis, axis) else 1 for i in range(tensor.ndim)])
-    tensor = torch.where(cond, 0.0, tensor)
+        mask = mask.transpose(0, 1)  # [dim,B]
+    mask = torch.reshape(mask, shape=[tensor.shape[i] if i in (batch_axis, axis) else 1 for i in range(tensor.ndim)])
+    tensor = torch.where(mask, 0.0, tensor)
     return tensor
 
 
@@ -36,18 +36,18 @@ def _random_mask(tensor: torch.Tensor, batch_axis: int, axis: int, min_num: int,
     :param max_amount: mask length drawn uniformly from [0, max_amount]
     """
 
-    batch_dim = tensor.shape[batch_axis]
+    batch_dim_size = tensor.shape[batch_axis]
     if max_num < min_num:
         max_num = min_num
-    num_masks = torch.randint(min_num, max_num + 1, size=(batch_dim,), device="cpu")  # [B]
+    num_masks = torch.randint(min_num, max_num + 1, size=(batch_dim_size,), device="cpu")  # [B]
 
     max_num_masks = num_masks.max().item()
 
-    z = torch.rand((batch_dim, tensor.shape[axis]), device=tensor.device)  # [B,dim]
+    z = torch.rand((batch_dim_size, tensor.shape[axis]), device=tensor.device)  # [B,dim]
     _, indices = torch.topk(z, max_num_masks, dim=1)
 
     # Make num_masks broadcastable to shape of tensor for torch.where.
-    num_masks = torch.reshape(num_masks, [1] * batch_axis + [batch_dim] + [1] * (tensor.dim() - batch_axis - 1))
+    num_masks = torch.reshape(num_masks, [1] * batch_axis + [batch_dim_size] + [1] * (tensor.dim() - batch_axis - 1))
 
     num_masks = num_masks.to(device=tensor.device)
 
--- a/i6_models/primitives/specaugment.py
+++ b/i6_models/primitives/specaugment.py
@@ -41,7 +41,7 @@ def _random_mask(tensor: torch.Tensor, batch_axis: int, axis: int, min_num: int,
     batch_dim = tensor.shape[batch_axis]
     num_masks = torch.randint(min_num, max_num, size=(batch_dim,), device=tensor.device)  # [B]
 
-    z = -torch.log(-torch.log(torch.rand((batch_dim, tensor.shape[axis])).to(device=tensor.device)))  # [B,dim]
+    z = torch.rand((batch_dim, tensor.shape[axis]), device=tensor.device)  # [B,dim]
     _, indices = torch.topk(z, num_masks.max().item(), dim=1)
 
     for i in range(num_masks.max().item()):
@@ -49,7 +49,7 @@ def _random_mask(tensor: torch.Tensor, batch_axis: int, axis: int, min_num: int,
     return tensor
 
 
-def zero_specaugment(
+def specaugment_v1(
     audio_features: torch.Tensor,
     *,
     time_min_num_masks: int,
@@ -74,21 +74,23 @@ def zero_specaugment(
     :param freq_mask_max_size: maximum size of masks along F
     :return: masked audio features
     """
-    assert len(tensor.shape) == 3
-    tensor = _random_mask(
+    assert len(audio_features.shape) == 3
+    assert time_min_num_masks <= time_max_num_masks
+    assert freq_min_num_masks <= freq_max_num_masks
+    masked_audio_features = _random_mask(
         audio_features, 0, 1, time_min_num_masks, time_max_num_masks, time_mask_max_size
     )  # time masking
-    tensor = _random_mask(
-        audio_features, 0, 2, freq_min_num_masks, freq_max_num_masks, freq_mask_max_size
+    masked_audio_features = _random_mask(
+        masked_audio_features, 0, 2, freq_min_num_masks, freq_max_num_masks, freq_mask_max_size
     )  # freq masking
-    return tensor
+    return masked_audio_features
 
 
-def zero_specaugment_by_length(
+def specaugment_v1_by_length(
     audio_features: torch.Tensor,
     *,
-    time_mask_per_n_frames: int,
     time_min_num_masks: int,
+    time_max_mask_per_n_frames: int,
     time_mask_max_size: int,
     freq_min_num_masks: int,
     freq_max_num_masks: int,
@@ -98,7 +100,8 @@ def zero_specaugment_by_length(
     Convenience wrapper around zero_specaugment with time-length adaptive number of masks
 
     :param audio_features: e.g. log-mel features as [B, T, F]
-    :param time_mask_per_n_frames: maximum number of masks depending on length T.
+    :param time_max_mask_per_n_frames: used for the maximum number time masks,
+        max_num_masks = T / max_mask_per_n_frames for each batch.
         They are still drawn depending on the full batch length, so shorter sequences
         might get more masks than that by chance, or none at all when all masks
         fall into the padding space.
@@ -109,10 +112,10 @@ def zero_specaugment_by_length(
     :param freq_mask_max_size: maximum size of masks along F
     :return: masked audio features
     """
-    return zero_specaugment(
+    return specaugment_v1(
         audio_features,
         time_min_num_masks=time_min_num_masks,
-        time_max_num_masks=audio_features.size(1) // time_mask_per_n_frames,
+        time_max_num_masks=audio_features.size(1) // time_max_mask_per_n_frames,
         time_mask_max_size=time_mask_max_size,
         freq_min_num_masks=freq_min_num_masks,
         freq_max_num_masks=freq_max_num_masks,
--- a/i6_models/primitives/specaugment.py
+++ b/i6_models/primitives/specaugment.py
@@ -20,7 +20,7 @@ def _mask(tensor: torch.Tensor, batch_axis: int, axis: int, pos: torch.Tensor, m
     if batch_axis > axis:
         cond = cond.transpose(0, 1)  # [dim,B]
     cond = torch.reshape(
-        cond, shape=[tensor.shape[i] if i in (batch_axis, axis) else 1 for i in range(len(tensor.shape))]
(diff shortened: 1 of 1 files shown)


i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,114 @@
...
+        might get more masks than that by chance, or none at all when all masks
+        fall into the padding space.
+    :param time_min_num_masks: minimum number of masks along T
+    :param time_mask_max_size: maximum size of masks along T
+    :param freq_min_num_masks: minimum number of masks along F
+    :param freq_max_num_masks: maximum number of masks along F
+    :param freq_mask_max_size: maximum size of masks along F
+    :return: masked audio features
+    """
+    return returnn_specaugment(
```
Atticus1806, line 106:
```suggestion
    return zero_specaugment(
```

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,114 @@
...
+    :param freq_mask_max_size: maximum size of masks along F
+    :return: masked audio features
+    """
+    assert len(tensor.shape) == 3
+    tensor = _random_mask(audio_features, 0, 1, 2, time_num_masks, time_mask_max_size)  # time masking
+    tensor = _random_mask(audio_features, 0, 2, 2, freq_num_masks, freq_mask_max_size)  # freq masking
+    return tensor
+
+
+def zero_specaugment_by_length(
```
Atticus1806, line 82: Since I assume this is the function to be called, can you add type hints here please?

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,114 @@
...
+
+    z = -torch.log(-torch.log(torch.rand((batch_dim, tensor.shape[axis])).to(device=tensor.device)))  # [B,dim]
+    _, indices = torch.topk(z, num_masks.max().item(), dim=1)
+
+    for i in range(num_masks.max().item()):
+        tensor = _mask(tensor, batch_axis, axis, indices[:, i], max_len)
+    return tensor
+
+
+def zero_specaugment(
```
Atticus1806, line 52: I am not sure, do we really need this wrapper? This seems like a lot of code for one assert and two calls

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,114 @@
...
+    :param time_min_num_masks: minimum number of masks along T
+    :param time_mask_max_size: maximum size of masks along T
+    :param freq_min_num_masks: minimum number of masks along F
+    :param freq_max_num_masks: maximum number of masks along F
+    :param freq_mask_max_size: maximum size of masks along F
+    :return: masked audio features
+    """
+    return returnn_specaugment(
+        audio_features,
+        time_num_masks=audio_features.size(1) // time_mask_per_n_frames,
```
Atticus1806, line 108: I assume this is time_max_num_masks?

Atticus1806 (changes requested):

i6_models/primitives/specaugment.py
JackTemaki: As it is written, it is just a convenience wrapper for the parameter naming. But as we used specaugment like that before, I would keep that.

Atticus1806 (approved):

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,118 @@
...
+    freq_mask_max_size: int,
+):
+    """
+    Convenience wrapper around zero_specaugment with time-length adaptive number of masks
+
+    :param audio_features: e.g. log-mel features as [B, T, F]
+    :param time_mask_per_n_frames: maximum number of masks depending on length T.
+        They are still drawn depending on the full batch length, so shorter sequences
+        might get more masks than that by chance, or none at all when all masks
+        fall into the padding space.
```
michelwi, line 102: The naming and documentation is not clear to me.

I would go with `mask_per_frame_ratio` or `frame_per_mask_ratio` (I prefer the first one, but then the implementation should be inverted and that might conflict with the expectations of people that have historically used this code...)

and then the description should be something  along "Insert (on average / at maximum) 1 mask for every frame_per_mask_ratio frames in the audio"

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,118 @@
...
+    :param time_mask_max_size: maximum size of masks along T
+    :param freq_min_num_masks: minimum number of masks along F
+    :param freq_max_num_masks: maximum number of masks along F
+    :param freq_mask_max_size: maximum size of masks along F
+    :return: masked audio features
+    """
+    return zero_specaugment(
+        audio_features,
+        time_min_num_masks=time_min_num_masks,
+        time_max_num_masks=audio_features.size(1) // time_mask_per_n_frames,
```
michelwi, line 113: Is it somewhere guaranteed that `min_num_masks is <= max_num_masks` ?
What happens if it is not? (e.g. `torch.randint` fails?)

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,118 @@
...
+
+    batch_dim = tensor.shape[batch_axis]
+    num_masks = torch.randint(min_num, max_num, size=(batch_dim,)).to(device=tensor.device)  # [B]
+
+    z = -torch.log(-torch.log(torch.rand((batch_dim, tensor.shape[axis])).to(device=tensor.device)))  # [B,dim]
+    _, indices = torch.topk(z, num_masks.max().item(), dim=1)
+
+    for i in range(num_masks.max().item()):
+        tensor = _mask(tensor, batch_axis, axis, indices[:, i], max_len)
+    return tensor
```
SimBe195, line 49: There are several things about this code that I don't understand.

First of all, `num_masks` has shape [B] but only the maximum value is ever used. This makes it so that the sampling is no longer uniform between `min_num` and `max_num`; for larger B the expected value will get closer to `max_num`. I would just use shape [1] and remove the `.max()` calls. Also it doesn't seem to be necessary that this is a tensor, let alone on-device, since the value is always retrieved via `.item()`
```
    import random
    num_masks = random.randint(min_num, max_num)
```

Next, the construction of `z` and `indices` makes no sense to me. The two `-torch.log` transformations don't change the result indices of `top_k` in `z` so they can just be removed. And then, since `z` just consists of random numbers anyway, the indices are also random. So why not just set `indices` randomly directly?
```
    indices = [random.sample(range(tensor.shape[axis]), num_masks) for _ in range(batch_dim)]
    indices = torch.tensor(indices, device=tensor.device)
```
So overall I would change this to:
```suggestion
    import random

    num_masks = random.randint(min_num, max_num)

    indices = [random.sample(range(tensor.shape[axis]), num_masks) for _ in range(batch_dim)]  # [B, num_masks]
    indices = torch.tensor(indices, device=tensor.device)

    for i in range(num_masks):
        tensor = _mask(tensor, batch_axis, axis, indices[:, i], max_len)
    return tensor
```

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    :param tensor: e.g. [B, ..., A, ...] but arbitrary axis order
+    :param batch_axis: index of the batch axis
+    :param axis: which axis A to mask
+    :param pos: at which positions along axis to start the mask (size [B])
+    :param max_len: mask length drawn uniformly from [0, max_len]
+    """
+    batch_dim = tensor.shape[batch_axis]
+    dim = tensor.shape[axis]
+    amount = torch.randint(low=1, high=max_len + 1, size=(batch_dim,), dtype=torch.int32).to(device=tensor.device)
```
SimBe195, line 14:
```suggestion
    amount = torch.randint(low=1, high=max_len + 1, size=(batch_dim,), dtype=torch.int32, device=tensor.device)
```

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,118 @@
...
+    pos2 = torch.min(pos + amount, torch.tensor([dim] * batch_dim).to(device=tensor.device))
```
SimBe195, line 15:
```suggestion
    pos2 = torch.min(pos + amount, torch.tensor([dim] * batch_dim, device=tensor.device))
```

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,118 @@
...
+    idxs = torch.arange(0, dim).to(device=tensor.device).unsqueeze(0)  # [1,dim]
```
SimBe195, line 16:
```suggestion
    idxs = torch.arange(0, dim, device=tensor.device).unsqueeze(0)  # [1,dim]
```

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,118 @@
...
+    pos_bc = pos.unsqueeze(1)  # [B,1]
+    pos2_bc = pos2.unsqueeze(1)  # [B,1]
+    cond = torch.logical_and(torch.greater_equal(idxs, pos_bc), torch.less(idxs, pos2_bc))  # [B,dim]
+    if batch_axis > axis:
+        cond = cond.transpose(0, 1)  # [dim,B]
+    cond = torch.reshape(
+        cond, shape=[tensor.shape[i] if i in (batch_axis, axis) else 1 for i in range(len(tensor.shape))]
```
SimBe195, line 23:
```suggestion
        cond, shape=[tensor.shape[i] if i in {batch_axis, axis} else 1 for i in range(len(tensor.shape))]
```

SimBe195 (changes requested):

i6_models/primitives/specaugment.py
JackTemaki: @mmz33 this is a question for you

i6_models/primitives/specaugment.py
JackTemaki: I like frame per mask more because then you stay in the integer range... This is why I avoided ratio because for me this implies float. But of course we can change it to work like that. The original idea was just to copy the old code to PyTorch, but I am also not against reformulating it. But then there are additional changes I would like to add, like a flag for mean instead of zero replacement.

michelwi: By all means keep the integer and behavior as is i.e. frame_per_mask . I also do not insist on the word "ratio" in the variable name. 
I just found it confusing as by the description "maximum number of masks depending on .." I would have put in the maximum number of masks. But actually we want the number of frames per which a new mask might be inserted..

JackTemaki: Ah I understand, I will try to come up with something better

i6_models/primitives/specaugment.py
JackTemaki: I will add an assert!

i6_models/primitives/specaugment.py
mmz33: regarding sampling part, this was implemented by @albertz and he provided these links for reference:
* https://github.com/tensorflow/tensorflow/issues/9260                                                                                                                                                    
* https://timvieira.github.io/blog/post/2014/08/01/gumbel-max-trick-and-weighted-reservoir-sampling/   
```
This is also the TF code:                                                                                                   
z = -tf.log(-tf.log(tf.random_uniform((n_batch, tf.shape(x)[axis]), 0, 1)))                                                                                                                               
  _, indices = tf.nn.top_k(z, num if isinstance(num, int) else tf.reduce_max(num))  
```
So i understand correctly, this is a better sampling method but i am not sure about its influence here. I guess @albertz can comment on this.

albertz: So the original code is this (as you find in the linked TF issue):
```python
z = -tf.log(-tf.log(tf.random_uniform(tf.shape(logits),0,1)))
_, indices = tf.nn.top_k(logits + z,K)
```

We also use the same code in the TF `ChoiceLayer` when you want to randomly sample from the distribution instead of taking the top_k.

I then removed the `logits` because I wanted it to be random sampling without replacement from a uniform distribution (logits = 0 everywhere).

I think I just kept the `-log(-log(..))` to keep it clear where this comes from and to potentially add some logits again in case you don't want to sample from uniform.

Now, without logits (or logits=0), sure you could remove the `-log(-log(..))`.

albertz (changes requested):

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,118 @@
...
+
+    z = -torch.log(-torch.log(torch.rand((batch_dim, tensor.shape[axis])).to(device=tensor.device)))  # [B,dim]
+    _, indices = torch.topk(z, num_masks.max().item(), dim=1)
+
+    for i in range(num_masks.max().item()):
+        tensor = _mask(tensor, batch_axis, axis, indices[:, i], max_len)
+    return tensor
+
+
+def zero_specaugment(
```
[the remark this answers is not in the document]
JackTemaki: I actually named it `returnn_specaugment` in my setups, but then it was never part of RETURNN itself, so this morning I named it `zero_specaugment`. I can surely go back to `returnn_specaugment` or `specaugment_returnn`.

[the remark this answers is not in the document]
albertz: Or just `specaugment_v1`?

i6_models/primitives/specaugment.py
JackTemaki: > I will try to come up with something better

I have to say I do not really like the alternatives...

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+
+    batch_dim = tensor.shape[batch_axis]
+    num_masks = torch.randint(min_num, max_num, size=(batch_dim,), device=tensor.device)  # [B]
+
+    z = torch.rand((batch_dim, tensor.shape[axis]), device=tensor.device)  # [B,dim]
+    _, indices = torch.topk(z, num_masks.max().item(), dim=1)
+
+    for i in range(num_masks.max().item()):
+        tensor = _mask(tensor, batch_axis, axis, indices[:, i], max_len)
```
albertz, line 48: I just notice, this is not like in the original implementation, where we do:
```python
_, x = tf.while_loop(
    cond=lambda i, _: tf.less(i, tf.reduce_max(num)),
    body=lambda i, x: (
        i + 1,
        tf.where(             # <
            tf.less(i, num),  # <--- this here is missing
            _mask(x, batch_axis=batch_axis, axis=axis, pos=indices[:, i], max_amount=max_dims, mask_value=mask_value),
            x)),
    loop_vars=(0, x))
```

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,123 @@
...
+    *,
+    time_min_num_masks: int,
+    time_max_mask_per_n_frames: int,
+    time_mask_max_size: int,
+    freq_min_num_masks: int,
+    freq_max_num_masks: int,
+    freq_mask_max_size: int,
+):
+    """
+    Convenience wrapper around zero_specaugment with time-length adaptive number of masks
```
albertz, line 100:
```suggestion
    Convenience wrapper around specaugment_v1 with time-length adaptive number of masks
```

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,123 @@
...
+    time_min_num_masks: int,
+    time_max_num_masks: int,
+    time_mask_max_size: int,
+    freq_min_num_masks: int,
+    freq_max_num_masks: int,
+    freq_mask_max_size: int,
+):
+    """
+    Specaugment from legacy rossenbach/zeineldeen/zeyer attention setups (usually called specaugment_v2.py or so),
+    but without any step-based scheduling. Fills masks with zeros.
```
albertz, line 64:
```suggestion
    but without any step-based scheduling
    and without dependence on length.
    See `specaugment_v1_by_length` for a variant which is more close to the original.
    
    Fills masks with zeros.
```

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,123 @@
...
+    *,
+    time_min_num_masks: int,
+    time_max_num_masks: int,
+    time_mask_max_size: int,
+    freq_min_num_masks: int,
+    freq_max_num_masks: int,
+    freq_mask_max_size: int,
+):
+    """
+    Specaugment from legacy rossenbach/zeineldeen/zeyer attention setups (usually called specaugment_v2.py or so),
```
albertz, line 63: I think we should put some link to returnn-experiments or so for reference.

albertz, line 63: SpecAugment maybe with ref (https://arxiv.org/abs/1904.08779).

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,123 @@
...
+    :param tensor: e.g. [B, ..., A, ...] but arbitrary axis order
+    :param batch_axis: index of the batch axis
+    :param axis: which axis to mask
+    :param min_num: minimum number of masks
+    :param max_num: maximum number of masks
+    :param max_amount: mask length drawn uniformly from [0, max_amount]
+    """
+
+    batch_dim = tensor.shape[batch_axis]
+    num_masks = torch.randint(min_num, max_num, size=(batch_dim,), device=tensor.device)  # [B]
```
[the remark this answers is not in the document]
SimBe195: I didn't check the old implementation again, I just assumed that this was a replication.

i6_models/primitives/specaugment.py
SimBe195: Yes, that makes much more sense. With this condition the actual values in `num_masks` will be used and not just the max. Something like this would be the torch equivalent:
```
    # Make num_masks broadcastable to shape of tensor for torch.where.
    for _ in range(tensor.dim() - 1):
        num_masks = num_masks.unsqueeze(-1)

    for i in range(num_masks.max().item()):
        tensor = torch.where(
            i < num_masks,
            _mask(tensor, batch_axis, axis, indices[:, i], max_len),
            tensor
        )
```

i6_models/primitives/specaugment.py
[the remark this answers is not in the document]
JackTemaki: I do not understand what you are saying @SimBe195. It is true that the maximum number of masks is determined by the longest sequence in the batch, so a shorter sequence might get more masks if there is another long sequence in the batch. But the average will stay the same, because a mask can also be put into the padding area. This means that in the "expected" case nothing changes.

i6_models/primitives/specaugment.py
JackTemaki: Ah, did not read this part yet, now I understand.

i6_models/primitives/specaugment.py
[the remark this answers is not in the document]
SimBe195: Currently, each sequence in the batch has `num_masks.max()` masks applied to it which is wrong and as it turns out also different from the original tensorflow implementation, so this must be fixed. This is addressed in [this thread](https://github.com/rwth-i6/i6_models/pull/34/files#r1318537237).

i6_models/primitives/specaugment.py
albertz: I'm not sure if broadcasting is correct in the code, but otherwise, yes.

i6_models/primitives/specaugment.py
albertz, line 42:
```suggestion
    num_masks = torch.randint(min_num, max_num, size=(batch_dim,))  # [B]
    max_num_masks = num_masks.max().item()
    num_masks = num_masks.to(tensor.device)
```
And below, you use `max_num_masks`.

Otherwise, I wonder a bit that the `range(num_masks.max().item())` even worked, because if that is on GPU, I thought you always need an explicit to-CPU copy, it would never automatically do that, so `range(num_masks.max().item())` should not work.

JackTemaki: @mmz33 for you to note

i6_models/primitives/specaugment.py
JackTemaki: this broadcasting is not correct if batch_axis != 0, but otherwise this should be ok.

SimBe195: > this broadcasting is not correct if batch_axis != 0, but otherwise this should be ok.

Ah that's true. In the general case
```
    for _ in range(batch_axis, tensor.dim() - 1):
        num_masks = num_masks.unsqueeze(-1)
```
should do the trick.

i6_models/primitives/specaugment.py
albertz: Well, later for the `torch.where`, we need to have in on GPU at some point.

SimBe195 (approved):

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,141 @@
...
+
+    z = torch.rand((batch_dim, tensor.shape[axis]), device=tensor.device)  # [B,dim]
+    _, indices = torch.topk(z, num_masks.max().item(), dim=1)
+
+    # Make num_masks broadcastable to shape of tensor for torch.where.
+    for i in range(tensor.dim() - 1):
+        if i < batch_axis:
+            num_masks = num_masks.unsqueeze(0)
+        else:
+            num_masks = num_masks.unsqueeze(-1)
```
albertz, line 54: Way too complicated and unintuitive.
```suggestion
    num_masks = torch.reshape(num_masks, [1] * batch_axis + [batch_dim] + [1] * (tensor.dim() - batch_axis - 1))
```

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,141 @@
...
+    *,
+    time_min_num_masks: int,
+    time_max_mask_per_n_frames: int,
+    time_mask_max_size: int,
+    freq_min_num_masks: int,
+    freq_max_num_masks: int,
+    freq_mask_max_size: int,
+):
+    """
+    Convenience wrapper around zero_specaugment with time-length adaptive number of masks.
```
albertz, line 118:
```suggestion
    Convenience wrapper around specaugment_v1 with time-length adaptive number of masks.
```

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,141 @@
...
+    :param max_amount: mask length drawn uniformly from [0, max_amount]
+    """
+
+    batch_dim = tensor.shape[batch_axis]
+    num_masks = torch.randint(min_num, max_num, size=(batch_dim,))  # [B]
+
+    max_num_masks = num_masks.max().item()
+
+    z = torch.rand((batch_dim, tensor.shape[axis]), device=tensor.device)  # [B,dim]
+    _, indices = torch.topk(z, num_masks.max().item(), dim=1)
```
albertz, line 47:
```suggestion
    _, indices = torch.topk(z, max_num_masks, dim=1)
```

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,141 @@
...
+    :param tensor: e.g. [B, ..., A, ...] but arbitrary axis order
+    :param batch_axis: index of the batch axis
+    :param axis: which axis to mask
+    :param min_num: minimum number of masks
+    :param max_num: maximum number of masks
+    :param max_amount: mask length drawn uniformly from [0, max_amount]
+    """
+
+    batch_dim = tensor.shape[batch_axis]
+    num_masks = torch.randint(min_num, max_num, size=(batch_dim,))  # [B]
```
albertz, line 42:
```suggestion
    num_masks = torch.randint(min_num, max_num, size=(batch_dim,), device="cpu")  # [B]
```
Otherwise the PyTorch 2 default device placement logic might place it on GPU, but I guess we really intent to place it on CPU here.

i6_models/primitives/specaugment.py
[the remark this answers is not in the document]
albertz: @SimBe195 So this is resolved?

vieting:
Is the rationale to be as close as possible to the old implementation or just to replicate the behavior as good as possible?

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,137 @@
...
+    amount = torch.randint(low=1, high=max_len + 1, size=(batch_dim,), dtype=torch.int32, device=tensor.device)
+    pos2 = torch.min(pos + amount, torch.tensor([dim] * batch_dim, device=tensor.device))
+    idxs = torch.arange(0, dim, device=tensor.device).unsqueeze(0)  # [1,dim]
+    pos_bc = pos.unsqueeze(1)  # [B,1]
+    pos2_bc = pos2.unsqueeze(1)  # [B,1]
+    cond = torch.logical_and(torch.greater_equal(idxs, pos_bc), torch.less(idxs, pos2_bc))  # [B,dim]
+    if batch_axis > axis:
+        cond = cond.transpose(0, 1)  # [dim,B]
+    cond = torch.reshape(
+        cond, shape=[tensor.shape[i] if i in (batch_axis, axis) else 1 for i in range(len(tensor.shape))]
```
vieting, line 23:
```suggestion
        cond, shape=[tensor.shape[i] if i in (batch_axis, axis) else 1 for i in range(tensor.ndim)]
```

i6_models/primitives/specaugment.py
```diff
@@ -0,0 +1,137 @@
...
+    :param tensor: e.g. [B, ..., A, ...] but arbitrary axis order
+    :param batch_axis: index of the batch axis
+    :param axis: which axis to mask
+    :param min_num: minimum number of masks
+    :param max_num: maximum number of masks
+    :param max_amount: mask length drawn uniformly from [0, max_amount]
+    """
+
+    batch_dim = tensor.shape[batch_axis]
+    num_masks = torch.randint(min_num, max_num, size=(batch_dim,), device="cpu")  # [B]
```
vieting, line 42: In the old SpecAugment implementation I'm looking at, we have
```
num = tf.random_uniform(shape=(n_batch,), minval=min_num, maxval=max_num + 1, dtype=tf.int32)
```
Note the `max_num + 1`. Doesn't this deviate from the behavior here?

vieting (commented):
Looks mostly good to me, I'm just curious about the maximum + 1 or not.

I think you changed a number of variable names, right? Imho, there are some remaining with suboptimal names (`pos`, `pos2`, maybe also `cond`). If the goal is not to stay as close to the original implementation as possible, I'd suggest to improve them as well.

i6_models/primitives/specaugment.py
JackTemaki: @mmz33 ? but yes I am quite sure it does. This should be corrected, good catch

JackTemaki:
> Is the rationale to be as close as possible to the old implementation or just to replicate the behavior as good as possible?

In my opinion it could also be only replicating, so yes I can go over the code and use better names.

vieting (approved):
Thanks for fixing the `max_num` issue!

I'm alright with merging, I just still don't quite get your point here:



 > > Is the rationale to be as close as possible to the old implementation or just to replicate the behavior as good as possible?
>
> In my opinion it could also be only replicating, so yes I can go over the code and use better names.

In case of only replicating, I'd keep the variable names consistent with the old version. In case of updating/improving, I'd use better names. Anyway, this is not too critical and I'm fine with merging if others don't mind this.

PR #35, opened 2023-09-06
Add mask_tensor util

JackTemaki:
Utility to create a boolean tensor mask on the tensors device based on sequence lengths.

### DIFF
--- a/i6_models/util/mask.py
+++ b/i6_models/util/mask.py
@@ -1,0 +1,16 @@
+import torch
+
+
+def mask_tensor(tensor: torch.Tensor, seq_len: torch.Tensor) -> torch.Tensor:
+    """
+    Create a "positive" mask for a tensor (boolean true means position is used)
+    on the same device as the tensor.
+
+    :param tensor: [B,T,....]
+    :param seq_len: [B]
+    :return: Mask of [B,T]
+    """
+    seq_len = seq_len.to(device=tensor.device)
+    r = torch.arange(tensor.shape[1], device=tensor.device)  # [T]
+    seq_mask = torch.less(r[None, :], seq_len[:, None])  # broadcast to [B,T]
+    return seq_mask


i6_models/util/mask.py
```diff
@@ -0,0 +1,16 @@
...
+def mask_tensor(tensor: torch.Tensor, seq_len: torch.Tensor) -> torch.Tensor:
+    """
+    Create a "positive" mask for a tensor (boolean true means position is used)
+    on the same device as the tensor.
+
+    :param tensor: [B,T,....]
+    :param seq_len: [B]
+    :return: Mask of [B,T]
+    """
+    seq_len = seq_len.to(device=tensor.device)
```
Atticus1806, line 13: I am not sure if this case should be made within this function.

SimBe195 (approved):

i6_models/util/mask.py
JackTemaki: The device handling? So completely removing tensor and having just a seq_len to mask function?

Atticus1806: you still need the tensor for the shape right? But I feel like this casting of the seq_len should be done beforehand (and at least in my setups usually already is done with the model call. So this might be redundant imho.

JackTemaki: You could do it without the tensor (using max(seq_len)), I think we can add this as a second function

JackTemaki: If seq_len is already on the device, this should have not more overhead than one integer comparison

Atticus1806: is this max then exportable? 
In general if the overhead is low I don't mind and we can leave it like this.

i6_models/util/mask.py
```diff
@@ -0,0 +1,16 @@
+import torch
+
+
+def mask_tensor(tensor: torch.Tensor, seq_len: torch.Tensor) -> torch.Tensor:
```
albertz, line 4: I'm not sure about the name. I find it a bit too generic. And also ambiguous. (Does it mask a tensor? No, actually, it returns a mask tensor. For what? Based on what?)

In TF, the corresponding function is called `tf.sequence_mask`. Why not just use the same name here? So all people who have used TF at some point in the past (almost everyone) knows what this is about.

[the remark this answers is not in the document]
JackTemaki: > What you actually want

What I want is to have it with a tensor and make sure the mask is on the same device as the tensor.  But maybe the majority disagrees, and I keep this for myself.

JackTemaki: tf.sequence_mask has a different signature, and we can also add such a function. This is just not what I wanted to have here.

JackTemaki: For the name, yes this can be more verbose if needed.

JackTemaki:
Closing, as this specific signature seems not to be beneficial for people other than me.

PR #37, opened 2023-10-12
SpecAugment: max_num_masks is at least min_num_masks

JackTemaki:
I triggered one "last batch" in Tedlium that had only one sequence that was apparently so short that max_masks got smaller than min_masks for the time.

### DIFF
--- a/i6_models/primitives/specaugment.py
+++ b/i6_models/primitives/specaugment.py
@@ -1,3 +1,4 @@
+import numpy as np
 import torch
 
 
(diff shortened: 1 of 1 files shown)


kuacakuaca (approved):

i6_models/primitives/specaugment.py
```diff
@@ -1,3 +1,4 @@
+import numpy as np
```
albertz, line 1: We mostly use `import numpy` in other code (consistent to `import torch`), here also in the repo (although not too much yet), but also in RETURNN.
Although this variant here is more the official recommended way.

So maybe for i6_models, we can independently decide now what variant to pick and then always keep it consistent.

albertz: You merged without resolving this here, or commenting here. So what do you mean?

albertz: Ah now I see #38.

JackTemaki: As this is an independent non-critical issue, I would not delay this PR by that. If we decide different to this PR we can just change it back.

PR #38, opened 2023-10-12
change to import numpy as np

JackTemaki:
See https://github.com/rwth-i6/i6_models/pull/37#discussion_r1356588208

We should have the numpy import consistent, and I would go with the established default of `import numpy as np`, although some of our other repos might have that still different.

### DIFF
--- a/tests/test_feature_extraction.py
+++ b/tests/test_feature_extraction.py
@@ -1,5 +1,5 @@
 import copy
-import numpy
+import numpy as np
 import torch
 
 from librosa.feature import melspectrogram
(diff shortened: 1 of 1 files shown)


kuacakuaca (approved):

albertz:
I always preferred `import numpy` over `import numpy as np`, as it is more readable to me. It's also more consistent to `import torch` (but inconsistent to `import tensorflow as tf`, although you could argue that `tensorflow` is longer than `torch`, which has the same length as `numpy`).

christophmluscher (approved):

PR #40, opened 2023-11-15
Extend feature extraction module to allow for RASR compatible logmel features

kuacakuaca:


i6_models/primitives/feature_extraction.py
```diff
@@ -72,32 +80,42 @@ def __init__(self, cfg: LogMelFeatureExtractionV1Config):
...
-                torch.stft(
-                    raw_audio,
-                    n_fft=self.n_fft,
-                    hop_length=self.hop_length,
-                    win_length=self.win_length,
-                    window=self.window,
-                    center=self.center,
-                    pad_mode="constant",
-                    return_complex=True,
+        if not self.rasr_compatible:
```
albertz, line 95: It is usually not recommended to do:
```python
if not <cond>:
    <false_case>
else:
    <true_case>
```
Instead, you should do:
```python
if <cond>:
    <true_case>
else:
    <false_case>
```

i6_models/primitives/feature_extraction.py
```diff
@@ -22,6 +22,9 @@ class LogMelFeatureExtractionV1Config(ModelConfiguration):
         min_amp: minimum amplitude for safe log
         num_filters: number of mel windows
         center: centered STFT with automatic padding
+        periodic: whether the window is assumed to be periodic
+        mel_options: extra options for mel filters
+        rasr_compatible: apply FFT to make features compatible to RASR's
```
albertz, line 27: You should add sth like:
```suggestion
        rasr_compatible: apply FFT to make features compatible to RASR's,
           otherwise (default) apply STFT
```

i6_models/primitives/feature_extraction.py
```diff
@@ -33,6 +36,9 @@ class LogMelFeatureExtractionV1Config(ModelConfiguration):
     num_filters: int
     center: bool
     n_fft: Optional[int] = None
+    periodic: Optional[bool] = True
```
albertz, line 39: I think the type is wrong?
```suggestion
    periodic: bool = True
```

i6_models/primitives/feature_extraction.py
```diff
@@ -33,6 +36,9 @@ class LogMelFeatureExtractionV1Config(ModelConfiguration):
...
+    mel_options: Optional[Dict[str, Any]] = None
+    rasr_compatible: Optional[bool] = False
```
albertz, line 41:
```suggestion
    rasr_compatible: bool = False
```

i6_models/primitives/feature_extraction.py
```diff
@@ -72,32 +80,42 @@ def __init__(self, cfg: LogMelFeatureExtractionV1Config):
...
-                    raw_audio,
-                    n_fft=self.n_fft,
-                    hop_length=self.hop_length,
-                    win_length=self.win_length,
-                    window=self.window,
-                    center=self.center,
-                    pad_mode="constant",
-                    return_complex=True,
+        if self.rasr_compatible:
+            windowed = raw_audio.unfold(1, size=self.win_length, step=self.hop_length)
```
albertz, line 96: Can you comment the shapes? I think this would be:
```suggestion
            windowed = raw_audio.unfold(1, size=self.win_length, step=self.hop_length)  # [B, W, T']
```
?

albertz:
I'm curious, did you actually test this, that you really get identical results to RASR (up to numerical errors)

kuacakuaca:
> I'm curious, did you actually test this, that you really get identical results to RASR (up to numerical errors)

there were still small differences. Not sure if that could be further eliminated.
```
3.922669 4.240274 4.526385 5.040951 4.915504 4.856473 5.144744 <- rasr
3.9254   4.2424   4.5246   5.0411   4.9170   4.8564   5.1468   <- torch module
```

albertz:
Oh, but that looks very good. Can you maybe link the relevant RASR code for comparison?

kuacakuaca:
> Oh, but that looks very good. Can you maybe link the relevant RASR code for comparison?

sure.
https://github.com/rwth-i6/rasr/blob/a3117a2fe4599ea7a371eeefc3d60ba85d17291f/src/Signal/FastFourierTransform.cc#L56

i6_models/primitives/feature_extraction.py
```diff
@@ -22,6 +22,9 @@ class LogMelFeatureExtractionV1Config(ModelConfiguration):
         min_amp: minimum amplitude for safe log
         num_filters: number of mel windows
         center: centered STFT with automatic padding
+        periodic: whether the window is assumed to be periodic
+        mel_options: extra options for mel filters
+        rasr_compatible: apply FFT to make features compatible to RASR's, otherwise (defalut) apply STFT
```
albertz, line 27:
```suggestion
        rasr_compatible: apply FFT to make features compatible to RASR's, otherwise (default) apply STFT
```

i6_models/primitives/feature_extraction.py
```diff
@@ -72,42 +80,52 @@ def __init__(self, cfg: LogMelFeatureExtractionV1Config):
                     n_mels=cfg.num_filters,
                     fmin=cfg.f_min,
                     fmax=cfg.f_max,
+                    **self.mel_options,
```
curufinwe, line 83: Instead of adding arbitrary *additional* options we could explicitly add all remaining options here. That should be `htk`, `norm` and `dtype`.

i6_models/primitives/feature_extraction.py
```diff
@@ -22,6 +22,9 @@ class LogMelFeatureExtractionV1Config(ModelConfiguration):
         min_amp: minimum amplitude for safe log
         num_filters: number of mel windows
         center: centered STFT with automatic padding
+        periodic: whether the window is assumed to be periodic
+        mel_options: extra options for mel filters
+        rasr_compatible: apply FFT to make features compatible to RASR's, otherwise (default) apply STFT
```
curufinwe, line 27: instead of making this a bool, I would suggest this should be an enum:
```
class SpectrumType(Enum):
    PYTORCH_STFT=1
    RASR_COMPATIBLE=2
```

curufinwe (changes requested):

i6_models/primitives/feature_extraction.py
```diff
@@ -72,42 +80,52 @@ def __init__(self, cfg: LogMelFeatureExtractionV1Config):
...
         log_melspec = torch.log10(torch.clamp(melspec, min=self.min_amp))
-        feature_data = torch.transpose(log_melspec, 1, 2)
+        feature_data = torch.transpose(log_melspec, 1, 2)  # [B, T', F']
 
-        if self.center:
+        if self.center and not self.rasr_compatible:
             length = (length // self.hop_length) + 1
         else:
-            length = ((length - self.n_fft) // self.hop_length) + 1
+            length = ((length - self.win_length) // self.hop_length) + 1
```
albertz, line 129: This logic looks suspicious to me. The else branch before was only for `self.center == False`, however, now the else branch can also execute when `self.center == True` (when `self.rasr_compatible == True`). Is this right?

Also, why the change of `n_fft` to `win_length`? Isn't that different from before?

i6_models/primitives/feature_extraction.py
albertz: Why `PYTORCH_STFT`? This is not at all specific about PyTorch?

Either we describe here better what on the actual difference type, spectrum type, then this is here STFT vs FFT (and not about PyTorch vs RASR or so), and mention just in the docs that RASR uses FFT.

Or we describe here on common known libraries which behave this way. But then this is more like librosa vs RASR.

curufinwe: The function called below is torch.stft, which is based on librosa stft, but not a drop-in replacement according to the documentation: https://pytorch.org/docs/stable/generated/torch.stft.html . So PYTORCH_STFT is justified in my opinion. The other option could be named RASR_COMPATIBLE_RFFTN if you prefer that.

albertz: All Torch functions start with `torch.`.

curufinwe: Yes, but the STFT implementation seems to be pytorch specific and not quite compatible with what other libraries do (librosa in this case). That's why I think it's OK to highlight that it's pytorch specific.

albertz: I actually investigated the exact differences on STFT a while ago, between TensorFlow, PyTorch, SciPy and librosa. See for example https://github.com/librosa/librosa/issues/1701 or https://github.com/pytorch/pytorch/issues/100177. They all have slightly different behavior w.r.t. padding. But you can always use any of those STFT functions and perfectly reproduce the behavior of any other STFT function, because they ultimately do the same thing.

curufinwe: OK, so what enum values would you prefer to have?

albertz: I don't really remember anymore, didn't we implement this in a way that we actually exactly get the librosa behavior? In that case, I would maybe use:
```python
class SpectrumType(Enum):
    LIBROSA_STFT=1
    RASR_FFT=2
```
Or is this independent and just our own new implementation? In that case, I would just use:
```python
class SpectrumType(Enum):
    STFT=1
    FFT=2
```
And if we want to have other aspects configurable about it (like type of windowing, e.g. Hanning or sth else, or the kind of padding, etc), I think that would be separate arguments.

curufinwe: How about STFT and RFTTN (since these are the 2 pytorch functions we call)?

curufinwe: I don't like FFT alone, because that's yet another function in pytorch which will give you the the full FFT output, whereas RFFT will only give you the non-redundant output from the Fourier transform.

i6_models/primitives/feature_extraction.py
```diff
@@ -115,6 +120,13 @@ def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
                 )
                 ** 2
             )
+        elif self.spectrum_type == SpectrumType.RFFTN:
+            windowed = raw_audio.unfold(1, size=self.win_length, step=self.hop_length)  # [B, T', W=win_length]
+            smoothed = windowed * self.window.unsqueeze(0)  # [B, T', W]
+
+            # Compute power spectrum using torch.fft.rfftn
+            power_spectrum = torch.abs(torch.fft.rfftn(smoothed, s=self.n_fft)) ** 2  # [B, T', F=n_fft//2+1]
+            power_spectrum = power_spectrum.transpose(1, 2)  # [B, F, T']
```
albertz, line 129: It should always have another `else:` branch raising an exception (`ValueError` "invalid spectrum type" or so).

i6_models/primitives/feature_extraction.py
```diff
@@ -123,9 +135,12 @@ def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
...
-        if self.center and not self.rasr_compatible:
-            length = (length // self.hop_length) + 1
-        else:
+        if self.spectrum_type == SpectrumType.STFT:
+            if self.center:
+                length = (length // self.hop_length) + 1
+            else:
+                length = ((length - self.n_fft) // self.hop_length) + 1
+        elif self.spectrum_type == SpectrumType.RFFTN:
             length = ((length - self.win_length) // self.hop_length) + 1
```
albertz, line 144: Same here, another `else:` raising an exception.

i6_models/primitives/feature_extraction.py
```diff
@@ -72,42 +90,60 @@ def __init__(self, cfg: LogMelFeatureExtractionV1Config):
...
-        )
+        elif self.spectrum_type == SpectrumType.RFFTN:
+            windowed = raw_audio.unfold(1, size=self.win_length, step=self.hop_length)  # [B, T', W=win_length]
+            smoothed = windowed * self.window.unsqueeze(0)  # [B, T', W]
+
+            # Compute power spectrum using torch.fft.rfftn
+            power_spectrum = torch.abs(torch.fft.rfftn(smoothed, s=self.n_fft)) ** 2  # [B, T', F=n_fft//2+1]
+            power_spectrum = power_spectrum.transpose(1, 2)  # [B, F, T']
+        else:
+            raise ValueError("Invalid spectrum type.")
```
albertz, line 131:
```suggestion
            raise ValueError(f"Invalid spectrum type {self. spectrum_type!r}.")
```

i6_models/primitives/feature_extraction.py
```diff
@@ -72,42 +90,60 @@ def __init__(self, cfg: LogMelFeatureExtractionV1Config):
...
+            if self.center:
+                length = (length // self.hop_length) + 1
+            else:
+                length = ((length - self.n_fft) // self.hop_length) + 1
+        elif self.spectrum_type == SpectrumType.RFFTN:
+            length = ((length - self.win_length) // self.hop_length) + 1
         else:
-            length = ((length - self.n_fft) // self.hop_length) + 1
-
+            raise ValueError("Invalid spectrum type.")
```
albertz, line 148:
```suggestion
            raise ValueError(f"Invalid spectrum type {self. spectrum_type!r}.")
```

i6_models/primitives/feature_extraction.py
albertz: So what about this? Why this change?

kuacakuaca: I've restored that after some testing. So using `n_fft` is correct for `torch.stft`.

albertz (approved):

curufinwe (approved):

i6_models/primitives/feature_extraction.py
```diff
@@ -33,6 +45,11 @@ class LogMelFeatureExtractionV1Config(ModelConfiguration):
     num_filters: int
     center: bool
     n_fft: Optional[int] = None
+    periodic: bool = True
+    htk: bool = False
+    norm: Optional[Union[Literal["slaney"], float]] = "slaney"
+    dtype: DTypeLike = np.float32
+    spectrum_type: SpectrumType = SpectrumType.STFT
```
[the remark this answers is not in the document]
michelwi: > [curufinwe](https://github.com/curufinwe) approved these changes [8 minutes ago](https://github.com/rwth-i6/i6_models/pull/40#pullrequestreview-1769726515)
[curufinwe](https://github.com/curufinwe) merged commit [c363a01](https://github.com/rwth-i6/i6_models/commit/c363a0179aa6dfd34e48020e743c1ad3b86850b8) into main [8 minutes ago](https://github.com/rwth-i6/i6_models/pull/40#event-11179874625)

oh well 🤷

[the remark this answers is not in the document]
albertz: I don't think the defaults are problematic here.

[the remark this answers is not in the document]
JackTemaki: It is not the defaults that are problematic, but adding new parameter automatically breaks setups. I will revert the PR. And then I would be in favor of Willis RasrCompatibleLogMelFeatureExtraction instead of the V2. But I would be fine with both.

PR #41, opened 2023-12-07
Revert "Extend feature extraction module to allow for RASR compatible…

JackTemaki:
From my understanding it should be possible to use "ModelConfiguration" objects directly in Sisyphus to define the model parameters, and I did so in my setups. Thus the PR in question breaks two of my pipelines. (In the others I still had my own custom feature extraction config object). 

I assume others are affected as well, at least @Atticus1806 also is.

### DIFF
--- a/i6_models/primitives/feature_extraction.py
+++ b/i6_models/primitives/feature_extraction.py
@@ -1,21 +1,13 @@
 __all__ = ["LogMelFeatureExtractionV1", "LogMelFeatureExtractionV1Config"]
 
 from dataclasses import dataclass
-from typing import Optional, Tuple, Union, Literal
-from enum import Enum
+from typing import Optional, Tuple
 
 from librosa import filters
 import torch
 from torch import nn
-import numpy as np
-from numpy.typing import DTypeLike
 
 from i6_models.config import ModelConfiguration
-
-
-class SpectrumType(Enum):
-    STFT = 1
-    RFFTN = 2
 
 
 @dataclass
@@ -30,10 +22,6 @@ class LogMelFeatureExtractionV1Config(ModelConfiguration):
         min_amp: minimum amplitude for safe log
         num_filters: number of mel windows
         center: centered STFT with automatic padding
-        periodic: whether the window is assumed to be periodic
-        htk: whether use HTK formula instead of Slaney
-        norm: how to normalize the filters, cf. https://librosa.org/doc/main/generated/librosa.filters.mel.html
-        spectrum_type: apply torch.stft on raw audio input (default) or torch.fft.rfftn on windowed audio to make features compatible to RASR's
     """
 
     sample_rate: int
@@ -45,11 +33,6 @@ class LogMelFeatureExtractionV1Config(ModelConfiguration):
     num_filters: int
     center: bool
     n_fft: Optional[int] = None
-    periodic: bool = True
-    htk: bool = False
-    norm: Optional[Union[Literal["slaney"], float]] = "slaney"
-    dtype: DTypeLike = np.float32
-    spectrum_type: SpectrumType = SpectrumType.STFT
 
     def __post_init__(self) -> None:
         super().__post_init__()
@@ -79,7 +62,6 @@ class LogMelFeatureExtractionV1(nn.Module):
         self.min_amp = cfg.min_amp
         self.n_fft = cfg.n_fft
         self.win_length = int(cfg.win_size * cfg.sample_rate)
-        self.spectrum_type = cfg.spectrum_type
 
         self.register_buffer(
             "mel_basis",
@@ -90,13 +72,10 @@ class LogMelFeatureExtractionV1(nn.Module):
                     n_mels=cfg.num_filters,
                     fmin=cfg.f_min,
                     fmax=cfg.f_max,
-                    htk=cfg.htk,
-                    norm=cfg.norm,
-                    dtype=cfg.dtype,
-                ),
+                )
             ),
         )
-        self.register_buffer("window", torch.hann_window(self.win_length, periodic=cfg.periodic))
+        self.register_buffer("window", torch.hann_window(self.win_length))
 
     def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
         """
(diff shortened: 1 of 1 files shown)


michelwi (approved):

Atticus1806 (approved):

JackTemaki:
It is not clear yet why (Benedikt reported this should not be the case), but the default values also get converted when using `asdict`, which is used to serialize (and hash) config.

JackTemaki:
> you get a crash

Well, it also crashed because it writes invalid code:

```python
error: cannot format returnn.config: Cannot parse: 331:273:     **{'model_config_dict': {'feature_extraction_config': {'sample_rate': 16000, 'win_size': 0.025, 'hop_size': 0.01, 'f_min': 60, 'f_max': 7600, 'min_amp': 1e-10, 'num_filters': 80, 'center': False, 'n_fft': 400, 'periodic': True, 'htk': False, 'norm': 'slaney', 'dtype': <class 'numpy.float32'>, 'spectrum_type': <SpectrumType.STFT: 1>}, 'frontend_config': {'in_features': 80, 'conv1_channels': 32, 'conv2_channels': 64, 'conv3_channels': 64, 'conv4_channels': 32, 'conv_kernel_size': (3, 3), 'conv_padding': None, 'pool1_kernel_size': (2, 1), 'pool1_stride': (2, 1), 'pool1_padding': None, 'pool2_kernel_size': (2, 1), 'pool2_stride': (2, 1), 'pool2_padding': None, 'activation': None, 'out_features': 384, 'activation_str': 'ReLU'}, 'specaug_config': {'repeat_per_n_frames': 25, 'max_dim_time': 20, 'num_repeat_feat': 5, 'max_dim_feat': 16}, 'specauc_start_epoch': 11, 'label_target_size': 356, 'conformer_size': 384, 'num_layers': 12, 'num_heads': 4, 'ff_dim': 1536, 'att_weights_dropout': 0.2, 'conv_dropout': 0.2, 'ff_dropout': 0.2, 'mhsa_dropout': 0.2, 'conv_kernel_size': 31, 'final_dropout': 0.2}}
```

But this is less relevant. While I would like the ModelConfiguration objects to be trivially serializable, we already have some that are not, and I can arrange myself with that. It is just that existing ones that I use should not suddenly break.

JackTemaki:
> > I would have rejected the PR for the same reasons Willi stated as well.
> 
> Well, I disagree. The default values are exactly such that you get exactly the old behavior. These are very reasonable default parameters.

This is not only about the defaults, but that the class provides too many option combinations that are never used. Which makes it harder to understand and use. I see it exactly the other way around. It makes a class too complex and too configurable. Having one fixed class for the "normal" librosa and one fixed class for RASR should be perfectly fine, and more understandable.

Atticus1806:
Actually deleted my last comment, since I just found out its related to the typing. So if you give a type its included into asdict, if you dont its not. 
So 
```
@dataclass
class Test:
no_def: int
with_def: int = 3
```
produces 
```
{'no_def': 1, 'with_def': 3}
```
while 
```
@dataclass
class Test:
no_def: int
with_def: = 3
```
produces 
```
asdict(Test(1)) = {'no_def': 1}

albertz:
Somehow your copy-paste is broken. @Atticus1806 you miss all the indentation, this is not valid Python code. Also `with_def: = 3` does not make sense?

albertz:
> This is not only about the defaults, but that the class provides too many option combinations that are never used.

Well, @curufinwe argued to add all the other options. His had valid reasons for it. (See the discussion in the PR, it might be hidden in some of the resolved parts.)

Atticus1806:
It was in a terminal and I figured copying the upper code where I removed terminal prefix was quicker than removing terminal prefix again. Sent you the code on slack too.

JackTemaki:
> > > Well, @curufinwe argued to add all the other options. His had valid reasons for it. (See the discussion in the PR, it might be hidden in some of the resolved parts.)
> > 
> > 
> > I only found a discussion about how the arguments should look like, and that he did not like kwargs options.
> 
> I found it: [#40 (comment)](https://github.com/rwth-i6/i6_models/pull/40#discussion_r1399566079) @curufinwe said:
> 
> > Instead of adding arbitrary _additional_ options we could explicitly add all remaining options here. That should be `htk`, `norm` and `dtype`.

Yes exactly, this is about kwargs vs explicit options. He argued against arbitrary undefined options, where I fully agree with Eugen. And has nothing to do with what I mean: Not adding ANY options to the existing class, that are solely there for the purpose of getting one exact other behavior. But to have one class that will give you "exactly" the RASR behavior without any other parameters you could accidentally set wrong.  Or am I misunderstanding why the other attributes were added? So are they actually not necessary to get the RASR behavior?

albertz:
> Yes exactly, this is about kwargs vs explicit options. He argued against arbitrary undefined options, where I fully agree with Eugen. 

No, I think this is about consistency. There are some options (`f_min`, `f_max` etc) which are directly specified, and then remaining ones in `mel_options`. To have it consistent, either all options should be inside `mel_options` or all options should be explicitly listed. As we already had some listed, there was no way without breaking it to go with the more generic `mel_options`. I think this was the main reason, otherwise having such generic `mel_options` might have been fine just as well. At least this is how I understood it and why I agreed with @curufinwe's comment on this.

However, I'm not really sure it was actually discussed *why* `mel_options` (or the explicit `htk`, `norm` and `dtype`) was added though.

albertz:
> the issue that this does break setups

You mean that it changes the hash? I would still not call this "break".

But we can just fix this, or not? What is the problem in fixing this?

albertz:
Btw, about the `dataclass` example, actually this is misleading, having just `with_def = 3` without the type is not valid anyway, `dataclass` will ignore it, i.e. you cannot do `Test(no_def=1, with_def=4)`.

JackTemaki:
> > the issue that this does break setups
> 
> You mean that it changes the hash? I would still not call this "break".
> 
> But we can just fix this, or not? What is the problem in fixing this?

Do you mean from i6_models side or from the side of my setup? From the side of my setup it is annoying but possible. I can make a local copy of the ModelConfig object before the PR, use this to write the config, and then in my model code map this to the new one. But if I would possibly have to do that after each pull, I would stop using the ModelConfig options completely and always use my custom ones (which is maybe not even bad?) 

From the i6_models side, you would have to prohibit dataclasses to write new defaults when calling as_dict. For sure this is possible somehow, I am just not confident this is something we should do.

albertz:
> > You mean that it changes the hash? I would still not call this "break".
> 
> I would disagree. I think if we allow changes that change hashes

I never said that we should allow changing hashes. Definitely we should not allow this. I just said that I would not call this "break", as I understand "break" as "does not work anymore" (crash, different incorrect output, whatever). Instead of calling this "break", I would just say explicitly "this changes the hash", nothing more, nothing less. When I read "break", this confused me multiple times here already that I thought you mean sth different.

albertz:
> > You mean that it changes the hash?
> > But we can just fix this, or not? What is the problem in fixing this?
> 
> Do you mean from i6_models side

Yes of course from i6_models side. The user should not need to change any code. We should not allow any changes here which would require changes on the user side as well (at least in most cases, unless there is maybe a very good reason).

> From the i6_models side, you would have to prohibit dataclasses to write new defaults when calling as_dict. For sure this is possible somehow, I am just not confident this is something we should do.

I'm not exactly sure I understand what you mean by "new defaults", but yes, sth along this direction would be one solution. There are more potential solutions, e.g. doing sth like we do in i6_core with jobs using `__sis_hash_exclude__`.

If you don't want this, it means that we never ever can add new options to an existing config/model, and I'm not sure this is a good idea.

albertz:
> > If you don't want this, it means that we never ever can add new options to an existing config/model, and I'm not sure this is a good idea.
> 
> I thought we agreed to deprecate the old and add a new version of the classes instead of doing extensions (that go beyond bugfixing)

I thought that there would be a line somewhere between small extensions are ok to be added, big changes would go to a new version. I don't remember that we agreed that we never want any extensions?

albertz:
> I would propose to merge this revert

Instead, I would just fix the issue.

Otherwise you also want to revert other PRs now? E.g. #33?

JackTemaki:
> > I would propose to merge this revert
> 
> Instead, I would just fix the issue.
> 
> Otherwise you also want to revert other PRs now? E.g. #33?

#33 was never merged.

PR #43, opened 2023-12-08
Frontends: Add tensor info to assert

Atticus1806:
I think it makes sense to print the failed shapes already with the assert, so the user can directly see the mismatch themselves.

### DIFF
--- a/i6_models/parts/frontend/generic_frontend.py
+++ b/i6_models/parts/frontend/generic_frontend.py
@@ -195,7 +195,7 @@ class GenericFrontendV1(nn.Module):
             )
 
     def forward(self, tensor: torch.Tensor, sequence_mask: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
-        assert tensor.shape[-1] == self.cfg.in_features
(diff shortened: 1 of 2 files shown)


i6_models/parts/frontend/generic_frontend.py
```diff
@@ -195,7 +195,7 @@ def __init__(self, model_cfg: GenericFrontendV1Config):
             )
 
     def forward(self, tensor: torch.Tensor, sequence_mask: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
-        assert tensor.shape[-1] == self.cfg.in_features
+        assert tensor.shape[-1] == self.cfg.in_features, f"{tensor.shape[-1]} vs {self.cfg.in_features}"
```
albertz, line 198:
```suggestion
        assert tensor.shape[-1] == self.cfg.in_features, f"shape {tensor.shape} vs in features {self.cfg.in_features}"
```

albertz (approved):

vieting (approved):

Atticus1806:
Thanks for the quick reviews @albertz @vieting

PR #44, opened 2023-12-15
Add RASR compatible feature extraction

kuacakuaca:
one frame rasr features
`0.234030 0.731747 1.082119 1.292996 0.890861 1.056064 1.390038 1.597611 1.604475 1.586611 1.939575 2.216516 2.233468 1.941573 1.820577 2.262333 2.291739 2.248550 2.066255 2.148235 2.549040 2.568474 2.229928 2.418101 2.622112 2.478587 2.042983 2.188321 2.437973 2.706243 2.692690 2.617883 2.669614 2.743750 2.857951 2.730998 2.775461 2.747587 2.674153 3.180011 2.949069 3.098550 3.110722 2.988153 2.890325 3.153568 3.039258 3.129505 3.048017 3.061022 3.084099 3.244599 3.241169 3.604517 3.651453 3.282753 2.902287 3.145018 3.163275 2.913236`
same frame torch features
`0.2297, 0.7334, 1.0803, 1.2995, 0.9143, 1.0722, 1.3999, 1.6022, 1.6045, 1.5851, 1.9403, 2.2169, `

`2.2317, 1.9365, 1.8187, 2.2637, 2.2955, 2.2538, 2.0692, 2.1491, 2.5498, 2.5663, 2.2258, 2.4154, `

`2.6246, 2.4799, 2.0473, 2.1879, 2.4400, 2.7045, 2.6909, 2.6192, 2.6692, 2.7444, 2.8586, 2.7310, `

`2.7798, 2.7500, 2.6750, 3.1814, 2.9498, 3.1026, 3.1138, 2.9888, 2.8930, 3.1551, 3.0379, 3.1301, `

`3.0457, 3.0599, 3.0900, 3.2458, 3.2438, 3.6075, 3.6512, 3.2803, 2.9031, 3.1472, 3.1635, 2.9173`

### DIFF
--- a/i6_models/primitives/feature_extraction.py
+++ b/i6_models/primitives/feature_extraction.py
@@ -1,5 +1,11 @@
-__all__ = ["LogMelFeatureExtractionV1", "LogMelFeatureExtractionV1Config"]
+__all__ = [
+    "LogMelFeatureExtractionV1",
+    "LogMelFeatureExtractionV1Config",
+    "RasrCompatibleLogMelFeatureExtractionV1",
+    "RasrCompatibleLogMelFeatureExtractionV1Config",
+]
 
+import math
 from dataclasses import dataclass
 from typing import Optional, Tuple
 
@@ -109,5 +115,88 @@ class LogMelFeatureExtractionV1(nn.Module):
             length = (length // self.hop_length) + 1
         else:
             length = ((length - self.n_fft) // self.hop_length) + 1
+
+        return feature_data, length.int()
+
+
+@dataclass
+class RasrCompatibleLogMelFeatureExtractionV1Config(ModelConfiguration):
+    """
+    Attributes:
+        sample_rate: audio sample rate in Hz
+        win_size: window size in seconds
+        hop_size: window shift in seconds
+        f_min: minimum filter frequency in Hz
+        f_max: maximum filter frequency in Hz
+        min_amp: minimum amplitude for safe log
+        num_filters: number of mel windows
+    """
+
+    sample_rate: int
+    win_size: float
+    hop_size: float
+    f_min: int
+    f_max: int
+    min_amp: float
+    num_filters: int
+
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        assert self.f_max <= self.sample_rate // 2, "f_max can not be larger than half of the sample rate"
+        assert self.f_min >= 0 and self.f_max > 0 and self.sample_rate > 0, "frequencies need to be positive"
+        assert self.win_size > 0 and self.hop_size > 0, "window settings need to be positive"
+        assert self.num_filters > 0, "number of filters needs to be positive"
+        assert self.hop_size <= self.win_size, "using a larger hop size than window size does not make sense"
+
+
+class RasrCompatibleLogMelFeatureExtractionV1(nn.Module):
+    """
+    Rasr-compatible log-mel feature extraction using log10. Does not use torchaudio.
+
+    Using it wrapped with torch.no_grad() is recommended if no gradient is needed
+    """
+
+    def __init__(self, cfg: RasrCompatibleLogMelFeatureExtractionV1Config):
+        super().__init__()
+        self.hop_length = int(cfg.hop_size * cfg.sample_rate)
+        self.min_amp = cfg.min_amp
+        self.win_length = int(cfg.win_size * cfg.sample_rate)
+        # smallest power if two which is greater than or equal to win_length
+        self.n_fft = 2 ** math.ceil(math.log2(self.win_length))
+
+        self.register_buffer(
+            "mel_basis",
+            torch.tensor(
+                filters.mel(
+                    sr=cfg.sample_rate,
+                    n_fft=cfg.n_fft,
+                    n_mels=cfg.num_filters,
+                    fmin=cfg.f_min,
+                    fmax=cfg.f_max,
+                    htk=True,
+                    norm=None,
+                ),
+            ),
+        )
+        self.register_buffer("window", torch.hann_window(self.win_length, periodic=False))
+
+    def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        :param raw_audio: [B, T]
+        :param length in samples: [B]
+        :return features as [B,T,F] and length in frames [B]
+        """
+        windowed = raw_audio.unfold(1, size=self.win_length, step=self.hop_length)  # [B, T', W=win_length]
+        smoothed = windowed * self.window.unsqueeze(0)  # [B, T', W]
+
+        # Compute power spectrum using torch.fft.rfftn
+        power_spectrum = torch.abs(torch.fft.rfftn(smoothed, s=self.n_fft)) ** 2  # [B, T', F=n_fft//2+1]
+        power_spectrum = power_spectrum.transpose(1, 2)  # [B, F, T']
+
+        melspec = torch.einsum("...ft,mf->...mt", power_spectrum, self.mel_basis)  # [B, F'=num_filters, T']
+        log_melspec = torch.log10(torch.clamp(melspec, min=self.min_amp))
+        feature_data = torch.transpose(log_melspec, 1, 2)  # [B, T', F']
+
+        length = ((length - self.win_length) // self.hop_length) + 1
 
         return feature_data, length.int()
--- a/i6_models/primitives/feature_extraction.py
+++ b/i6_models/primitives/feature_extraction.py
@@ -181,22 +181,33 @@ class RasrCompatibleLogMelFeatureExtractionV1(nn.Module):
     def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
         """
         :param raw_audio: [B, T]
-        :param length in samples: [B]
+        :param length: in samples [B]
         :return features as [B,T,F] and length in frames [B]
         """
+        assert raw_audio.shape[1] > 0  # also same for length
+        res_size = max(raw_audio.shape[1] - self.win_length + self.hop_length - 1, 0) // self.hop_length + 1
+        res_length = (
+            torch.maximum(length - self.win_length + self.hop_length - 1, torch.zeros_like(length)) // self.hop_length
+            + 1
+        )
+
         # preemphasize
         preemphasized = raw_audio.clone()
         preemphasized[..., 1:] -= self.alpha * preemphasized[..., :-1]
         preemphasized[..., 0] = 0.0
 
         # zero pad for the last frame
-        padded = torch.cat(
-            [preemphasized, torch.zeros(preemphasized.shape[0], (self.hop_length - 1), device=preemphasized.device)],
-            dim=1,
-        )
+        last_win_size = preemphasized.shape[1] - (res_size - 1) * self.hop_length
+        last_pad = self.win_length - last_win_size
+        padded = torch.nn.functional.pad(preemphasized, (0, last_pad))
 
         windowed = padded.unfold(1, size=self.win_length, step=self.hop_length)  # [B, T', W=win_length]
-        smoothed = windowed * self.window[None, None, :]  # [B, T', W]
+        smoothed = windowed[:, :-1] * self.window[None, None, :]  # [B, T'-1, W]
+
+        # The last window might be shorter. Will use a shorter Hanning window then. Need to fix that.
+        last_win = torch.hann_window(last_win_size, periodic=False, dtype=torch.float64).to(torch.float32)
+        last_win = torch.nn.functional.pad(last_win, (0, last_pad))  # [W]
+        smoothed = torch.cat([smoothed, (windowed[:, -1] * last_win[None, :])[:, None]], dim=1)  # [B, T', W]
 
         # compute amplitude spectrum using torch.fft.rfftn
         amplitude_spectrum = torch.abs(torch.fft.rfftn(smoothed, s=self.n_fft))  # [B, T', F=n_fft//2+1]
@@ -206,6 +217,4 @@ class RasrCompatibleLogMelFeatureExtractionV1(nn.Module):
         log_melspec = torch.log10(melspec + self.min_amp)
         feature_data = torch.transpose(log_melspec, 1, 2)  # [B, T', F']
 
-        length = torch.ceil((length - self.win_length) / self.hop_length) + 1
-
-        return feature_data, length.int()
+        return feature_data, res_length
--- a/i6_models/primitives/feature_extraction.py
+++ b/i6_models/primitives/feature_extraction.py
@@ -205,7 +205,9 @@ class RasrCompatibleLogMelFeatureExtractionV1(nn.Module):
         smoothed = windowed[:, :-1] * self.window[None, None, :]  # [B, T'-1, W]
 
         # The last window might be shorter. Will use a shorter Hanning window then. Need to fix that.
-        last_win = torch.hann_window(last_win_size, periodic=False, dtype=torch.float64).to(torch.float32)
+        last_win = torch.hann_window(last_win_size, periodic=False, dtype=torch.float64).to(
+            self.window.device, torch.float32
+        )
         last_win = torch.nn.functional.pad(last_win, (0, last_pad))  # [W]
         smoothed = torch.cat([smoothed, (windowed[:, -1] * last_win[None, :])[:, None]], dim=1)  # [B, T', W]
 
--- a/i6_models/primitives/feature_extraction.py
+++ b/i6_models/primitives/feature_extraction.py
@@ -146,8 +146,6 @@ class RasrCompatibleLogMelFeatureExtractionV1Config(ModelConfiguration):
 class RasrCompatibleLogMelFeatureExtractionV1(nn.Module):
     """
     Rasr-compatible log-mel feature extraction using log10. Does not use torchaudio.
-
-    Using it wrapped with torch.no_grad() is recommended if no gradient is needed
     """
 
     def __init__(self, cfg: RasrCompatibleLogMelFeatureExtractionV1Config):
@@ -180,17 +178,24 @@ class RasrCompatibleLogMelFeatureExtractionV1(nn.Module):
         :param length in samples: [B]
         :return features as [B,T,F] and length in frames [B]
         """
-        windowed = raw_audio.unfold(1, size=self.win_length, step=self.hop_length)  # [B, T', W=win_length]
+        # preemphasis
+        preemphasized = raw_audio.clone()
+        preemphasized[..., 1:] -= 1.0 * preemphasized[..., :-1]
+
+        # zero pad for the last frame
+        padded = torch.cat([preemphasized, torch.zeros(preemphasized.shape[0], (self.hop_length - 1))], dim=1)
+
+        windowed = padded.unfold(1, size=self.win_length, step=self.hop_length)  # [B, T', W=win_length]
         smoothed = windowed * self.window.unsqueeze(0)  # [B, T', W]
 
-        # Compute power spectrum using torch.fft.rfftn
-        power_spectrum = torch.abs(torch.fft.rfftn(smoothed, s=self.n_fft)) ** 2  # [B, T', F=n_fft//2+1]
-        power_spectrum = power_spectrum.transpose(1, 2)  # [B, F, T']
+        # Compute amplitude spectrum using torch.fft.rfftn
+        amplitude_spectrum = torch.abs(torch.fft.rfftn(smoothed, s=self.n_fft))  # [B, T', F=n_fft//2+1]
+        amplitude_spectrum = amplitude_spectrum.transpose(1, 2)  # [B, F, T']
 
-        melspec = torch.einsum("...ft,mf->...mt", power_spectrum, self.mel_basis)  # [B, F'=num_filters, T']
-        log_melspec = torch.log10(torch.clamp(melspec, min=self.min_amp))
+        melspec = torch.einsum("...ft,mf->...mt", amplitude_spectrum, self.mel_basis)  # [B, F'=num_filters, T']
+        log_melspec = torch.log10(melspec + self.min_amp)
         feature_data = torch.transpose(log_melspec, 1, 2)  # [B, T', F']
 
-        length = ((length - self.win_length) // self.hop_length) + 1
+        length = math.ceil((length - self.win_length) / self.hop_length) + 1
 
         return feature_data, length.int()
--- a/i6_models/primitives/feature_extraction.py
+++ b/i6_models/primitives/feature_extraction.py
@@ -152,6 +152,8 @@ class RasrCompatibleLogMelFeatureExtractionV1(nn.Module):
 
     def __init__(self, cfg: RasrCompatibleLogMelFeatureExtractionV1Config):
         super().__init__()
+
+        self.sample_rate = int(cfg.sample_rate)
         self.hop_length = int(cfg.hop_size * cfg.sample_rate)
         self.min_amp = cfg.min_amp
         self.win_length = int(cfg.win_size * cfg.sample_rate)
@@ -211,8 +213,9 @@ class RasrCompatibleLogMelFeatureExtractionV1(nn.Module):
         last_win = torch.nn.functional.pad(last_win, (0, last_pad))  # [W]
         smoothed = torch.cat([smoothed, (windowed[:, -1] * last_win[None, :])[:, None]], dim=1)  # [B, T', W]
 
-        # compute amplitude spectrum using torch.fft.rfftn
-        amplitude_spectrum = torch.abs(torch.fft.rfftn(smoothed, s=self.n_fft))  # [B, T', F=n_fft//2+1]
+        # compute amplitude spectrum using torch.fft.rfftn with Rasr specific scaling
+        fft = torch.fft.rfftn(smoothed, s=self.n_fft) / self.sample_rate # [B, T', F=n_fft//2+1]
+        amplitude_spectrum = torch.abs(fft)  # [B, T', F=n_fft//2+1]
 
         melspec = torch.einsum("...tf,mf->...tm", amplitude_spectrum, self.mel_basis)  # [B, T', F'=num_filters]
         log_melspec = torch.log10(melspec + self.min_amp)
--- a/i6_models/primitives/feature_extraction.py
+++ b/i6_models/primitives/feature_extraction.py
@@ -213,10 +213,8 @@ class RasrCompatibleLogMelFeatureExtractionV1(nn.Module):
 
         # compute amplitude spectrum using torch.fft.rfftn
         amplitude_spectrum = torch.abs(torch.fft.rfftn(smoothed, s=self.n_fft))  # [B, T', F=n_fft//2+1]
-        amplitude_spectrum = amplitude_spectrum.transpose(1, 2)  # [B, F, T']
 
-        melspec = torch.einsum("...ft,mf->...mt", amplitude_spectrum, self.mel_basis)  # [B, F'=num_filters, T']
+        melspec = torch.einsum("...tf,mf->...tm", amplitude_spectrum, self.mel_basis)  # [B, T', F'=num_filters]
         log_melspec = torch.log10(melspec + self.min_amp)
-        feature_data = torch.transpose(log_melspec, 1, 2)  # [B, T', F']
 
-        return feature_data, res_length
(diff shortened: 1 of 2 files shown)


i6_models/primitives/feature_extraction.py
```diff
@@ -111,3 +117,80 @@ def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
...
+        self.register_buffer("window", torch.hann_window(self.win_length, periodic=False))
+
+    def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        :param raw_audio: [B, T]
+        :param length in samples: [B]
+        :return features as [B,T,F] and length in frames [B]
+        """
+        windowed = raw_audio.unfold(1, size=self.win_length, step=self.hop_length)  # [B, T', W=win_length]
+        smoothed = windowed * self.window.unsqueeze(0)  # [B, T', W]
```
michelwi, line 184: There is preemphasis missing compared to the default rasr feature extraction. should we add it?

michelwi (commented):
Just one comment when comparing to what I know from usual rasr feature flows. 

Otherwise looks good to me

i6_models/primitives/feature_extraction.py
kuacakuaca: Rasr again [rasr](https://pytorch.org/audio/stable/generated/torchaudio.functional.preemphasis.html) does this differently than for example [torchaudio](https://pytorch.org/audio/stable/generated/torchaudio.functional.preemphasis.html), as Rasr keeps the last state from last segment(?).

i6_models/primitives/feature_extraction.py
```diff
@@ -111,3 +117,80 @@ def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
...
+        assert self.win_size > 0 and self.hop_size > 0, "window settings need to be positive"
+        assert self.num_filters > 0, "number of filters needs to be positive"
+        assert self.hop_size <= self.win_size, "using a larger hop size than window size does not make sense"
+
+
+class RasrCompatibleLogMelFeatureExtractionV1(nn.Module):
+    """
+    Rasr-compatible log-mel feature extraction using log10. Does not use torchaudio.
+
+    Using it wrapped with torch.no_grad() is recommended if no gradient is needed
```
albertz, line 150: I don't really understand this comment on `no_grad`. Isn't this just true in general for any PyTorch code? Why mention this explicitly here? Or is this module somehow special w.r.t. this?

i6_models/primitives/feature_extraction.py
albertz: > Rasr keeps the last state from last segment(?).

So RASR is stateful? This function here is stateless, and I think that makes more sense. I would not add this. Not even sure how you would add this, as you don't know the last segment here in any case.

albertz (approved):
I cannot really judge just from the code whether it is correct. I guess you tested it, that you get the same output? So I would say, then all is fine.

If you want, maybe a test could be added, where you feed some dummy data through RASR, and then get the output, and compare it with this module output. You can store the RASR output directly in the test case, so no need to run RASR in the test case. But not really necessary now.

i6_models/primitives/feature_extraction.py
curufinwe: I guess this is only a problem in streaming. But this feature extraction is mostly for training where we see the full utterance and thus don't need to worry about state-carryover. So it would be nice to add it (as an option).

kuacakuaca: maybe we should change the name of the class, maybe to `RasrReproducibleLogMelFeatureExtractionV1Config`? @michelwi @curufinwe

curufinwe: > > I guess this is only a problem in streaming. But this feature extraction is mostly for training where we see the full utterance and thus don't need to worry about state-carryover. So it would be nice to add it (as an option).
> 
> but if we don't do preemphais in training, we would also not do in recognition in RASR for the compatibility.

Correct.

> maybe we should change the name of the class, maybe to `RasrReproducibleLogMelFeatureExtractionV1Config`? @michelwi @curufinwe

If you want, I have no preference.

i6_models/primitives/feature_extraction.py
```diff
@@ -111,3 +117,80 @@ def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
...
+        )
+        self.register_buffer("window", torch.hann_window(self.win_length, periodic=False))
+
+    def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        :param raw_audio: [B, T]
+        :param length in samples: [B]
+        :return features as [B,T,F] and length in frames [B]
+        """
+        windowed = raw_audio.unfold(1, size=self.win_length, step=self.hop_length)  # [B, T', W=win_length]
```
michelwi, line 183: `torch.Tensor.unfold` will throw away any excess samples that do not form a full frame. (round down)
rasr on the other hand will do zero padding and emit an additional frame. (round up)

Will this be a problem for "rasr compatibility"?
cc @DanEnergetics

kuacakuaca: will rasr do zero-padding anyway, even there is one sample left?

i6_models/primitives/feature_extraction.py
```diff
@@ -180,17 +178,24 @@ def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
         :param length in samples: [B]
         :return features as [B,T,F] and length in frames [B]
         """
-        windowed = raw_audio.unfold(1, size=self.win_length, step=self.hop_length)  # [B, T', W=win_length]
+        # preemphasis
+        preemphasized = raw_audio.clone()
+        preemphasized[..., 1:] -= 1.0 * preemphasized[..., :-1]
```
michelwi, line 183: make the factor `1.0` a parameter?

i6_models/primitives/feature_extraction.py
```diff
@@ -153,8 +155,10 @@ def __init__(self, cfg: RasrCompatibleLogMelFeatureExtractionV1Config):
         self.hop_length = int(cfg.hop_size * cfg.sample_rate)
         self.min_amp = cfg.min_amp
         self.win_length = int(cfg.win_size * cfg.sample_rate)
-        # smallest power if two which is greater than or equal to win_length
-        self.n_fft = 2 ** math.ceil(math.log2(self.win_length))
+        self.n_fft = 2 ** math.ceil(
+            math.log2(self.win_length)
+        )  # smallest power if two which is greater than or equal to win_length
+        self.alpha = cfg.alpha
```
michelwi, line 161: I liked the old one better, but ok^^

michelwi (approved):

michelwi:
> one frame rasr features: 0.234030 0.731747 1.082119 
> same frame torch features: 0.2297, 0.7334, 1.0803,

the difference is on the 3rd decimal position, that is more than just numeric precision so there is still something differen?

i6_models/primitives/feature_extraction.py
```diff
@@ -111,3 +117,89 @@ def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
...
+
+        # compute amplitude spectrum using torch.fft.rfftn
+        amplitude_spectrum = torch.abs(torch.fft.rfftn(smoothed, s=self.n_fft))  # [B, T', F=n_fft//2+1]
+        amplitude_spectrum = amplitude_spectrum.transpose(1, 2)  # [B, F, T']
+
+        melspec = torch.einsum("...ft,mf->...mt", amplitude_spectrum, self.mel_basis)  # [B, F'=num_filters, T']
+        log_melspec = torch.log10(melspec + self.min_amp)
+        feature_data = torch.transpose(log_melspec, 1, 2)  # [B, T', F']
+
+        length = math.ceil((length - self.win_length) / self.hop_length) + 1
```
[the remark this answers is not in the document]
kuacakuaca: I'll fix, thanks for testing!

kuacakuaca:
> > one frame rasr features: 0.234030 0.731747 1.082119
> > same frame torch features: 0.2297, 0.7334, 1.0803,
> 
> the difference is on the 3rd decimal position, that is more than just numeric precision so there is still something different?

not sure if it is worth investigating further. i don't think our tensorflow features and rasr features match 100%.

i6_models/primitives/feature_extraction.py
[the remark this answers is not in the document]
michelwi: I had successfully tested commit `5fa4bf3abba33659cc8731c2d0e5b2702486d787` before.
Now I am testing the current version and got the changes I complained about. 
One additional problem I got was that in one instance the batch size was off by one. I am still investigating how that happened.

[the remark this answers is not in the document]
kuacakuaca: > I had successfully tested commit `5fa4bf3abba33659cc8731c2d0e5b2702486d787` before. Now I am testing the current version and got the changes I complained about. One additional problem I got was that in one instance the batch size was off by one. I am still investigating how that happened.

why in one instance? you observed was that the batch dimension size changed before and after feature extraction sub-network? and was the batch dimension size 1 larger or 1 smaller?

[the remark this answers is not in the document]
kuacakuaca: I tried to look into RASR, but can't really comprehend how it's done there, the relevant code could be in https://bitbucket.org/omnifluent/asr-rwth-asr/src/b9df49d2dcd26488259bab399637d102a538bfec/src/Signal/WindowBuffer.cc?at=master#lines-99.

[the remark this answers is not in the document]
kuacakuaca: do these 300 segments have something common in number of samples (like always have xxx21 samples)? I remember for Korean training I either had to use this magical `min_chunk_size` or turn on filtering to get rid of that mismatch error.

i6_models/primitives/feature_extraction.py
michelwi: FYI: Adding preemphasis improved the WER from 15.9 to 15.6 one some random model training I did to test the pipeline.

michelwi (approved):

i6_models/primitives/feature_extraction.py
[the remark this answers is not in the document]
michelwi: I don't see a clear pattern. most have `num_samples % 20 == 0` but then there are also some with xxx01.

albertz:
> one frame rasr features
> same frame torch features

I'm checking this currently. What was the RASR flow file to get these? And what were the corresponding settings for `RasrCompatibleLogMelFeatureExtractionV1Config`? So far, I always get very different results.

michelwi:
> I'm checking this currently. What was the RASR flow file to get these? And what were the corresponding settings for `RasrCompatibleLogMelFeatureExtractionV1Config`? So far, I always get very different results (off by some shift of ~4.515).

I cannot tell you what was used to create the numbers above and I think those might be from a version before preemphasis and zero padding was added. Here are the config/flow that I have used recently:
```
feat_cfg = RasrCompatibleLogMelFeatureExtractionV1Config(
            sample_rate=sampling_rate,
            win_size=0.025,
            hop_size=0.01,
            min_amp=1e-6,
            num_filters=num_channels,
)
```

```
<?xml version="1.0" ?>
<network name="network">
  <out name="features"/>
  <param name="end-time"/>
  <param name="input-file"/>
  <param name="start-time"/>
  <param name="track"/>
  <node end-time="$(end-time)" file="$(input-file).wav" filter="audio-input-file-wav" name="samples" start-time="$(start-time)"/>
  <node filter="generic-vector-s16-demultiplex" name="demultiplex" track="$(track)"/>
  <link from="samples" to="demultiplex"/>
  <node filter="generic-convert-vector-s16-to-vector-f32" name="convert"/>
  <link from="demultiplex" to="convert"/>
  <node alpha="1.0" filter="signal-preemphasis" name="preemphasis"/>
  <link from="convert" to="preemphasis"/>
  <node filter="signal-window" length="0.025" name="window" shift="0.01" type="hanning"/>
  <link from="preemphasis" to="window"/>
  <node filter="signal-real-fast-fourier-transform" maximum-input-size="0.025" name="fft"/>
  <link from="window" to="fft"/>
  <node filter="generic-vector-f32-multiplication" name="scaling" value="16000"/>
  <link from="fft" to="scaling"/>
  <node filter="signal-vector-alternating-complex-f32-amplitude" name="amplitude-spectrum"/>
  <link from="scaling" to="amplitude-spectrum"/>
  <node filter="signal-filterbank" filter-width="70.12402584464985" name="filterbank" warp-differential-unit="false" warping-function="mel"/>
  <link from="amplitude-spectrum" to="filterbank"/>
  <node filter="generic-vector-f32-log-plus" name="nonlinear" value="1.175494e-38"/>
  <link from="filterbank" to="nonlinear"/>
  <link from="nonlinear" to="network:features"/>
</network>
```

Just by copy pasting I see that the epsilon defined in config and flow does not match.. 🙈

albertz:
I still wonder about the shift (off by some shift of ~4.515): So, I figured out now, this means, before the log10, this is exactly a factor 2^15 = 32,768. But I think I used the same flow as you. (Well, actually, I did not, @curufinwe gave me some generated feature cache, and he said it was generated with this flow network.) So, I'm not sure, maybe this here is actually the wrong flow network?

I also still don't exactly understand what we mean by "rasr compatible feature extraction" here. So this means it supposed to do the same as this specific flow network? Then shouldn't this reference flow network be documented somehow? Otherwise, "rasr compatible feature extraction" can mean anything or nothing?

michelwi:
> I still wonder about the shift (off by some shift of ~4.515): So, I figured out now, this means, before the log10, this is exactly a factor 2^15 = 32,768.

Rasr reads the wav samples as a short and then just casts them to float. If the other code reads the samples as float `\in [-1, 1]` directly, this would explain an offset of 2^15.

> But I think I used the same flow as you. (Well, actually, I did not, @curufinwe gave me some generated feature cache, and he said it was generated with this flow network.) So, I'm not sure, maybe this here is actually the wrong flow network?

As I said, this is the flow that I used, it is based on existing flow implementations but I changed it to be more similar to this PR. It could be that @curufinwe by accident (or virtue of obvious implementation) arrived at the same flow as me, but I would not guarantee it.

> I also still don't exactly understand what we mean by "rasr compatible feature extraction" here. So this means it supposed to do the same as this specific flow network? Then shouldn't this reference flow network be documented somehow? Otherwise, "rasr compatible feature extraction" can mean anything or nothing?

I agree to this question. I think as a first goal "rasr compatible" would mean something that can be exactly reproduced as a rasr flow network, although not necessarily the same as "the one flow that we always use".

But then, when I tested the version of this PR without preemphasis (and also without preemphasis in the rasr flow) to train a model, and compared to the current version of this PR (and the flow I posted above), I found an improvement in WER from 16.1 -> 15.7

So maybe as a stretch goal we should aim to make the implementation as similar as "the one flow we always use" (which is different from the one I posted above) as possible.

But then nobody would argue against using different features if the final WER is the same (or better) and the structure is easier.

albertz:
> Rasr reads the wav samples as a short and then just casts them to float. If the other code reads the samples as float `\in [-1, 1]` directly, this would explain an offset of 2^15.

Yes. [-1,1] is what you normally would get. But that means, the flow you posted here, also the flow @curufinwe gave me and used to generate the feature cache, is wrong. We should not have this factor 2^15 in there. The values should be within [-1,1].

The current version of the code here is not compatible to the flow network you posted, as either the flow network need to divide by 2^15, or the code here must multiply by 2^15. I would say the only case which make sense it to divide by 2^15 on RASR side, so use a different flow.

But if this incorrect flow is "the one flow we always use", and the goal is really to get that, then we should also have the factor 2^15 here. I wonder a bit why @kuacakuaca did not have this issue? Without this factor, she cannot possibly get similar features.

> ... as similar as ... nobody would argue against using different features ...

As I understand @curufinwe, the goal was actually to have 100% exactly the same features (up to numerical differences, e.g. 1e-7 or whatever), to be able to use pure PyTorch for training, and in recognition use RASR feature extraction instead, which makes streaming easier or so (I didn't fully understand what the actual issue is, why we can't just always use pure PyTorch...).

michelwi:
> I wonder a bit why @kuacakuaca did not have this issue?

for training we currently use hdf files that were generated with the [BlissToPcmHDFJob](https://github.com/rwth-i6/i6_core/blob/add09a8b640a2ba5928b815fa65f7504242be038/returnn/hdf.py#L207) using int16 dtype. This makes the factor 2^15 compatible in training and recognition using rasr.

For the problem at hand I would not worry about this factor of 2^15 at this stage. If we decide to use [-1,1] samples, we can adapt the flow in rasr to accommodate that. e.g. insert this `samples_scaling` node
```
  <node filter="generic-convert-vector-s16-to-vector-f32" name="convert"/>
  <link from="demultiplex" to="convert"/>
  <node filter="generic-vector-f32-multiplication" name="samples_scaling" value="0.00003051757"/>
  <link from="convert" to="samples_scaling"/>
  <node alpha="1.0" filter="signal-preemphasis" name="preemphasis"/>
  <link from="samples_scaling" to="preemphasis"/>
```

> > ... as similar as ... nobody would argue against using different features ...
>
>As I understand @curufinwe, the goal was actually to have 100% exactly the same features (up to numerical differences, e.g. 1e-7 or whatever),

exactly. as long as we can get the same features (up to numerical precision) in training and in rasr, we can use it. E.g. scaling the waveform samples would be possible. But that is easy on either side (rasr or pytorch).

If this constant offset goes away, are the features otherwise identically?

albertz:
> for training we currently use hdf files that were generated with the [BlissToPcmHDFJob](https://github.com/rwth-i6/i6_core/blob/add09a8b640a2ba5928b815fa65f7504242be038/returnn/hdf.py#L207) using int16 dtype. This makes the factor 2^15 compatible in training and recognition using rasr.

Oh right. I don't think it was a good idea to have this default. Using such strange range will lead to all kinds of unexpected behavior. All existing code will expect [-1,1] values, and if you use any existing feature extraction code, it will be quite off.

For example, using the code here on some Librispeech audio:
* i6_models `RasrCompatibleLogMelFeatureExtractionV1` with audio samples in [-1,1]: tensor[1407, 80] n=112560 x∈[-3.831, 1.524] μ=-1.372 σ=0.762
* i6_models, `RasrCompatibleLogMelFeatureExtractionV1` with audio samples in [-2^15,2^15-1]:  tensor[1407, 80] n=112560 x∈[0.685, 6.039] μ=3.143 σ=0.762
* RASR: tensor[1407, 80] n=112560 x∈[0.686, 6.039] μ=3.143 σ=0.762

I think the features in the first case with samples in [-1,1] behave much better.

But anyway. Even if I use audio samples in [-2^15,2^15-1] with i6_models, although the values look close from the output above, there is still some difference:
Mismatched elements: 108058 / 112560 (96.0%)
Greatest absolute difference: 0.005605936050415039 at index (297, 9) (up to 1e-05 allowed)
Greatest relative difference: 0.005547972578959174 at index (720, 4) (up to 1e-05 allowed)

michelwi:
Ok, thanks for the update. I think there are two orthogonal questions open

1. decide on `[-1, 1]` (compatible with other toolkits) vs `[-2^15, 2^15]` (compatible with default rasr settings)
2. find the cause of the remaining difference

> Greatest absolute difference: 0.005605936050415039 at index (297, 9) (up to 1e-05 allowed)
Greatest relative difference: 0.005547972578959174 at index (720, 4) (up to 1e-05 allowed)

albertz:
> 2\. find the cause of the remaining difference

I'm on this now. For this, I created a test case where I can test the flow file step by step (see `test_rasr_compatible` and `test_rasr_compatible_*`).

I found many more differences:

* Preemphasize was wrong in the first frame, should be 0. (https://github.com/rwth-i6/i6_models/pull/44/commits/f41a60c9ccfa26422da29f91769fd9083efe2c41)
* The Hanning window was slightly off. It was created in RASR with float64 and then afterwards converted to float32. This makes a difference for `cos`. I do the same now. (https://github.com/rwth-i6/i6_models/pull/44/commits/44aba814698a2251e0f614538c7dc3a19c496c0e) (Note, this caused a difference more than 1e-05.)
* The Hanning window in the last output frame was wrong. RASR applies a shorter Hanning window for the last output frame, just of the size of the remaining frames. (https://github.com/rwth-i6/i6_models/pull/44/commits/d84565047266048141c0aa20103a21236c6e7a4a)

Now, up to that point in the flow network (`window` node), the outputs match almost perfectly (at least 1e-30 absolute and relative).

And there are still remaining differences:

The FFT seems different. I need to understand this better. But looking at `amplitude-spectrum`, there is again some difference. Via `test_rasr_compatible_fft`:
```
>       torch.testing.assert_close(amplitude_spectrum, rasr_feat, rtol=1e-30, atol=1e-30)
E       AssertionError: Tensor-likes are not close!
E       
E       Mismatched elements: 126540 / 128243 (98.7%)
E       Greatest absolute difference: 0.01171875 at index (245, 5) (up to 1e-30 allowed)
E       Greatest relative difference: 0.028590014204382896 at index (206, 99) (up to 1e-30 allowed)
```

albertz:
I replicated the RASR C++ FFT code in Python. Now I still get some difference in the amplitude-spectrum:
```
>       torch.testing.assert_close(amplitude_spectrum, rasr_feat, rtol=1e-30, atol=1e-30)
E       AssertionError: Tensor-likes are not close!
E       
E       Mismatched elements: 72884 / 128243 (56.8%)
E       Greatest absolute difference: 0.0078125 at index (5, 6) (up to 1e-30 allowed)
E       Greatest relative difference: 2.3795448100827343e-07 at index (53, 75) (up to 1e-30 allowed)
```
I'm not sure if this is because I still (by mistake) do sth a little bit different than the corresponding C++ code, or because some of the math just behaves different in PyTorch (e.g. `torch.sin`, or maybe the order of multiplications, or so). I think RASR was also compiled using `fastmath`, so it could have reordered the formulas, also resulting in differences.

I now spent already quite a bit of time on this, and I'm a bit questioning whether this is really worth the effort? I also have the feeling, to get really close results, we probably need to have a lot of very custom implementations, our own custom FFT, etc. But on the other side, if we don't do this, it looks like we will still have quite large differences, too large that we can just use one as a drop-in for the other...

albertz:
Note, as a next step, to verify it even more whether the FFT produces exactly the same results or not, I would probably copy the RASR C++ code into some standalone test tool, and then verify step by step where it differs from the Python output. Maybe I still find some things I can fix on the Python side to make it really the same. Or if not, that will at least give more insights where it becomes different, where errors are accumulating.

michelwi:
> we probably need to have a lot of very custom implementations, our own custom FFT, etc. 

if you think it would be easier to change/augment the rasr implementation, we could try to go that route instead.

albertz:
> if you think it would be easier to change/augment the rasr implementation, we could try to go that route instead.

I'm really not an expert when it comes to FFT implementations. I know there are a couple of different ways to implement a Fourier transformation, but even when looking at the same algorithm, there are so much details which might result in such differences.

First, we should maybe better understand whether there is still some bug, or some difference we can fix. I don't know if there is. I also don't know if it is expected that the numerical differences of FFT implementations can be so large (at least it was unexpected to me, but I don't know enough about it).

On changing RASR: I'm not sure if this is really simpler. Maybe. For that, we should better understand what the Torch FFT is actually doing. I also wonder how different Torch FFT is on CPU vs GPU. E.g. on GPU, this uses CuFFT, so again a very specific implementation, and also closed source (I think), so we cannot even really check the implementation.

michelwi:
> I replicated the RASR C++ FFT code in Python. Now I still get some difference in the amplitude-spectrum:
> ```
> Greatest absolute difference: 0.0078125 at index (5, 6) (up to 1e-30 allowed)
> Greatest relative difference: 2.3795448100827343e-07 at index (53, 75) (up to 1e-30 allowed)
> ```


I ran your test cases and found that with `rtol=1e-6, atol=1e-30` there are no differences found. Given that the implementation uses float32 I would say the deviation by `2e-7` relative is acceptable.
If now applying the filterbanks also leaves the deviation below 1e-6 relative, then I would say we are done.

Is your implementation of `my_fft` "sufficiently fast" in practice so that we can use it in the `RasrCompatibleLogMelFeatureExtraction`?

albertz:
> I ran your test cases and found that with `rtol=1e-6, atol=1e-30` there are no differences found. Given that the implementation uses float32 I would say the deviation by `2e-7` relative is acceptable.

What test case exactly? But as you see from my output, the absolute error was quite high, so I would need atol=1e-2 or so that it passes.

> If now applying the filterbanks also leaves the deviation below 1e-6 relative, then I would say we are done.

But the output is already the best I could get, including `my_fft`. (Right? I don't remember exactly.)

> Is your implementation of `my_fft` "sufficiently fast" in practice so that we can use it in the `RasrCompatibleLogMelFeatureExtraction`?

I did not measure it, but I would expect it is way too slow. There are several nestings of Python loops. Already the C++ code was a bit too slow when done on-the-fly, and I expect this Python code will be factor 10 or 100 times slower. I don't think this is an option.

michelwi:
> What test case exactly?

`test_rasr_compatible_fft` and `test_rasr_compatible_amplitude_spectrum` as implemented in this PR, where I changed `rtol=1e-6` and `atol=1e-30`

> the absolute error was quite high, so I would need atol=1e-2 or so that it passes.

The definition of the closeness criterion according to https://pytorch.org/docs/stable/testing.html#torch.testing.assert_allclose is
```
∣ actual − expected ∣ ≤ atol + rtol ⋅ ∣expected∣
```
so if the value at the specific position in `expected` is large, then also a large absolute difference can be covered by `rtol * value`. My guess is, that this is happening.

kuacakuaca:
running test case `test_rasr_compatible_amplitude_spectrum` using `torch.fft.rfftn` yields following mismatch: 
```
RASR: torch.Size([499, 257]) x∈[0.005153656471520662, 77318.9375] μ=425.7894592285156 σ=3692.595947265625 torch.float32
i6_models torch.Size([499, 257]) x∈[0.005153656005859375, 77318.9296875] μ=425.7894287109375 σ=3692.595703125 torch.float32

Mismatched elements: 60230 / 128243 (47.0%)
Greatest absolute difference: 0.00764918327331543 at index (248, 250) (up to 1e-05 allowed)
Greatest relative difference: 0.03186587807087847 at index (253, 28) (up to 1e-05 allowed)
```
are those still not close enough?

tests/test_feature_extraction.py
```diff
@@ -65,3 +73,811 @@ def test_logmel_length():
...
+    except ImportError:
+        raise unittest.SkipTest("i6_core not available")
+    try:
+        import soundfile
+    except ImportError:
+        raise unittest.SkipTest("soundfile not available")
+    rasr_feature_extractor_bin_path = (
+        "/work/tools22/asr/rasr/rasr_onnx_haswell_0623/arch/linux-x86_64-standard/"
+        "feature-extraction.linux-x86_64-standard"
+    )
```
michelwi, line 90: As these test depend on the i6_cluster, should we remove them before finally merging?

michelwi (approved):
I re-approve the changes. I would deem the remaining difference small enough to not influence the WER.

tests/test_feature_extraction.py
albertz: I would leave them for reference. They will just raise `SkipTest` when it does not find the library, so not cause the tests to break elsewhere.

i6_models/primitives/feature_extraction.py
```diff
@@ -111,3 +117,107 @@ def forward(self, raw_audio, length) -> Tuple[torch.Tensor, torch.Tensor]:
...
+        # zero pad for the last frame
+        last_win_size = preemphasized.shape[1] - (res_size - 1) * self.hop_length
+        last_pad = self.win_length - last_win_size
+        padded = torch.nn.functional.pad(preemphasized, (0, last_pad))
+
+        windowed = padded.unfold(1, size=self.win_length, step=self.hop_length)  # [B, T', W=win_length]
+        smoothed = windowed[:, :-1] * self.window[None, None, :]  # [B, T'-1, W]
+
+        # The last window might be shorter. Will use a shorter Hanning window then. Need to fix that.
+        last_win = torch.hann_window(last_win_size, periodic=False, dtype=torch.float64).to(
```
curufinwe, line 210: Is this correct if you have sequences of different lengths in the batch? Maybe it's necessary to do this as a loop over all sequences in the batch and modify some parts of the output with this?

curufinwe (approved):

PR #47, opened 2024-04-18
Add the data augmentation method mixup

Judyxujj:
- This implementation of mixup only mix up the audio inputs x, but not on the targets y. 
- The implementation is partly based on Albert's implementation in returnn front end (c.f. [https://github.com/rwth-i6/i6_experiments/blob/main/users/zeyer/returnn/models/rf_mixup.py]) and Zoltan's implementation at Apptek.

### DIFF
--- a/i6_models/primitives/mixup.py
+++ b/i6_models/primitives/mixup.py
@@ -26,7 +26,7 @@ class MixupConfig(ModelConfiguration):
     lambda_max: float = 0.3
 
 
-class FeatureBuffer:
+class FeatureBuffer(torch.nn.Module):
     """
     The FeatureBuffer saves the feature from previous timeframes
     after the buffer is full, the buffer will not be updated
--- a/i6_models/primitives/mixup.py
+++ b/i6_models/primitives/mixup.py
@@ -1,6 +1,5 @@
 __all__ = ["MixupConfig", "Mixup"]
 
-import random
 from dataclasses import dataclass
 
 import torch
@@ -14,8 +13,8 @@ class MixupConfig(ModelConfiguration):
     Attributes:
         feature_dim: feature dimension
         apply_prob: probability to apply mixup at all
-        lambda_min: minimum lambda value
-        lambda_max: maximum lambda value
+        lambda_min: minimum of lambda_ value (suppose input = input + lambda_ * mixup_values)
+        lambda_max: maximum of lambda_ value (suppose input = input + lambda_ * mixup_values)
         max_num_mix: maximum number of mixups (random int in [1, max_num_mix])
         buffer_size: number of frames.
     """
@@ -35,6 +34,7 @@ class FeatureBuffer(torch.nn.Module):
     """
 
     def __init__(self, *, buffer_size: int, feature_dim: int):
+        super().__init__()
         self.filled = False
         self.pos = 0
         self.buffer_size = buffer_size
@@ -93,6 +93,7 @@ class Mixup(torch.nn.Module):
     """
 
     def __init__(self, *, cfg: MixupConfig):
+        super().__init__()
         self.apply_prob = cfg.apply_prob
         self.lambda_min = cfg.lambda_min
         self.lambda_max = cfg.lambda_max
@@ -103,6 +104,11 @@ class Mixup(torch.nn.Module):
         self.feature_buffer = FeatureBuffer(buffer_size=self.buffer_size, feature_dim=self.feature_dim)
 
     def forward(self, input: torch.Tensor, /, sequence_mask: torch.Tensor) -> torch.Tensor:
+        """
+        :param input: [B, T, F]
+        :param sequence_mask: [B, T]
+        :return tensor as [B, T, F]
+        """
         assert (
             input.size()[-1] == self.feature_dim
         ), "the given feature dimension does not match input feature dimension"
--- a/i6_models/primitives/mixup.py
+++ b/i6_models/primitives/mixup.py
@@ -56,7 +56,7 @@ class FeatureBuffer:
             self.cache[:end_pos] = tensor[delta_pos:]
             self.pos = end_pos
 
-    def getRandom(self, b_dim: int, t_dim: int, max_num_mixup: int, n_mask: torch.tensor) -> torch.Tensor:
+    def get_random(self, b_dim: int, t_dim: int, max_num_mixup: int, n_mask: torch.tensor) -> torch.Tensor:
         if not self.filled and self.pos == 0:
             return None
         else:
--- a/i6_models/primitives/mixup.py
+++ b/i6_models/primitives/mixup.py
@@ -12,18 +12,20 @@ from i6_models.config import ModelConfiguration
 class MixupConfig(ModelConfiguration):
     """
     Attributes:
-        buffer_size: number of frames.
+        feature_dim: feature dimension
         apply_prob: probability to apply mixup at all
         lambda_min: minimum lambda value
         lambda_max: maximum lambda value
         max_num_mix: maximum number of mixups (random int in [1, max_num_mix])
+        buffer_size: number of frames.
     """
 
+    feature_dim: int
+    apply_prob: float
+    max_num_mix: int
+    lambda_min: float
+    lambda_max: float
     buffer_size: int = 1_000_000
-    apply_prob: float = 1.0
-    max_num_mix: int = 4
-    lambda_min: float = 0.003
-    lambda_max: float = 0.3
 
 
 class FeatureBuffer(torch.nn.Module):
@@ -90,15 +92,15 @@ class Mixup(torch.nn.Module):
     C.f. https://github.com/rwth-i6/i6_experiments/blob/main/users/zeyer/returnn/models/rf_mixup.py
     """
 
-    def __init__(self, *, feature_dim: int, cfg: MixupConfig):
+    def __init__(self, *, cfg: MixupConfig):
         self.apply_prob = cfg.apply_prob
         self.lambda_min = cfg.lambda_min
         self.lambda_max = cfg.lambda_max
         self.max_num_mix = cfg.max_num_mix
         self.buffer_size = cfg.buffer_size
-        self.feature_dim = feature_dim
+        self.feature_dim = cfg.feature_dim
 
-        self.feature_buffer = FeatureBuffer(self.buffer_size, feature_dim)
+        self.feature_buffer = FeatureBuffer(buffer_size=self.buffer_size, feature_dim=self.feature_dim)
 
     def forward(self, input: torch.Tensor, /, sequence_mask: torch.Tensor) -> torch.Tensor:
         assert (
--- a/i6_models/primitives/mixup.py
+++ b/i6_models/primitives/mixup.py
@@ -75,7 +75,7 @@ class FeatureBuffer(torch.nn.Module):
             )  # [B, M] (M denotes maximum of num_mixup over the batch)
             start_indicies_flat = torch.masked_select(
                 start_indicies, n_mask
-            )  # [B, M'] (M' denotes sum of num_mixup over the batch)
+            )  # [M'] (M' denotes sum of num_mixup over the batch)
 
             idx = torch.arange(t_dim)  # [T]
             idx = idx[:, None] + start_indicies_flat[None, :]  # [T, M']
--- a/i6_models/primitives/mixup.py
+++ b/i6_models/primitives/mixup.py
@@ -76,7 +76,7 @@ class FeatureBuffer(torch.nn.Module):
             return mixup_values
 
 
-class Mixup:
+class Mixup(torch.nn.Module):
     """
     Implement the Mixup data augmentation method
     C.f. https://arxiv.org/abs/1710.09412
--- a/i6_models/primitives/mixup.py
+++ b/i6_models/primitives/mixup.py
@@ -32,7 +32,7 @@ class FeatureBuffer:
     after the buffer is full, the buffer will not be updated
     """
 
-    def __init__(self, buffer_size: int, feature_dim: int):
+    def __init__(self, *, buffer_size: int, feature_dim: int):
         self.filled = False
         self.pos = 0
         self.buffer_size = buffer_size
--- a/i6_models/primitives/mixup.py
+++ b/i6_models/primitives/mixup.py
@@ -1,0 +1,151 @@
+__all__ = ["MixupConfig", "Mixup"]
+
+import random
+from dataclasses import dataclass
+
+import torch
+
+from i6_models.config import ModelConfiguration
+
+
+@dataclass
+class MixupConfig(ModelConfiguration):
+    """
+    Attributes:
+        buffer_size: number of frames.
+        apply_prob: probability to apply mixup at all
+        lambda_min: minimum lambda value
+        lambda_max: maximum lambda value
+        max_num_mix: maximum number of mixups (random int in [1, max_num_mix])
+    """
+
+    buffer_size: int = 1_000_000
+    apply_prob: float = 1.0
+    max_num_mix: int = 4
+    lambda_min: float = 0.003
+    lambda_max: float = 0.3
+
+
+class FeatureBuffer:
+    """
+    The FeatureBuffer saves the feature from previous timeframes
+    after the buffer is full, the buffer will not be updated
+    """
+
+    def __init__(self, buffer_size: int, feature_dim: int):
+        self.filled = False
+        self.pos = 0
+        self.buffer_size = buffer_size
+        self.cache = torch.nn.parameter.Parameter(
+            data=torch.zeros((self.buffer_size, feature_dim)), requires_grad=False
+        )
+
+    def append(self, tensor: torch.Tensor):
+        t_dim = tensor.shape[0]
+        if t_dim > self.buffer_size:
+            tensor = tensor[: self.buffer_size]
+            t_dim = self.buffer_size
+        delta_pos = min(self.buffer_size - self.pos, t_dim)
+        end_pos = self.pos + delta_pos
+        self.cache[self.pos : end_pos] = tensor[:delta_pos]
+        self.pos = end_pos
+
+        if end_pos == self.buffer_size:
+            self.filled = True
+            end_pos = t_dim - delta_pos
+            self.cache[:end_pos] = tensor[delta_pos:]
+            self.pos = end_pos
+
+    def getRandom(self, b_dim: int, t_dim: int, max_num_mixup: int, n_mask: torch.tensor) -> torch.Tensor:
+        if not self.filled and self.pos == 0:
+            return None
+        else:
+            end_idx = self.buffer_size if self.filled else self.pos
+            max_end_idx = end_idx - t_dim
+
+            start_indicies = torch.randint(
+                high=max_end_idx, size=(b_dim, max_num_mixup)
+            )  # [B, M] (M denotes maximum of num_mixup over the batch)
+            start_indicies_flat = torch.masked_select(
+                start_indicies, n_mask
+            )  # [B, M'] (M' denotes sum of num_mixup over the batch)
+
+            idx = torch.arange(t_dim)
+            idx = torch.unsqueeze(idx, dim=-1) + start_indicies_flat  # [T, M']
+            mixup_values = self.cache[idx]  # [T, M', F]
+            return mixup_values
+
+
+class Mixup:
+    """
+    Implement the Mixup data augmentation method
+    C.f. https://arxiv.org/abs/1710.09412
+
+    The code is partly based on Albert's implementation in returnn front-end
+    C.f. https://github.com/rwth-i6/i6_experiments/blob/main/users/zeyer/returnn/models/rf_mixup.py
+    """
+
+    def __init__(self, feature_dim: int, cfg: MixupConfig):
+        self.apply_prob = cfg.apply_prob
+        self.lambda_min = cfg.lambda_min
+        self.lambda_max = cfg.lambda_max
+        self.max_num_mix = cfg.max_num_mix
+        self.buffer_size = cfg.buffer_size
+        self.feature_dim = feature_dim
+
+        self.feature_buffer = FeatureBuffer(self.buffer_size, feature_dim)
+
+    def __call__(self, input: torch.Tensor, sequence_mask: torch.Tensor) -> torch.Tensor:
+        assert (
+            input.size()[-1] == self.feature_dim
+        ), "the given feature dimension does not match input feature dimension"
+        input_ = self._maybe_apply_mixup(input, sequence_mask)
+        self._append_to_buffer(input, sequence_mask)
+        return input_
+
+    def _append_to_buffer(self, input: torch.Tensor, sequence_mask: torch.Tensor):
+        _, _, f_dim = input.size()
+        # mask out the padded frames before append it to feature_buffer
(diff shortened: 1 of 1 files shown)


i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+        self.cache[self.pos : end_pos] = tensor[:delta_pos]
+        self.pos = end_pos
+
+        if end_pos == self.buffer_size:
+            self.filled = True
+            end_pos = t_dim - delta_pos
+            self.cache[:end_pos] = tensor[delta_pos:]
+            self.pos = end_pos
+
+    def getRandom(self, b_dim: int, t_dim: int, max_num_mixup: int, n_mask: torch.tensor) -> torch.Tensor:
```
albertz, line 59: You should follow PEP8 style.

```suggestion
    def get_random(self, b_dim: int, t_dim: int, max_num_mixup: int, n_mask: torch.tensor) -> torch.Tensor:
```

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+    lambda_max: float = 0.3
+
+
+class FeatureBuffer:
+    """
+    The FeatureBuffer saves the feature from previous timeframes
+    after the buffer is full, the buffer will not be updated
+    """
+
+    def __init__(self, buffer_size: int, feature_dim: int):
```
albertz, line 35:
```suggestion
    def __init__(self, *, buffer_size: int, feature_dim: int):
```

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+        self.filled = False
+        self.pos = 0
+        self.buffer_size = buffer_size
+        self.cache = torch.nn.parameter.Parameter(
+            data=torch.zeros((self.buffer_size, feature_dim)), requires_grad=False
+        )
+
+    def append(self, tensor: torch.Tensor):
```
albertz, line 43: Please document the expected shape of `tensor`.

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+class Mixup:
+    """
+    Implement the Mixup data augmentation method
+    C.f. https://arxiv.org/abs/1710.09412
+
+    The code is partly based on Albert's implementation in returnn front-end
+    C.f. https://github.com/rwth-i6/i6_experiments/blob/main/users/zeyer/returnn/models/rf_mixup.py
+    """
+
+    def __init__(self, feature_dim: int, cfg: MixupConfig):
```
albertz, line 88:
```suggestion
    def __init__(self, *, feature_dim: int, cfg: MixupConfig):
```

albertz, line 88: I wonder whether `feature_dim` should also be part of `MixupConfig`?

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+                start_indicies, n_mask
+            )  # [B, M'] (M' denotes sum of num_mixup over the batch)
+
+            idx = torch.arange(t_dim)
+            idx = torch.unsqueeze(idx, dim=-1) + start_indicies_flat  # [T, M']
+            mixup_values = self.cache[idx]  # [T, M', F]
+            return mixup_values
+
+
+class Mixup:
```
albertz, line 79:
```suggestion
class Mixup(torch.nn.Module):
```
?

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+        self.apply_prob = cfg.apply_prob
+        self.lambda_min = cfg.lambda_min
+        self.lambda_max = cfg.lambda_max
+        self.max_num_mix = cfg.max_num_mix
+        self.buffer_size = cfg.buffer_size
+        self.feature_dim = feature_dim
+
+        self.feature_buffer = FeatureBuffer(self.buffer_size, feature_dim)
+
+    def __call__(self, input: torch.Tensor, sequence_mask: torch.Tensor) -> torch.Tensor:
```
albertz, line 98: If you make `Mixup` a proper Torch module (I think it should be), then here:
```suggestion
    def forward(self, input: torch.Tensor, *, sequence_mask: torch.Tensor) -> torch.Tensor:
```

albertz, line 98: Document the expected shapes.

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+        assert (
+            input.size()[-1] == self.feature_dim
+        ), "the given feature dimension does not match input feature dimension"
+        input_ = self._maybe_apply_mixup(input, sequence_mask)
+        self._append_to_buffer(input, sequence_mask)
+        return input_
+
+    def _append_to_buffer(self, input: torch.Tensor, sequence_mask: torch.Tensor):
```
albertz, line 106: Can you document the shapes?

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+        _, _, f_dim = input.size()
+        # mask out the padded frames before append it to feature_buffer
+        input_flat = torch.masked_select(input, sequence_mask.unsqueeze(-1)).view(-1, f_dim)
+        self.feature_buffer.append(input_flat)
+
+    def _maybe_apply_mixup(self, input: torch.Tensor, sequence_mask: torch.Tensor) -> torch.Tensor:
```
albertz, line 112: Document the shapes.

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    The FeatureBuffer saves the feature from previous timeframes
+    after the buffer is full, the buffer will not be updated
+    """
+
+    def __init__(self, buffer_size: int, feature_dim: int):
+        self.filled = False
+        self.pos = 0
+        self.buffer_size = buffer_size
+        self.cache = torch.nn.parameter.Parameter(
```
albertz, line 39: If you make it a Torch module, I think you should use `register_buffer` here.

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+            data=torch.zeros((self.buffer_size, feature_dim)), requires_grad=False
+        )
```
albertz, line 41:
```suggestion
        self.register_buffer("cache", torch.zeros((self.buffer_size, feature_dim))
```

albertz:
> The Feature buffer will not be updated after it is full, I am not sure whether it would be better to keep updating the feature cache after it is full

Why not? In my implementation, it is being updated all the time. I think this is important. It is just overwriting the previous data in the cache with the more recent data.

i6_models/primitives/mixup.py
michelwi, line 43: I would instead of filling the buffer and then stopping to add data, start overwriting again from the start of the buffer.
e.g. set `self.pos=0` again (but keep `self.filled=True`)

michelwi (commented):
In this implementation 
* for each sequence a random number `n` between 1 and `max_num_mix` (e.g. 4) is drawn
* 4 different audio tracks are selected from the buffered data
* each sequence is perturbed with track 1
* the sequences with `n<=2` are perturbed with track 1 and 2
* the sequences with `n<=3` are perturbed with track 1, 2, and 3 (and so on)

is there any reason why to make it so complicated?

What I would have done:
* for each sequence a random number `n` between 1 and `max_num_mix` (e.g. 4) is drawn
* `\sum_b n_b`  different audio tracks are selected from the buffered data (this is M' from the inline comments)
* each sequence is perturbed with n separate tracks

Bonus implementation: with probability `1-apply_prob` `n_b` is set to 0 i.e. no mixup for a specific sequence.

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def _append_to_buffer(self, input: torch.Tensor, sequence_mask: torch.Tensor):
+        _, _, f_dim = input.size()
+        # mask out the padded frames before append it to feature_buffer
+        input_flat = torch.masked_select(input, sequence_mask.unsqueeze(-1)).view(-1, f_dim)
+        self.feature_buffer.append(input_flat)
+
+    def _maybe_apply_mixup(self, input: torch.Tensor, sequence_mask: torch.Tensor) -> torch.Tensor:
+        if random.random() >= self.apply_prob:
+            return input
```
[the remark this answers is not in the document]
albertz: This is intentionally on the whole batch for efficiency reasons.

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+
+        b_dim, t_dim, f_dim = input.size()  # [B, T, F]
+
+        max_end_idx = self.feature_buffer.buffer_size if self.feature_buffer.filled else self.feature_buffer.pos
+        if t_dim > max_end_idx:
+            return input
```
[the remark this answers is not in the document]
albertz: It would make the whole logic more complicated. That's why I decided against it in my implementation. As far as I remember, Zoltan also does the same. It's also maybe only the very first mini batch. I don't think this justifies more complicated code.

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+
+        num_mixup = torch.randint(low=1, high=self.max_num_mix + 1, size=(b_dim,))  # [B]
+        max_num_mixup = max(num_mixup)  # [M] (M denotes maximum of num_mixup)
```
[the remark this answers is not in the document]
albertz: Yea, the comment is wrong. The shape is [], you could write this like:
```suggestion
        max_num_mixup = max(num_mixup)  # scalar, =M (M denotes maximum of num_mixup)
```

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+
+        row_vector = torch.arange(0, max_num_mixup, 1)  # [M]
+        n_mask = torch.unsqueeze(num_mixup, dim=-1) > row_vector  # [B, M]
```
albertz, line 126: I think the broadcasting is more clear this way:
```suggestion
        n_mask = row_vector[None, :] < num_mixup[:, None]  # [B, M]
```

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+
+        mixup_values = self.feature_buffer.getRandom(b_dim, t_dim, max_num_mixup, n_mask).to(
```
albertz, line 128:
```suggestion
        mixup_values = self.feature_buffer.get_random(b_dim, t_dim, max_num_mixup, n_mask).to(
```

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+        self.cache[self.pos : end_pos] = tensor[:delta_pos]
+        self.pos = end_pos
+
+        if end_pos == self.buffer_size:
+            self.filled = True
+            end_pos = t_dim - delta_pos
+            self.cache[:end_pos] = tensor[delta_pos:]
+            self.pos = end_pos
+
+    def get_random(self, b_dim: int, t_dim: int, max_num_mixup: int, n_mask: torch.tensor) -> torch.Tensor:
```
albertz, line 59: Can you document the shapes (of `n_mask` and returned tensor)?

Also, wrong type:
```suggestion
    def get_random(self, b_dim: int, t_dim: int, max_num_mixup: int, n_mask: torch.Tensor) -> torch.Tensor:
```

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+        else:
+            end_idx = self.buffer_size if self.filled else self.pos
+            max_end_idx = end_idx - t_dim
+
+            start_indicies = torch.randint(
+                high=max_end_idx, size=(b_dim, max_num_mixup)
+            )  # [B, M] (M denotes maximum of num_mixup over the batch)
+            start_indicies_flat = torch.masked_select(
+                start_indicies, n_mask
+            )  # [B, M'] (M' denotes sum of num_mixup over the batch)
```
albertz, line 71: I think the shape is wrong? Shouldn't it be:
```suggestion
            start_indicies_flat = torch.masked_select(
                start_indicies, n_mask
            )  # [M'] (M' denotes sum of num_mixup over the batch)
```
?

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+
+            idx = torch.arange(t_dim)
+            idx = torch.unsqueeze(idx, dim=-1) + start_indicies_flat  # [T, M']
+            mixup_values = self.cache[idx]  # [T, M', F]
+            return mixup_values
```
albertz, line 76: I think writing it this way makes the shapes more clear:
```suggestion
            idx = torch.arange(t_dim)  # [T]
            idx = idx[:, None] + start_indicies_flat[None, :]  # [T, M']
            mixup_values = self.cache[idx]  # [T, M', F]
            return mixup_values
```

i6_models/primitives/mixup.py
albertz: Yes. In my original implementation, and also in Zoltans implementation, this is also the case. Why did you change this?

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+
+        row_vector = torch.arange(0, max_num_mixup, 1)  # [M]
+        n_mask = torch.unsqueeze(num_mixup, dim=-1) > row_vector  # [B, M]
+
+        mixup_values = self.feature_buffer.getRandom(b_dim, t_dim, max_num_mixup, n_mask).to(
+            input.device
+        )  # [T, M', F] (M' denotes sum of num_mixup over the batch)
+
+        lambda_ = torch.FloatTensor(b_dim, max(num_mixup)).uniform_(self.lambda_min, self.lambda_max)  # [B, M]
+        mixup_scales = torch.FloatTensor(b_dim, max(num_mixup)).uniform_(0.001, 1)  # [B, M]
```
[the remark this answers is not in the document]
albertz: This originates from Zoltans implementation. I'm not sure what would be the best. I think the scales themselves make sense on a linear scale (when we tune other scales, we also always do grid search on linear scale).

i6_models/primitives/mixup.py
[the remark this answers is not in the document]
michelwi: I doubt that it will be so much more efficient - especially when we select the weight of the mix-in sequences logarithmically and then do the mixup almost always and sometimes just with a small SNR... but ok.

i6_models/primitives/mixup.py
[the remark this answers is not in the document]
michelwi: ok.

i6_models/primitives/mixup.py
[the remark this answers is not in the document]
michelwi: I don't quite understand the lambda/mixup_scales. On a high level we mix the original audio with some sequence-noise which is some other audio in the corpus. The mixing should be done with different signal to noise ratios (SNR). And I would select the SNR on a log-uniform scale between some reasonable minimum (barely audible) and maximum (almost 1).

Is the lambda the selector for SNR and the mixup_scales is the weight how the 1,...,n individual noise sequences are mixed to get "the noise"? Then I would agree this should be uniform. And we can even do uniform from [0,1] - no need for lower cutoff.

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
+__all__ = ["MixupConfig", "Mixup"]
+
+import random
```
albertz, line 3: Do not use `random`. Use the Torch random number generators instead.

i6_models/primitives/mixup.py
```diff
@@ -0,0 +1,153 @@
...
+        return input_
+
+    def _append_to_buffer(self, input: torch.Tensor, sequence_mask: torch.Tensor):
+        _, _, f_dim = input.size()
+        # mask out the padded frames before append it to feature_buffer
+        input_flat = torch.masked_select(input, sequence_mask.unsqueeze(-1)).view(-1, f_dim)
+        self.feature_buffer.append(input_flat)
+
+    def _maybe_apply_mixup(self, input: torch.Tensor, sequence_mask: torch.Tensor) -> torch.Tensor:
+        if random.random() >= self.apply_prob:
```
albertz, line 113:
```suggestion
        if torch.randn(()) >= self.apply_prob:
```

i6_models/primitives/mixup.py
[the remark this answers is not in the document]
albertz: First think about the simple case, you take only one slice from the mixup cache. This is added to the input via same scale `lambda_` taken randomly from Uniform([lambda_min,lambda_max]), like `input = input + lambda_ * mixup_values`.

To have more noise, we don't just take a single slice, but multiple. But no matter how much are taken (as this is also a random var), we want that the total magnitude of what we add stays the same, i.e. like `lambda_`. So if we take N slices, we could just sample N `lambda_`s, or say `lambda_` with shape [N]. Now we could just weight the `lambda_` with 1/N each, such that when we sum over them, the total weight is still in [lambda_min,lambda_max].

Now, to add a bit more randomness, instead of using weights 1/N, we have these `mixup_scales`, and then multiply the `lambda_`s by `mixup_scales/sum(mixup_scales)`. I.e. again this makes sure that the total weight is still in [lambda_min,lambda_max], but just with even more randomness.

i6_models/primitives/mixup.py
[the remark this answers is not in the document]
michelwi: I meant that I would (conditioned on log-normal weights) select a very high value for `apply_prob` and then the case that mixup is not applied will not occur very often.

But then also the implementation of what happens (skip whole batch vs skip single sequence) is not so important (to me).

So I do accept the current implementation.

i6_models/primitives/mixup.py
Judyxujj: The current implementation of `self.pos` is exactly the same as Zoltan's and I think it is right.
```
        delta_pos = min(self.buffer_size - self.pos, t_dim)
        end_pos = self.pos + delta_pos
        self.cache[self.pos : end_pos] = tensor[:delta_pos]
        self.pos = end_pos

        if end_pos == self.buffer_size:
            self.filled = True
            end_pos = t_dim - delta_pos
            self.cache[:end_pos] = tensor[delta_pos:]
            self.pos = end_pos
```
After `self.pos `reaches `self.buffer_size`, if we want to add new values, here `delta_pos` is 0, `delta_pos` is  `self.pos`. The if condition is right, `end_pos` is `t_dim` and the `self.pos` becomes `t_dim`(which is the length of the newly added values)

Judyxujj:
> > The Feature buffer will not be updated after it is full, I am not sure whether it would be better to keep updating the feature cache after it is full
> 
> Why not? In my implementation, it is being updated all the time. I think this is important, esp if this is somewhere within the model, and not just at the beginning. It is just overwriting the previous data in the cache with the more recent data.



> > The Feature buffer will not be updated after it is full, I am not sure whether it would be better to keep updating the feature cache after it is full
> 
> Why not? In my implementation, it is being updated all the time. I think this is important, esp if this is somewhere within the model, and not just at the beginning. It is just overwriting the previous data in the cache with the more recent data.

The current implementation should be correct, I overlooked that after `end_pos == self.buffer_size`,    `self.pos` will be updated to `end_pos` which is `t_dim`        
```
            end_pos = t_dim - delta_pos
            self.cache[:end_pos] = tensor[delta_pos:]
            self.pos = end_pos
```

Judyxujj:
> In this implementation
> 
> * for each sequence a random number `n` between 1 and `max_num_mix` (e.g. 4) is drawn
> * 4 different audio tracks are selected from the buffered data
> * each sequence is perturbed with track 1
> * the sequences with `n<=2` are perturbed with track 1 and 2
> * the sequences with `n<=3` are perturbed with track 1, 2, and 3 (and so on)
> 
> is there any reason why to make it so complicated?
> 
> What I would have done:
> 
> * for each sequence a random number `n` between 1 and `max_num_mix` (e.g. 4) is drawn
> * `\sum_b n_b`  different audio tracks are selected from the buffered data (this is M' from the inline comments)
> * each sequence is perturbed with n separate tracks
> 
> Bonus implementation: with probability `1-apply_prob` `n_b` is set to 0 i.e. no mixup for a specific sequence.

I think there is some misunderstanding here. For each sequence, `n` is randomly selected as the number of times to apply mixup, we also have `n` corresponding mixup scales (they will be normalised by multiplying `lambda_/(sum(n mixup values)))`. Then `n` sequences of feature will be randomly selected from the feature buffer.  Then the `n` mixup scales will be  element wise multiply with `n` features.

i6_models/primitives/mixup.py
michelwi: The current implementation is technically correct and would work as-is.

I agree to 
> After self.pos reaches self.buffer_size, if we want to add new values, here delta_pos is 0

and then `end_pos` is `self.pos` so that we execute
```
self.cache[self.buffer_size: self.buffer_size] = tensor[:0]
```
i.e. no update is performed.

But what we would which to do is to change the current behavior and instead reset `self.pos=0` after we reach `self.buffer_size` so that the buffer is again overwritten from the start. This way, over the course of the training, we will not repeat the same mixup-noise over and over again but get fresh noise samples.

Judyxujj:
> > I think there is some misunderstanding here. In current implementation, for each sequence, `n` is randomly selected as the number of times to apply mixup, we also have `n` corresponding mixup scales (they will be normalised by multiplying `lambda_/(sum(n mixup values)))`. Then `n` sequences of feature will be randomly selected from the feature buffer. Then the `n` mixup scales will be element wise multiply with `n` features squences. I think the current implementation is same as you suggested except that there we apply normalisation to the mixup scales.
> 
> My suggestion was to sample `M'` sequences of features instead of `n`, where `M' = \sum_s n_s`.
> 
> E.g. if there is a batch of 3 sequences and the `n`'s that are sampled would be `[2, 4, 2]` then in (my understanding of) the current implementation 4 noise sequences are sampled and in my suggestion it would be 2+4+2=8
> 
> So am I wrong in my understanding of the current implementation or is my suggestion unclear?
> 
> > I think the current implementation is same as you suggested except that there we apply normalisation to the mixup scales.
> 
> we can/should keep correct normalization of mixup scales, I agree to that.

i6_models/primitives/mixup.py
Judyxujj: no, for the current implementation, `delta_pos` is 0, so `end_pos` is `t_dim`, then the cache will be updated 
`self.cache[:t_dim] = tensor[0:]`

michelwi: ah yes, sorry, I see now how in the if clause in lines 57ff there would indeed be an update and the buffer will be overwritten from the start.

michelwi:
> okay, now I understand what you meant. But why it is simpler than the current one?

haha, it is not simpler. More complicated actually. But I thought it might be helpful for the training to have different noises for different sequences.

Judyxujj:
> > okay, now I understand what you meant. But why it is simpler than the current one?
> 
> haha, it is not simpler. More complicated actually. But I thought it might be helpful for the training to have different noises for different sequences.

Using linear combination of same random sequences with different mix up scales should also give different noises. But I agree that it might be more robust to use different noises sequences. Albert and Zoltan both did mixup in this way. @albertz do you think it would help?

michelwi:
> i.e. every mixup of every sequence will get a different noise.

then I must have misread the code and am happy now.

michelwi (approved):

albertz (approved):

PR #48, opened 2024-04-19
Weight dropout

Judyxujj:
This PR adds the weight dropout

### DIFF
--- a/i6_models/primitives/weight_dropout.py
+++ b/i6_models/primitives/weight_dropout.py
@@ -1,0 +1,36 @@
+import torch.nn as nn
+import collections
+
+
+def _weight_drop(module, weights, dropout):
+    for name_w in weights:
+        w = getattr(module, name_w)
+        del module._parameters[name_w]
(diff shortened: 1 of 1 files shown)


Judyxujj:
> I think following `torch.nn.utils.weight_norm` or `torch.nn.utils.parametrizations.weight_norm` for this makes more sense, or not? This doesn't need to be (or rather should not be) specific for one module (here it is specific to `torch.nn.Linear`) but it should be generic, and less hacky, and simpler.

Judyxujj:
> I think following `torch.nn.utils.weight_norm` or `torch.nn.utils.parametrizations.weight_norm` for this makes more sense, or not? This doesn't need to be (or rather should not be) specific for one module (here it is specific to `torch.nn.Linear`) but it should be generic, and less hacky, and simpler.

But here we can also use the WeightDrop class to wrap other torch.nn.Module. I add

curufinwe:
No longer needed. If we are interested again we can use the pytorch functions that Albert mentioned.

PR #49, opened 2024-04-26
Add ConformerEncoderV2

Judyxujj:
For the second version of Conformer, two additional options are added:
1. enable swapping MHSA and Conv module
2. enable adding auxiliary losses on intermediate layers


i6_models/assemblies/conformer/conformer_v2.py
```diff
@@ -0,0 +1,132 @@
...
+        mhsa_cfg: Configuration for ConformerMHSAV1
+        conv_cfg: Configuration for ConformerConvolutionV1
+        swap_mhsa_conv: swap the execution order of MHSA and Conv module
+    """
+
+    # nested configurations
+    ff_cfg: ConformerPositionwiseFeedForwardV1Config
+    mhsa_cfg: ConformerMHSAV1Config
+    conv_cfg: ConformerConvolutionV1Config
+    swap_mhsa_conv: bool = True
```
mmz33, line 35: Probably better to have this `False` by default here.

i6_models/assemblies/conformer/conformer_v2.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def __init__(self, cfg: ConformerEncoderV2Config):
+        """
+        :param cfg: conformer encoder configuration with subunits for frontend and conformer blocks
+        """
+        super().__init__()
+
+        self.frontend = cfg.frontend()
+        self.module_list = torch.nn.ModuleList([ConformerBlockV2(cfg.block_cfg) for _ in range(cfg.num_layers)])
+        self.output_layers = list(cfg.loss_scales.keys())
```
mmz33, line 108: I am wondering if you really need to pass the whole `loss_scales` dict since the scales are not used in the `forward function`. Instead, you can pass the list of indicies directly.

Judyxujj: but later in train step, we need the loss scale, I think it would be easier to get the scales from `ctc.conformer. loss_scales` instead of passing it as an argument to train step function

i6_models/assemblies/conformer/conformer_v2.py
```diff
@@ -0,0 +1,132 @@
...
+        assert cfg.num_layers in self.output_layers, "The final layer must be included in loss layers"
+
+    def forward(
+        self, data_tensor: torch.Tensor, sequence_mask: torch.Tensor
```
albertz, line 112:
```suggestion
        self, data_tensor: torch.Tensor, /, sequence_mask: torch.Tensor
```
?
(Would be consistent to above.)

i6_models/assemblies/conformer/conformer_v2.py
```diff
@@ -0,0 +1,132 @@
...
+    ) -> Tuple[List[torch.Tensor], torch.Tensor]:
+        """
+        :param data_tensor: input tensor of shape [B, T', F]
+        :param sequence_mask: mask tensor where 1 defines positions within the sequence and 0 outside, shape: [B, T']
+        :return: (output, out_seq_mask)
+            where output is torch.Tensor of shape [B, T, F'],
```
albertz, line 118:
```suggestion
        :return: (outputs, out_seq_mask)
            where outputs is a list of torch.Tensor of shape [B, T, F']
            for each of the layers in output_layers,
```

i6_models/assemblies/conformer/conformer_v2.py
JackTemaki: After the latest experiments I would also agree to have it false, we still do not see consistent and reliable improvements...

i6_models/assemblies/conformer/conformer_v2.py
JackTemaki: I think Mohammad is right here, the `loss_scales` should be attached to your (own. custom) model config, not the encoder config. You can even put the parameter for which layers to return into the forward instead of the class That way a user can have different auxiliary losses during different training stages without having to modify the encoder code here.

i6_models/assemblies/conformer/conformer_v2.py
```diff
@@ -0,0 +1,132 @@
...
+    def __init__(self, cfg: ConformerEncoderV2Config):
+        """
+        :param cfg: conformer encoder configuration with subunits for frontend and conformer blocks
+        """
+        super().__init__()
+
+        self.frontend = cfg.frontend()
+        self.module_list = torch.nn.ModuleList([ConformerBlockV2(cfg.block_cfg) for _ in range(cfg.num_layers)])
+        self.output_layers = list(cfg.loss_scales.keys())
+        assert cfg.num_layers in self.output_layers, "The final layer must be included in loss layers"
```
JackTemaki, line 109: when you put this into the forward instead of the class init, you could even say the user is allowed to not access the final layer, and then you do a `min(len(...), max(return_layers))` to determine how long the module loop should work.

JackTemaki (changes requested):

i6_models/assemblies/conformer/conformer_v2.py
Judyxujj: I don't fully get the point here. If the final layer is not included and user is allowed to not access the final layer, then it would waste the memory to load the layers between [max(loss_layers),  num_layers].

JackTemaki: Yes, it would waste the memory but not the computation. I am just saying this is an unnecessary limitation in my view. I would just let this be user choice and not enforce it.

i6_models/assemblies/conformer/conformer_v2.py
JackTemaki: This is not resolved, I would still vote to have a `return_layers: List[int]` in the forward instead of having this static in the config. @mmz33 @albertz what is your opinion on this?

JackTemaki: @Judyxujj I know this might be overkill, but one use case would be to have "moving" auxiliary losses, that are not at fixed layers but start in early layers and move higher over the course of training.

i6_models/assemblies/conformer/conformer_v2.py
Judyxujj: could you explain more about `min(len(...), max(return_layers))`, I don't understand the idea here

JackTemaki: What I meant:
```python
        outputs = []
        for i in range(min(len(self.module_list), max(return_layers)):
            x = self.module_list[i](x, sequence_mask)  # [B, T, F']
            if i + 1 in return_layers:
                outputs.append(x)
                ```
But probably much better idea:
```python
        outputs = []
        assert max(return_layers) <= len(self.module_list)
        for i in range(max(return_layers)):
            x = self.module_list[i](x, sequence_mask)  # [B, T, F']
            if i + 1 in return_layers:
                outputs.append(x)
                ```

File: .github/workflows/black.yml [hunk 6]
```diff
@@ -10,8 +10,8 @@ jobs:
   check-black-formatting:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v4
```
albertz, line 14: I think this does not belong to the PR here. Is this needed for the PR? I would do a separate PR for this, and then rebase this one here.

.github/workflows/model_tests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
         with:
           repository: "rwth-i6/i6_models"
           path: ""
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v4
```
albertz, line 17: See above.

i6_models/assemblies/conformer/conformer_v2.py
```diff
@@ -0,0 +1,154 @@
+from __future__ import annotations
```
albertz, line 1: At some point, maybe in the (missing) docstring of the module, you should very briefly explain what's new in `conformer_v2` (just bullet points).

i6_models/assemblies/conformer/conformer_v2.py
```diff
@@ -0,0 +1,154 @@
...
+
+@dataclass
+class ConformerBlockV2Config(ModelConfiguration):
+    """
+    Attributes:
+        ff_cfg: Configuration for ConformerPositionwiseFeedForwardV1
+        mhsa_cfg: Configuration for ConformerMHSAV1
+        conv_cfg: Configuration for ConformerConvolutionV1
+        modules: List of modules to use for ConformerBlockV2,
+                       "ff" for feed forward module, "mhsa" for multi-head self attention module, "conv" for conv module
```
albertz, line 30: Indentation looks off. I think it should be indented by 4 spaces.
```suggestion
        modules: List of modules to use for ConformerBlockV2,
            "ff" for feed forward module, "mhsa" for multi-head self attention module, "conv" for conv module
```

i6_models/assemblies/conformer/conformer_v2.py
```diff
@@ -0,0 +1,154 @@
...
+        self.frontend = cfg.frontend()
+        self.module_list = torch.nn.ModuleList([ConformerBlockV2(cfg.block_cfg) for _ in range(cfg.num_layers)])
+
+    def forward(
+        self, data_tensor: torch.Tensor, /, sequence_mask: torch.Tensor, return_layers: Optional[List[int]] = None
+    ) -> Tuple[List[torch.Tensor], torch.Tensor]:
+        """
+        :param data_tensor: input tensor of shape [B, T', F]
+        :param sequence_mask: mask tensor where 1 defines positions within the sequence and 0 outside, shape: [B, T']
+        :param return_layers: list of layer indices specifying which layers to return, starting from 1
```
albertz, line 130: Why do you start counting from 1? To me (as someone who writes code) this was unexpected. I'm usually expecting always that indices start from 0 (at least in code).

.github/workflows/black.yml
JackTemaki: https://github.com/rwth-i6/i6_models/pull/52

i6_models/assemblies/conformer/conformer_v2.py
Judyxujj: they are now added to the doctoring of the module

i6_models/assemblies/conformer/conformer_v2.py
Judyxujj: The counting is changed to start from 0

JackTemaki (approved):
Did a full setup run and it works as intended

Atticus1806 (approved):

SimBe195 (approved):

Issue #50, opened 2024-04-26
Even kernel size for pooling layers in the GenericFrontend not permitted


kuacakuaca:
https://github.com/rwth-i6/i6_models/blob/3c9173691521778b1e8b4070c172cbe929e4826b/i6_models/parts/frontend/generic_frontend.py#L88

kuacakuaca:
> What is this issue about? You think this is a problem that even kernel sizes are not permitted? So you want to change that?

I want to use even kernel size for the pooling layers, is any reason we don’t want to allow this?

albertz:
As far as I remember, there were technical reasons why it did not work. (I actually was against having the check here at that point, because this makes it unclear why it was not allowed. Just trying it out shows the problem much more clearly.) Maybe it's actually not a problem anymore in a more recent PyTorch version? I don't know. Just try it out.

JackTemaki:
I think the limitation (from PyTorch side) was only for Convolution with "padding=same"...

kuacakuaca:
thanks! I'll just try.

PR #53, opened 2024-05-14
Feed forward layer, frontend and encoder

DanEnergetics:
Implementation of a simple feed forward encoder that serves to generate good alignments in full-sum HMM training. Apart from the layers, it implements a simple convolutional front-end that acts as a feed-forward layer that takes a window of feature time frames into account.


i6_models/parts/frontend/window_ffnn.py
```diff
@@ -0,0 +1,88 @@
...
+    def forward(self, x: torch.Tensor, /, sequence_mask: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        T might be reduced to T' on stride
+
+        :param x: input tensor of shape [B,T,F]
+        :param sequence_mask: the sequence mask for the tensor
+        :return: torch.Tensor of shape [B,T',F'] and the shape of the sequence mask
+        """
+        x = x.transpose(1, 2)  # torch 1d convolution is over last dim but we want time conv
+        x = self.conv(x).transpose(1, 2)
```
michelwi, line 76: I think we usually add annotations about the axes in comments, especially if we change them
```suggestion
        # torch 1d convolution is over last dim but we want time conv
        x = x.transpose(1, 2)  # [B, F, T]
        x = self.conv(x).transpose(1, 2)  # [B, T', F']
```

i6_models/parts/frontend/window_ffnn.py
```diff
@@ -0,0 +1,88 @@
...
+
+        # these settings apparently apply stride correctly to the masking whatever the kernel size
```
michelwi, line 78:
```suggestion
        # change masking according to stride value
```
do be confident about the implementation of others xP

i6_models/parts/frontend/window_ffnn.py
michelwi, line 1: I would maybe not name it "ffnn" / "FeedForward" but rather "convolution" since that is the layer that is being used?

DanEnergetics: I was unsure about naming and open to changes. Would you then rename both file and classes? Something like `window_convolution.py` then for the file?

i6_models/parts/frontend/window_ffnn.py
```diff
@@ -0,0 +1,88 @@
...
+    in_features: int
+    out_features: int
+    dropout: float
+    window_size: int
+    stride: int
+    activation: Union[nn.Module, Callable[[torch.Tensor], torch.Tensor]]
+
+    def __post_init__(self):
+        super().__post_init__()
+        assert self.window_size % 2 == 1, "Only odd kernel sizes are supported so far"
```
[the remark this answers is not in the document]
DanEnergetics: I actually found the same issue as mentioned in https://github.com/rwth-i6/i6_models/issues/50. It still does not work with even padding. However, I have a PR lined up with a simple padding function that would fix this.

i6_models/parts/frontend/window_ffnn.py
DanEnergetics: Actually I was not confident in my own implementation here. I had to choose `kernel_size = 1` and `padding = 0` to achieve correct masking, which is not how this function is supposed to be used I think.

i6_models/parts/frontend/window_ffnn.py
michelwi: yes, sounds good to me

i6_models/parts/ffnn.py
```diff
@@ -0,0 +1,59 @@
...
+import torch.nn.functional as F
+
+from i6_models.config import ModelConfiguration
+
+
+@dataclass
+class FeedForwardLayerV1Config(ModelConfiguration):
+    """
+    Attributes:
+        in_features: input feature dimension
```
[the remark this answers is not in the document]
michelwi: I think I would have preferred the parameter naming `input_dim` + `output_dim` (or use the same names as in `torch.nn.Linear`) but now it is at least consistent. :)

i6_models/parts/frontend/window_ffnn.py
michelwi: And with `kernel_size = cfg.window_size` and `padding = get_same_padding(cfg.window_size)` (i.e. the values given to `Conv1d`) it does not work?

michelwi (approved):

i6_models/parts/frontend/window_ffnn.py
DanEnergetics: Weirdly enough yes. I came up with a failing test case in a new branch https://github.com/rwth-i6/i6_models/blob/55a6fd6542037ed437a2a8d6d85ad98a1e8d73ee/tests/test_masking.py#L7

Atticus1806 (approved):
Took the liberty to commit one last change :)

PR #55, opened 2024-06-25
Add MHSA module, Conformer block and encoder with relative PE

kuacakuaca:


i6_models/parts/conformer/mhsa_rel_pos.py
```diff
@@ -0,0 +1,170 @@
...
+        # projection matrices
+        self.q_proj_weight = nn.parameter.Parameter(torch.empty((self.embed_dim, self.embed_dim)))
+        self.k_proj_weight = nn.parameter.Parameter(torch.empty((self.embed_dim, self.embed_dim)))
+        self.v_proj_weight = nn.parameter.Parameter(torch.empty((self.embed_dim, self.embed_dim)))
+
+        self.in_proj_bias = nn.parameter.Parameter(torch.empty(3 * self.embed_dim))
+
+        self.out_proj = nn.Linear(self.embed_dim, self.embed_dim, bias=True)
+
+        if self.rel_pos_clip:
```
michelwi, line 67: should we just disallow a value of 0 (and below) and then get rid of all the `if self.rel_pos_clip` blocks?

michelwi (approved):

michelwi:
In a discussion with @albertz and @JackTemaki I learned that MHSA tends to behave unstable under torch amp mixed precision training and that low precision is therefore explicitly disabled for the native attention implementation.

If we want to use this PR in training, we should think about manually disabling low precision for this mhsa module, or if it would be possible to add rel. pos. encoding to the native attention implementation instead.?

albertz:
> In a discussion with @albertz and @JackTemaki I learned that MHSA tends to behave unstable under torch amp mixed precision training

I'm not 100% sure though that I remember the details correctly. Currently I don't find this. Also, this is maybe more a problem with float16 AMP training and not really an issue with bfloat16.

(Btw, when I use AMP, I always use it with bfloat16 and no grad scaler; and I also have my custom (RF-based) self-attention where I don't have any special treatment for AMP, so at least for me, it also works fine.)

(I was searching now a bit. Maybe related: https://github.com/pytorch/pytorch/issues/40497)

i6_models/parts/conformer/mhsa_rel_pos.py
```diff
@@ -0,0 +1,180 @@
...
+                batch_dim_size, self.num_heads, time_dim_size, time_dim_size
+            )  # [B, #heads, T, T']
+
+            attn = (attn1 + attn2 + mask) * (math.sqrt(1.0 / float(self.embed_dim_per_head)))  # [B, #heads, T, T']
+        else:
+            attn = (attn1 + mask) * (math.sqrt(1.0 / float(self.embed_dim_per_head)))  # [B, #heads, T, T']
+
+        # softmax and dropout
+        attn_output_weights = F.dropout(
+            F.softmax(attn, dim=-1), p=self.att_weights_dropout, training=self.training
```
Judyxujj, line 158: For the tensorflow backend, is the broadcast also applied to the attention weight dropout?

albertz: That depends on how the user configures it. The RETURNN TF backend can do whatever you want. But I think most people did not apply the broadcast here.

i6_models/parts/conformer/convolution.py
```diff
@@ -20,13 +20,15 @@ class ConformerConvolutionV1Config(ModelConfiguration):
         dropout: dropout probability
         activation: activation function applied after normalization
         norm: normalization layer with input of shape [N,C,T]
+        broadcast_dropout: whether to broadcast dropout on the feature axis to time axis
```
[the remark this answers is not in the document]
albertz: In the RETURNN TF backend, we simply specified a list of axes where to apply broadcasting or not broadcasting. I find this more intuitive and explicit than having here an enum with a list of all possibilities.

i6_models/parts/conformer/mhsa_rel_pos.py
Judyxujj: Okay, then it is fine

Judyxujj (approved):

kuacakuaca:
> I also want to take a look.
> 
> What kind of rel pos encoding is implemented here?

Shaw's https://arxiv.org/abs/1803.02155.

albertz:
> > What kind of rel pos encoding is implemented here?
> 
> Shaw's https://arxiv.org/abs/1803.02155.

But is this a good idea?

I think in general, Dai et al. Transformer-XL style 2019 is more standard, and I think usually better. I have also done direct comparisons where it was always consistently better.

In RETURNN frontend, I have a generic `RelPosSelfAttention` implementation which covers both (and more) cases:

https://github.com/rwth-i6/returnn/blob/23d666ccf3ac9e748fce4e0d27afe353133eca48/returnn/frontend/attention.py#L267-L401

In the end, you have five options, and can get both cases:

To get Shawn et al. self-att rel pos 2018 / RETURNN SelfAttentionLayer + RelativePositionalEncodingLayer:

- with_bias = False (at least that was the RETURNN behavior)
- with_linear_pos = False
- with_pos_bias = False
- learnable_pos_emb = True
- separate_pos_emb_per_head = False (at least that was the RETURNN default)

To get Dai et al. Transformer-XL style 2019:

- with_bias = False would be like the paper, however, in most implementations it is True (default)
- with_linear_pos = True (default)
- with_pos_bias = True (default)
- learnable_pos_emb = True (default)
- separate_pos_emb_per_head = True (default)

kuacakuaca:
> > > What kind of rel pos encoding is implemented here?
> > 
> > 
> > Shaw's https://arxiv.org/abs/1803.02155.
> 
> But is this a good idea?
> 
> I think in general, Dai et al. Transformer-XL style 2019 is more standard, and I think usually better. I have also done direct comparisons where it was always consistently better.
> 
> In RETURNN frontend, I have a generic `RelPosSelfAttention` implementation which covers both (and more) cases:
> 
> https://github.com/rwth-i6/returnn/blob/23d666ccf3ac9e748fce4e0d27afe353133eca48/returnn/frontend/attention.py#L267-L401
> 
> In the end, you have five options, and can get both cases:
> 
> To get Shawn et al. self-att rel pos 2018 / RETURNN SelfAttentionLayer + RelativePositionalEncodingLayer:
> 
>     * with_bias = False (at least that was the RETURNN behavior)
> 
>     * with_linear_pos = False
> 
>     * with_pos_bias = False
> 
>     * learnable_pos_emb = True
> 
>     * separate_pos_emb_per_head = False (at least that was the RETURNN default)
> 
> 
> To get Dai et al. Transformer-XL style 2019:
> 
>     * with_bias = False would be like the paper, however, in most implementations it is True (default)
> 
>     * with_linear_pos = True (default)
> 
>     * with_pos_bias = True (default)
> 
>     * learnable_pos_emb = True (default)
> 
>     * separate_pos_emb_per_head = True (default)

so you suggest to make the type of relative positional encoding configurable? I'll check your code. thx.

albertz:
> so you suggest to make the type of relative positional encoding configurable? I'll check your code. thx.

Ideally that, yes. You could simply follow the existing linked code from RETURNN frontend. It also doesn't really increase the code complexity to have those options configurable. But if you really want only one case, I would go with the more standard and known-to-be better case.

albertz:
I wonder, what's your motivation of changing from matmul to einsum? While einsum is definitely more readable, matmul is usually a bit faster. The overhead of einsum is probably not so much but I just want to raise this that we make a deliberate choice, what we prefer: more efficient code, or more readable code.

kuacakuaca:
> I wonder, what's your motivation of changing from matmul to einsum? While einsum is definitely more readable, matmul is usually a bit faster. The overhead of einsum is probably not so much but I just want to raise this that we make a deliberate choice, what we prefer: more efficient code, or more readable code.

Initially it was just an attempt to solve the unstable AMP training issue Willi had, but it turned out it was not because of `torch.matmul`. I will just revert the change.

kuacakuaca:
@albertz I didn't see in the `_rel_pos_enc_shift` method https://github.com/rwth-i6/returnn/blob/eb10e7ce90a4e12568f58b50b39116ca139a3012/returnn/frontend/attention.py#L404 the upper triangular part is set to 0, isn't set necessary?

albertz:
> @albertz I didn't see in the `_rel_pos_enc_shift` method https://github.com/rwth-i6/returnn/blob/eb10e7ce90a4e12568f58b50b39116ca139a3012/returnn/frontend/attention.py#L404 the upper triangular part is set to 0, isn't that necessary?

Why?

Are you referring to the ESPnet code?
`RelPositionMultiHeadedAttention`: https://github.com/espnet/espnet/blob/b8b31c06da081ee1690c8dc1bf102a7811124529/espnet/nets/pytorch_backend/transformer/attention.py#L232
`RelPositionalEncoding`: https://github.com/espnet/espnet/blob/b8b31c06da081ee1690c8dc1bf102a7811124529/espnet/nets/pytorch_backend/transformer/embedding.py#L260
There is such `zero_triu` option but it's False by default. Maybe it's to only make it casual, i.e. only attend to the past? I don't remember the details. As I write on `_rel_pos_enc_shift`:

>  Reshape + slice trickery. You need to draw the 2D arrays on paper to understand this.
> Also see similar trickery in :func:`window`.

albertz:
Btw, while looking at the ESPnet code that I linked again, I noticed that my statement was a bit wrong that they use exactly the XL variant:
* Instead of learnable rel pos encoding, they use sinusoidal rel pos encoding. (In RF: `learnable_pos_emb=False`)
* They apply some dropout by default on the pos encoding (same dropout rate as everywhere else: 0.1). (In RF: `pos_emb_dropout=0.1`.)

But I'm pretty sure that I saw learnable pos encoding more often elsewhere.

I think no-one really investigated each of these aspects individually. I only compared Shawn et al. vs Dai et al. so far.

kuacakuaca:
> Btw, while looking at the ESPnet code that I linked again, I noticed that my statement was a bit wrong that they use exactly the XL variant:
> 
>     * Instead of learnable rel pos encoding, they use sinusoidal rel pos encoding. (In RF: `learnable_pos_emb=False`)
> 
>     * They apply some dropout by default on the pos encoding (same dropout rate as everywhere else: 0.1). (In RF: `pos_emb_dropout=0.1`.)
> 
> 
> But I'm pretty sure that I saw learnable pos encoding more often elsewhere.
> 
> I think no-one really investigated each of these aspects individually. I only compared Shawn et al. vs Dai et al. so far.

in case of Transformer-XL relative PE (as in the original paper), `learnable_pe_emb` should be set to `False` and `with_pos_bias` to `True`. But you mean the community uses learnable PE (`learnable_pos_emb=True`) + pos bias (`with_pos_bias=True`) more often?
so basically 4 cases, `learnable/non-learnable PE `X` w./w.o. pos bias` should be covered here. Have you or anyone else tested `with_linear_pos`, `separate_pos_emb_per_head` or `pos_emb_dropout`?

albertz:
Sorry for not answering earlier, but I didn't have time yet to look into it. I was pretty sure that my Dai et al. Transformer-XL style 2019 settings that I stated were correct, but maybe I confused sth? I don't remember the details anymore. I will check that again. Maybe they also had multiple variants? I think they also have a reference implementation. Maybe I followed that one? It would also be useful to give some further references on how other people in the community use it, e.g. ESPnet, but also Fairseq and whatever else you find.

In any case, with those settings that I stated for Dai et al. Transformer-XL style 2019, i.e. the current default RF settings, that is where I got consistently better than with Shawn et al. self-att rel pos 2018.

JackTemaki (changes requested):
In the PR you are adding new parameters into existing modules, this would break all existing hashes. You need to do new versions for that (which can be derived of version 1 to save code though).

kuacakuaca:
> In the PR you are adding new parameters into existing modules, this would break all existing hashes. You need to do new versions for that (which can be derived of version 1 to save code though).
what goes into the hashes regarding i6_models in i6? the config objects?

JackTemaki:
> what goes into the hashes regarding i6_models in i6? the config objects?

Yes, exactly.

christophmluscher (approved):
After offline discussion a bit more documentation will be added but other wise LGTM

i6_models/parts/conformer/convolution.py
```diff
@@ -85,3 +90,78 @@ def forward(self, tensor: torch.Tensor) -> torch.Tensor:
...
+
+    def __init__(self, model_cfg: ConformerConvolutionV2Config):
+        """
+        :param model_cfg: model configuration for this module
+        """
+        super().__init__(model_cfg)
+
+        self.dropout_broadcast_axes = model_cfg.dropout_broadcast_axes
+        self.dropout = (
+            nn.Dropout1d(model_cfg.dropout) if model_cfg.dropout_broadcast_axes else nn.Dropout(model_cfg.dropout)
```
JackTemaki, line 129: Not a strong issue, but why do you have dropout here (in convolution) as explicit module but not for Feed-Forward and MHSA? I think this should be consistent. @michelwi @christophmluscher @curufinwe I would prefer to have it explicit, because then it is printed when you print the model.

JackTemaki (changes requested):
Reject to discuss the dropout inconsistency, but I would accept if nobody else cares about this.

i6_models/parts/conformer/convolution.py
kuacakuaca: > Not a strong issue, but why do you have dropout here (in convolution) as explicit module but not for Feed-Forward and MHSA? I think this should be consistent. @michelwi @christophmluscher @curufinwe I would prefer to have it explicit, because then it is printed when you print the model.

I just followed the V1 implementations. But I could make them all modules. Anyone against that?

christophmluscher: I am OK with the suggested change. It would improve readability :)

JackTemaki (approved):
I did not do a super detailed review, but no more comments from my side

albertz:
Now I checked again. In my earlier post "To get Dai et al. Transformer-XL style 2019" (https://github.com/rwth-i6/i6_models/pull/55#issuecomment-2232805347), where I stated what defaults we have for RF `RelPosSelfAttention`, I made a copy-and-paste error for `learnable_pos_emb`. The default in RF is False, which is as in the Trafo-XL paper, and also like in ESPnet.

i6_models/parts/conformer/convolution.py
```diff
@@ -85,3 +90,78 @@ def forward(self, tensor: torch.Tensor) -> torch.Tensor:
...
+            tensor = self.dropout(tensor.permute(1, 2, 0)).permute(2, 0, 1)
+        elif self.dropout_broadcast_axes == "BT":
+            batch_dim_size = tensor.shape[0]
+            feature_dim_size = tensor.shape[-1]
+
+            tensor = (
+                self.dropout(tensor.reshape(-1, feature_dim_size).transpose(0, 1))
+                .transpose(0, 1)
+                .reshape(batch_dim_size, -1, feature_dim_size)
+            )
```
albertz, line 165: I would suggest to have our own `Dropout` module which support this `dropout_broadcast_axes` argument, and then does this logic, but better commented. The code here is really hard to follow, esp also because there are no comments on what shape you get. Then the code here simply becomes:
```suggestion
        tensor = self.dropout(tensor)
```

i6_models/parts/conformer/feedforward.py
```diff
@@ -53,3 +58,70 @@ def forward(self, tensor: torch.Tensor) -> torch.Tensor:
...
+    """
+
+    dropout_broadcast_axes: Optional[str] = None
+
+    def check_valid(self):
+        assert self.dropout_broadcast_axes is None or self.dropout_broadcast_axes in [
+            "B",
+            "T",
+            "BT",
+        ], "invalid value, supported are None, 'B', 'T' and 'BT'"
```
albertz, line 78:
```suggestion
        assert self.dropout_broadcast_axes in [
            None,
            "B",
            "T",
            "BT",
        ], "invalid value, supported are None, 'B', 'T' and 'BT'"
```
Looks a bit cleaner to me.

i6_models/parts/conformer/feedforward.py
```diff
@@ -53,3 +58,70 @@ def forward(self, tensor: torch.Tensor) -> torch.Tensor:
...
+
+@dataclass
+class ConformerPositionwiseFeedForwardV2Config(ConformerPositionwiseFeedForwardV1Config):
+    """
+    New attribute:
+        dropout_broadcast_axes: string of axes to which dropout is broadcast, e.g. "T" for broadcasting to the time axis
+                                setting to None to disable broadcasting
+    """
+
+    dropout_broadcast_axes: Optional[str] = None
```
albertz, line 71: I forgot, do we assume Python >=3.8? Then we can also have this here:
```suggestion
    dropout_broadcast_axes: Optional[Literal["B", "T", "BT"]] = None
```

i6_models/parts/conformer/feedforward.py
```diff
@@ -53,3 +58,70 @@ def forward(self, tensor: torch.Tensor) -> torch.Tensor:
...
+            tensor = self.dropout(tensor.permute(1, 2, 0)).permute(2, 0, 1)
+        elif self.dropout_broadcast_axes == "BT":
+            batch_dim_size = tensor.shape[0]
+            feature_dim_size = tensor.shape[-1]
+
+            tensor = (
+                self.dropout(tensor.reshape(-1, feature_dim_size).transpose(0, 1))
+                .transpose(0, 1)
+                .reshape(batch_dim_size, -1, feature_dim_size)
+            )
```
albertz, line 111: As before, move the dropout logic to an own module, and then you have simply:
```suggestion
        tensor = self.dropout(tensor)
```

i6_models/parts/conformer/mhsa.py
```diff
@@ -1,12 +1,14 @@
...
 
-__all__ = ["ConformerMHSAV1", "ConformerMHSAV1Config"]
+__all__ = ["ConformerMHSAV1", "ConformerMHSAV1Config", "ConformerMHSAV2", "ConformerMHSAV2Config"]
 from dataclasses import dataclass
 import torch
 
 from i6_models.config import ModelConfiguration
 from i6_models.util import compat
 
+from typing import Optional
```
albertz, line 10: The order of imports is inconsistent. Usually we would have all stdlib imports before our own imports.

i6_models/parts/conformer/convolution.py
```diff
@@ -85,3 +90,78 @@ def forward(self, tensor: torch.Tensor) -> torch.Tensor:
...
+    dropout_broadcast_axes: Optional[str] = None
+
+    def check_valid(self):
+        assert self.kernel_size % 2 == 1, "ConformerConvolutionV1 only supports odd kernel sizes"
+
+        assert self.dropout_broadcast_axes is None or self.dropout_broadcast_axes in [
+            "B",
+            "T",
+            "BT",
+        ], "invalid value, supported are None, 'B', 'T' and 'BT'"
```
albertz, line 113:
```suggestion
        assert self.dropout_broadcast_axes in [
            None,
            "B",
            "T",
            "BT",
        ], "invalid value, supported are None, 'B', 'T' and 'BT'"
```

i6_models/parts/conformer/convolution.py
```diff
@@ -85,3 +90,78 @@ def forward(self, tensor: torch.Tensor) -> torch.Tensor:
...
+@dataclass
+class ConformerConvolutionV2Config(ConformerConvolutionV1Config):
+    """
+    New attribute:
+        dropout_broadcast_axes: string of axes to which dropout is broadcast, e.g. "T" for broadcasting to the time axis
+                                setting to None to disable broadcasting
+    Allows even kernel size
+    """
+
+    dropout_broadcast_axes: Optional[str] = None
```
albertz, line 104:
```suggestion
    dropout_broadcast_axes: Optional[Literal["B", "T", "BT"]] = None
```

i6_models/parts/conformer/mhsa.py
```diff
@@ -60,3 +62,71 @@ def forward(self, input_tensor: torch.Tensor, sequence_mask: torch.Tensor) -> to
...
+
+@dataclass
+class ConformerMHSAV2Config(ConformerMHSAV1Config):
+    """
+    New attribute:
+        dropout_broadcast_axes: string of axes to which dropout is broadcast, e.g. "T" for broadcasting to the time axis
+                                setting to None to disable broadcasting
+    """
+
+    dropout_broadcast_axes: Optional[str] = None
```
albertz, line 75:
```suggestion
    dropout_broadcast_axes: Optional[Literal["B", "T", "BT"]] = None
```

i6_models/parts/conformer/mhsa.py
```diff
@@ -60,3 +62,71 @@ def forward(self, input_tensor: torch.Tensor, sequence_mask: torch.Tensor) -> to
...
+
+    def check_valid(self):
+        assert self.dropout_broadcast_axes is None or self.dropout_broadcast_axes in [
+            "B",
+            "T",
+            "BT",
+        ], "invalid value, supported are None, 'B', 'T' and 'BT'"
```
albertz, line 82:
```suggestion
        assert self.dropout_broadcast_axes in [
            "B",
            "T",
            "BT",
        ], "invalid value, supported are None, 'B', 'T' and 'BT'"
```

i6_models/parts/conformer/mhsa_rel_pos.py
```diff
@@ -0,0 +1,248 @@
...
+        self.dropout_broadcast_axes = cfg.dropout_broadcast_axes
+
+        self._reset_parameters()  # initialize parameters
+
+    def _reset_parameters(self):
+        nn.init.xavier_uniform_(self.q_proj_weight)
+        nn.init.xavier_uniform_(self.k_proj_weight)
+        nn.init.xavier_uniform_(self.v_proj_weight)
+
+        # TODO: choose kind of initialization
```
albertz, line 114: Do we really want to leave such a TODO in the code? Why is it even here? Do you mean you did not play around with variants here? But this is true to some degree probably for everything else as well? We can never really now what is the best variant.

A good starting point is just to follow other existing implementations, e.g. ESPnet. But if you do that already, I don't think you need the TODO here.

i6_models/parts/conformer/mhsa_rel_pos.py
```diff
@@ -0,0 +1,248 @@
...
+        if self.learnable_pos_emb:
+            nn.init.normal_(self.rel_pos_embeddings)
+        if self.with_pos_bias:
+            nn.init.constant_(self.pos_bias_u, 0.0)
+            nn.init.constant_(self.pos_bias_v, 0.0)
```
albertz, line 119: I just checked, why do you do this different to ESPnet? In ESPnet, they do:
```python
        torch.nn.init.xavier_uniform_(self.pos_bias_u)
        torch.nn.init.xavier_uniform_(self.pos_bias_v)
```
So do the same?
```suggestion
            nn.init.xavier_uniform_(self.pos_bias_u)
            nn.init.xavier_uniform_(self.pos_bias_v)
```

i6_models/parts/conformer/mhsa_rel_pos.py
```diff
@@ -0,0 +1,248 @@
...
+            output_tensor = self.dropout(output_tensor.permute(1, 2, 0)).permute(2, 0, 1)
+        elif self.dropout_broadcast_axes == "BT":
+            batch_dim_size = output_tensor.shape[0]
+            feature_dim_size = output_tensor.shape[-1]
+
+            output_tensor = (
+                self.output(output_tensor.reshape(-1, feature_dim_size).transpose(0, 1))
+                .transpose(0, 1)
+                .reshape(batch_dim_size, -1, feature_dim_size)
+            )
```
albertz, line 218: Again simplify this.
```suggestion
        output_tensor = self.dropout(output_tensor)
```

i6_models/parts/conformer/mhsa_rel_pos.py
```diff
@@ -0,0 +1,248 @@
...
+        if self.learnable_pos_emb:
+            self.rel_pos_embeddings = nn.parameter.Parameter(
+                torch.empty(self.rel_pos_clip * 2 + 1, self.embed_dim // self.num_heads)
+            )
+        if self.with_pos_bias:
+            self.pos_bias_u = nn.parameter.Parameter(torch.empty(self.num_heads, self.embed_dim_per_head))
+            self.pos_bias_v = nn.parameter.Parameter(torch.empty(self.num_heads, self.embed_dim_per_head))
+
+        self.dropout = nn.Dropout1d(cfg.dropout) if cfg.dropout_broadcast_axes else nn.Dropout(cfg.dropout)
+        self.dropout_broadcast_axes = cfg.dropout_broadcast_axes
```
albertz, line 105: Again use our own custom `Dropout`. Or like this:
```suggestion
        self.dropout = Dropout(cfg.dropout, cfg.dropout_broadcast_axes) if cfg.dropout_broadcast_axes else nn.Dropout(cfg.dropout)
```
Or just:
```suggestion
        self.dropout = Dropout(cfg.dropout, cfg.dropout_broadcast_axes)
```

i6_models/parts/conformer/mhsa_rel_pos.py
```diff
@@ -0,0 +1,248 @@
...
+        dropout: multi-headed self attention output dropout
+        dropout_broadcast_axes: string of axes to which dropout is broadcast, e.g. "T" for broadcasting to the time axis
+                                setting to None to disable broadcasting
+    """
+
+    input_dim: int
+    num_att_heads: int
+    att_weights_dropout: float
+    dropout: float
+    learnable_pos_emb: bool = True
```
albertz, line 38: This default is different from Trafo XL (sorry, my bad that I wrote it wrong earlier). But also: I thought that we said we do not want any defaults? So why here? But then, what default do you want? To what is commonly used (ESPnet for Conformer, or also Trafo-XL, etc?), then it should be False. Or what we had before?
```suggestion
    learnable_pos_emb: bool = False
```

i6_models/parts/conformer/mhsa_rel_pos.py
```diff
@@ -0,0 +1,248 @@
...
+    rel_pos_clip: Optional[int] = None
+    with_pos_bias: bool = False
+    pos_emb_dropout: float = 0.0
```
albertz, line 41: You currently have those self-att related settings:

```
    num_att_heads: int  # num_heads in RF
    att_weights_dropout: float  # att_dropout in RF
    learnable_pos_emb: bool = True  # learnable_pos_emb in RF
    rel_pos_clip: Optional[int] = None  # learnable_pos_emb_clipping in RF
    with_pos_bias: bool = False  # with_pos_bias in RF
    pos_emb_dropout: float = 0.0  # pos_emb_dropout in RF
```

So, in RF, we have in addition those:
```
        with_bias: bool = True,
        with_linear_pos: bool = True,
        separate_pos_emb_per_head: bool = True,
```

`with_bias`: You implement it for True. But I have recently seen many people using False, e.g. Llama. So we might want to have this option.

`with_linear_pos`: I think you miss this. Why? This is different to ESPnet and I think also to how it is commonly done.

`separate_pos_emb_per_head`: You implement it for True. So maybe it's fine to not have this configurable now.

i6_models/parts/conformer/mhsa_rel_pos.py
```diff
@@ -0,0 +1,248 @@
...
+        self.att_weights_dropout = nn.Dropout(cfg.att_weights_dropout)
+
+        assert self.embed_dim % self.num_heads == 0, "embed_dim must be divisible by num_heads"
+
+        # projection matrices
+        self.q_proj_weight = nn.parameter.Parameter(torch.empty((self.embed_dim, self.embed_dim)))
+        self.k_proj_weight = nn.parameter.Parameter(torch.empty((self.embed_dim, self.embed_dim)))
+        self.v_proj_weight = nn.parameter.Parameter(torch.empty((self.embed_dim, self.embed_dim)))
+
+        self.in_proj_bias = nn.parameter.Parameter(torch.empty(3 * self.embed_dim))
```
albertz, line 88: I don't understand, why do you keep the weight matrices for qkv separate, but then concat it for the bias? Also calling it `in_proj` was confusing to me, I would have called it `qkv_proj`. But anyway, why not simply use `nn.Linear`? So like this:
```suggestion
        self.qkv = nn.Linear(self.embed_dim, 3 * self.embed_dim)
```
Also see the RF code, which does it exactly like that.

tests/test_conformer_rel_pos.py
```diff
@@ -0,0 +1,121 @@
...
+
+    for learnable_pos_emb, with_pos_bias, pos_emb_dropout in product([True, False], [True, False], [0.0, 0.1]):
+        assert get_output_shape(
+            input_shape,
+            seq_len,
+            32,
+            learnable_pos_emb=learnable_pos_emb,
+            with_pos_bias=with_pos_bias,
+            pos_emb_dropout=pos_emb_dropout,
+        ) == [4, 15, 32]
```
albertz, line 121: I think I would add some test case which compares the output to ESPnet, so that you are sure to have it 100% correct.

albertz (changes requested):

i6_models/parts/conformer/mhsa_rel_pos.py
kuacakuaca: `separate_pos_emb_per_head` was actually False in my code, now I added it.

tests/test_conformer_rel_pos.py
kuacakuaca: I did a local comparison, the outputs were close enough. but I needed to modify our `ConformerMHSARelPosV1` to make the parameters compatible, e.g. espnet uses separate q, k and v matrices.

i6_models/parts/conformer/convolution.py
kuacakuaca: do you any objection to setting the default behavior to no broadcasting?

tests/test_conformer_rel_pos.py
albertz: You don't need to modify this. When you import the model parameters, you just need to convert it properly. I think it makes sense to add such a test case.

i6_models/parts/conformer/mhsa_rel_pos.py
```diff
@@ -0,0 +1,248 @@
...
+        attn_output_weights = self.att_weights_dropout(F.softmax(attn_scaled, dim=-1))  # [B, #heads, T, T']
+
+        # sequence of weighted sums over value sequence
+        v = value_seq.view(batch_dim_size, -1, self.num_heads, self.embed_dim_per_head)  # [B, T, H, F']
+        attn_output = (
+            torch.einsum("bhij, bjhf -> bhif", attn_output_weights, v)
+            .transpose(1, 2)
+            .contiguous()
+            .view(batch_dim_size, -1, self.embed_dim)
+        )
```
albertz, line 214: I don't understand:
* Why `contiguous` and then `view`? There is `reshape` exactly for that.
* Why the `transpose` when you already have `einsum` before?

So, shouldn't this be the same but just simpler?
```suggestion
        attn_output = (
            torch.einsum("bhij, bjhf -> bihf", attn_output_weights, v)
            .reshape(batch_dim_size, -1, self.embed_dim)
        )
```

i6_models/parts/dropout.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def forward(self, tensor: torch.Tensor) -> torch.Tensor:
+        """
+        assumes input tensor of shape [B, T, F]
+        return tensor of shape [B, T, F]
+        """
+        if self.dropout_broadcast_axes is None:
+            tensor = torch.nn.functional.dropout(tensor, p=self.p, training=self.training)
+        elif self.dropout_broadcast_axes == "T":  # [B, T, F] -> [B, F, T] -> [B, T, F]
+            tensor = torch.nn.functional.dropout1d(tensor.transpose(1, 2), p=self.p, training=self.training).transpose(
```
albertz, line 36: I think it would be helpful to add a small comment here saying that `torch.nn.functional.dropout1d` expects a 3D tensor and broadcasts in the last dimension.
```suggestion
            # torch.nn.functional.dropout1d expects a 3D tensor and broadcasts in the last dimension.
            tensor = torch.nn.functional.dropout1d(tensor.transpose(1, 2), p=self.p, training=self.training).transpose(
```

i6_models/parts/conformer/mhsa_rel_pos.py
```diff
@@ -0,0 +1,248 @@
...
+        time_dim_size = output_tensor.shape[1]
+        batch_dim_size = output_tensor.shape[0]
+
+        # attention mask
+        # T: query seq. length, T' key/value seg length; T = T' if same input tensor
+        inv_sequence_mask = compat.logical_not(sequence_mask)  # [B, T']
+        mask = (
+            torch.zeros_like(inv_sequence_mask, dtype=input_tensor.dtype)
+            .masked_fill(inv_sequence_mask, float("-inf"))
+            .view(batch_dim_size, 1, 1, time_dim_size)
```
albertz, line 147:
```suggestion
            .reshape(batch_dim_size, 1, 1, time_dim_size)
```
(Btw, in general, you should almost never use `view`, unless you have a very good reason for it.)

kuacakuaca: why? I thought view() could ensure no extra copy is done and save some memory.

kuacakuaca: okay, reshape could do the same thing.

albertz: `view` fails if it cannot do it, while `reshape` always works, and does the same saving of memory if possible.

i6_models/parts/conformer/mhsa_rel_pos.py
```diff
@@ -0,0 +1,245 @@
...
+        att_weights_dropout: attention weights dropout
+        learnable_pos_emb: whether to use learnable relative positional embeddings instead of fixed sinusoidal ones
+        rel_pos_clip: maximal relative postion for embedding
+        with_linear_pos: whether to linearly transform the positional embeddings
+        separate_pos_emb_per_head: whether to create head-dependent positional embeddings
+        with_pos_bias: whether to add additional position bias terms to the attention scores
+        pos_emb_dropout: dropout for the positional embeddings
+        dropout: multi-headed self attention output dropout
+        dropout_broadcast_axes: string of axes to which dropout is broadcast, e.g. "T" for broadcasting to the time axis
+                                setting to None to disable broadcasting
```
albertz, line 34: I think it would be good to extend the docs, to 1. list what you need to set to get Shawn et al. self-att rel pos 2018 / RETURNN SelfAttentionLayer + RelativePositionalEncodingLayer, and 2. for Dai et al. Transformer-XL style 2019. More or less like the docstring in RETURNN:
```
To get Shawn et al. self-att rel pos 2018 / RETURNN SelfAttentionLayer + RelativePositionalEncodingLayer:

with_bias = False (at least that was the RETURNN behavior)
with_linear_pos = False
with_pos_bias = False
learnable_pos_emb = True
separate_pos_emb_per_head = False (at least that was the RETURNN default)

To get Dai et al. Transformer-XL style 2019:

with_bias = False would be like the paper, however, in most implementations it is True (default)
with_linear_pos = True (default)
with_pos_bias = True (default)
learnable_pos_emb = True False (default)
separate_pos_emb_per_head = True (default)
```

i6_models/parts/conformer/mhsa_rel_pos.py
```diff
@@ -0,0 +1,245 @@
...
+    with_bias: bool
+    att_weights_dropout: float
+    learnable_pos_emb: bool
+    rel_pos_clip: Optional[int]
+    with_linear_pos: bool
+    with_pos_bias: bool
+    separate_pos_emb_per_head: bool
+    pos_emb_dropout: float
+    dropout: float
+    dropout_broadcast_axes: Optional[Literal["B", "T", "BT"]] = None
```
albertz, line 48: I wonder a bit about the choice where we allow defaults and where not. So you have a default for `dropout_broadcast_axes`, but not for others. Why?

i6_models/parts/conformer/mhsa_rel_pos.py
kuacakuaca: I'm not sure about the two general parameters `with_linear_pos` and `separate_pos_emb_per_head`. In my view, their values are not tied to Shaw's or Transformer-XL style relative PE.

tests/test_conformer_rel_pos.py
kuacakuaca: In the end I did find something wrong in my code by comparing.

i6_models/parts/conformer/mhsa_rel_pos.py
albertz: I don't exactly understand.

> because for `ConformerPositionwiseFeedForwardV1Config` there is already a default value set for `activation`

Why is that relevant here?

> I think dataclass attributes without default can't follow a attribute with a default.

Yes, so just reorder them?

> But I can make the config which doesn't inherit from the V1 config if you prefer no default.

You don't inherit the config here?

I'm not saying you should not have any defaults. Actually the opposite. I like to have reasonable defaults. But as I understood, we don't want to have any defaults in i6_models? Or did we already weaken this rule? So what is really the rule now? I just don't understand that you have a default for `dropout_broadcast_axes` but not for the others.

tests/test_conformer_rel_pos.py
albertz: Yea, that's what I mean, you always find differences. It's always good to do such exact test.

i6_models/parts/conformer/convolution.py
```diff
@@ -85,3 +91,62 @@ def forward(self, tensor: torch.Tensor) -> torch.Tensor:
...
+        dropout_broadcast_axes: string of axes to which dropout is broadcast, e.g. "T" for broadcasting to the time axis
+                                setting to None to disable broadcasting
+    Allows even kernel size
+    """
+
+    dropout_broadcast_axes: Optional[Literal["B", "T", "BT"]] = None
+
+    def check_valid(self):
+        assert self.kernel_size % 2 == 1, "ConformerConvolutionV1 only supports odd kernel sizes"
+
```
albertz, line 109: Above you say it supports even kernel sizes?
```suggestion
```

i6_models/parts/conformer/convolution.py
```diff
@@ -85,3 +91,62 @@ def forward(self, tensor: torch.Tensor) -> torch.Tensor:
...
         return self.dropout(tensor)
+
+
+@dataclass
+class ConformerConvolutionV2Config(ConformerConvolutionV1Config):
+    """
+    New attribute:
+        dropout_broadcast_axes: string of axes to which dropout is broadcast, e.g. "T" for broadcasting to the time axis
+                                setting to None to disable broadcasting
+    Allows even kernel size
```
albertz, line 102: I don't understand this comment. You mean `ConformerConvolutionV2` can handle even kernel sizes? But how? I don't see that the code is different from earlier? Why is it supported now?

tests/test_conformer_rel_pos.py
```diff
@@ -0,0 +1,205 @@
...
+    )
+    own_mhsa_module.qkv_proj.bias = nn.Parameter(
+        torch.cat(
+            [espnet_mhsa_module.linear_q.bias, espnet_mhsa_module.linear_k.bias, espnet_mhsa_module.linear_v.bias],
+            dim=0,
+        )
+    )
+
+    input_tensor = torch.rand((batch_dim_size, time_dim_size, embed_size))
+    sequence_mask = torch.ones((batch_dim_size, time_dim_size))
```
albertz, line 189: Can you make a mask which is not just 1 everywhere?

i6_models/parts/conformer/mhsa_rel_pos.py
kuacakuaca: I don't know how the rules are like.

i6_models/parts/conformer/convolution.py
kuacakuaca: sorry, I thought it would support but not really and I forgot to revert the docstring.

i6_models/parts/conformer/mhsa_rel_pos.py
michelwi: from https://github.com/rwth-i6/i6_models/blob/main/CONTRIBUTING.md

```
default values for Part/Assembly config:
* defaults for common hyper-parameters (e.g. number of layers): no
* defaults for rarely/never tuned hyper-parameters (e.g. optimizer constants): maybe
* defaults for structural elements (parts / assemblies / PyTorch modules/functions): yes
* if a PR contains a default (not covered by the third point) there should be a reason given for it in the PR / code
```

i6_models/parts/conformer/convolution.py
michelwi: Support for even kernel sizes can be added by manually padding the input tensors correctly. We have a helper function added in https://github.com/rwth-i6/i6_models/pull/54.

kuacakuaca: [torch.nn.Conv1d](https://pytorch.org/docs/stable/generated/torch.nn.Conv1d.html) supports only one single value for padding both ends. but as you said, we could pad the input by ourselves. Just support for even kernel size is not critical.

i6_models/parts/conformer/mhsa_rel_pos.py
albertz: So, what about the `dropout_broadcast_axes` then? We don't know really yet what variant is the best. Or do we? So no default?

None of the settings here are really structural. (Or maybe the learned pos enc is?)

We also don't really know what's best. Probably people will rarely tune it and just use the common hyper-parameters, so it's a no/maybe.

tests/test_conformer_rel_pos.py
```diff
@@ -0,0 +1,206 @@
...
+            learnable_pos_emb=learnable_pos_emb,
+            with_pos_bias=with_pos_bias,
+            pos_emb_dropout=pos_emb_dropout,
+            with_linear_pos=with_linear_pos,
+            separate_pos_emb_per_head=separate_pos_emb_per_head,
+        ) == [4, 15, 32]
+
+
+def test_ConformerMHSARelPosV1_against_Espnet():
+    from espnet2.asr_transducer.encoder.modules.attention import RelPositionMultiHeadedAttention
```
albertz, line 135:
```suggestion
        from espnet.nets.pytorch_backend.transformer.attention import RelPositionMultiHeadedAttention

```

tests/test_conformer_rel_pos.py
```diff
@@ -0,0 +1,206 @@
...
+    from espnet2.asr_transducer.encoder.modules.positional_encoding import RelPositionalEncoding
```
albertz, line 136:
```suggestion
    from espnet.nets.pytorch_backend.transformer.embedding import RelPositionalEncoding
```

tests/test_conformer_rel_pos.py
kuacakuaca: these two modules are unfortunately not identical, e.g. in the rel_shift function and the mask conversion logic.

albertz: So you say the test does not work then?

albertz:
Regarding typeguard and espnet, I actually reported that here: https://github.com/espnet/espnet/issues/5660
And this was supposed to be fixed since April 2024.
I think there is no release since then, though. But I think you can specify a git revision? Sth like `git+https://github.com/espnet/espnet.git@9d607c1e7af91c2f611f7dd181ad091649f215c4`.

requirements_dev.txt
```diff
@@ -1,2 +1,4 @@
 onnx
-onnxruntime
\ No newline at end of file
+onnxruntime
+espnet >= 202402
```
albertz, line 3:
```suggestion
git+https://github.com/espnet/espnet.git@9d607c1e7af91c2f611f7dd181ad091649f215c4
```
?

kuacakuaca: it worked, thanks!

albertz (approved):
Now it all looks fine for me. Note that I didn't verified all the formulas in detail, but I trust that the test you have is working correctly.

PR #56, opened 2024-06-27
Add diphone factored hybrid model output block

NeoLegends:
No triphone yet as there isn't an ONNX-compatible factored hybrid feature scorer yet and the diphone is compatible w/ `PrecomputedHybridFeatureScorer`. Support can be trivially amended though.

### DIFF
--- a/i6_models/parts/factored_hybrid/diphone.py
+++ b/i6_models/parts/factored_hybrid/diphone.py
@@ -24,7 +24,7 @@ class FactoredDiphoneBlockV1Config(ModelConfiguration):
         dropout: dropout probabilty
         left_context_embedding_dim: embedding dimension of the left context
             values. Good choice is in the order of n_contexts.
-        n_contexts: the number of raw phonemes/acoustic contexts
+        num_contexts: the number of raw phonemes/acoustic contexts
         num_hmm_states_per_phone: the number of HMM states per phoneme
         num_inputs: input dimension of the output block, must match w/ output dimension
             of main encoder (e.g. Conformer)
@@ -33,7 +33,7 @@ class FactoredDiphoneBlockV1Config(ModelConfiguration):
     """
 
     left_context_embedding_dim: int
-    n_contexts: int
+    num_contexts: int
     num_hmm_states_per_phone: int
     boundary_class: Union[int, BoundaryClassV1]
 
@@ -48,7 +48,7 @@ class FactoredDiphoneBlockV1Config(ModelConfiguration):
 
         assert self.context_mix_mlp_dim > 0
         assert self.context_mix_mlp_num_layers > 0
-        assert self.n_contexts > 0
+        assert self.num_contexts > 0
         assert self.num_hmm_states_per_phone > 0
         assert self.num_inputs > 0
         assert self.left_context_embedding_dim > 0
@@ -66,20 +66,20 @@ class FactoredDiphoneBlockV1(nn.Module):
     def __init__(self, cfg: FactoredDiphoneBlockV1Config):
         super().__init__()
 
-        self.n_contexts = cfg.n_contexts
+        self.n_contexts = cfg.num_contexts
 
         self.left_context_encoder = get_mlp(
             num_input=cfg.num_inputs,
-            num_output=cfg.n_contexts,
+            num_output=cfg.num_contexts,
             hidden_dim=cfg.context_mix_mlp_dim,
             num_layers=cfg.context_mix_mlp_num_layers,
             dropout=cfg.dropout,
             activation=cfg.activation,
         )
-        self.left_context_embedding = nn.Embedding(cfg.n_contexts, cfg.left_context_embedding_dim)
+        self.left_context_embedding = nn.Embedding(cfg.num_contexts, cfg.left_context_embedding_dim)
         self.center_encoder = get_mlp(
             num_input=cfg.num_inputs + cfg.left_context_embedding_dim,
-            num_output=get_center_dim(cfg.n_contexts, cfg.num_hmm_states_per_phone, cfg.boundary_class),
+            num_output=get_center_dim(cfg.num_contexts, cfg.num_hmm_states_per_phone, cfg.boundary_class),
             hidden_dim=cfg.context_mix_mlp_dim,
             num_layers=cfg.context_mix_mlp_num_layers,
             dropout=cfg.dropout,
@@ -90,6 +90,7 @@ class FactoredDiphoneBlockV1(nn.Module):
         self,
         features: Tensor,  # B, T, F
         contexts_left: Tensor,  # B, T
+        **kwargs,
     ) -> Tuple[Tensor, Tensor, Tensor]:
         """
         :param features: Main encoder output. shape B, T, F. F=num_inputs
--- a/i6_models/parts/factored_hybrid/diphone.py
+++ b/i6_models/parts/factored_hybrid/diphone.py
@@ -51,11 +51,11 @@ class FactoredDiphoneBlockV1Config(ModelConfiguration):
 
         assert self.num_contexts > 0
         assert self.num_hmm_states_per_phone > 0
-        
+
         assert self.context_mix_mlp_dim > 0
         assert self.context_mix_mlp_num_layers > 0
         assert self.left_context_embedding_dim > 0
-        
+
         assert self.num_inputs > 0
         assert 0.0 <= self.dropout <= 1.0, "dropout must be a probability"
 
--- a/i6_models/parts/factored_hybrid/diphone.py
+++ b/i6_models/parts/factored_hybrid/diphone.py
@@ -1,12 +1,10 @@
 __all__ = [
-    "DiphoneLogitsV1",
-    "DiphoneProbsV1",
-    "DiphoneBackendV1Config",
-    "DiphoneBackendV1",
+    "FactoredDiphoneBlockV1Config",
+    "FactoredDiphoneBlockV1",
 ]
 
 from dataclasses import dataclass
-from typing import Callable, Optional, Union
+from typing import Callable, Tuple, Union
 
 import torch
 from torch import nn, Tensor
@@ -14,29 +12,11 @@ import torch.nn.functional as F
 
 from i6_models.config import ModelConfiguration
 
-from .util import PhonemeStateClassV1, get_center_dim, get_mlp
+from .util import BoundaryClassV1, get_center_dim, get_mlp
 
 
 @dataclass
-class DiphoneLogitsV1:
-    """outputs of a diphone factored hybrid model"""
-
-    embeddings_left_context: Tensor
-    """the embedded left context values"""
-
-    output_center: Tensor
-    """center output"""
-
-    output_left: Tensor
-    """left output"""
-
-
-class DiphoneProbsV1(DiphoneLogitsV1):
-    """marker class indicating that the output tensors store probabilities and not logits"""
-
-
-@dataclass
-class DiphoneBackendV1Config(ModelConfiguration):
+class FactoredDiphoneBlockV1Config(ModelConfiguration):
     """
     Attributes:
         context_mix_mlp_dim: inner dimension of the context mixing MLP layers
@@ -46,7 +26,7 @@ class DiphoneBackendV1Config(ModelConfiguration):
             values. Good choice is in the order of n_contexts.
         n_contexts: the number of raw phonemes/acoustic contexts
         num_hmm_states_per_phone: the number of HMM states per phoneme
-        num_inputs: input dimension of the backend, must match w/ output dimension
+        num_inputs: input dimension of the output block, must match w/ output dimension
             of main encoder (e.g. Conformer)
         phoneme_state_class: the phoneme state augmentation to apply
         activation: activation function to use in the context mixing MLP.
@@ -55,7 +35,7 @@ class DiphoneBackendV1Config(ModelConfiguration):
     left_context_embedding_dim: int
     n_contexts: int
     num_hmm_states_per_phone: int
-    phoneme_state_class: Union[int, PhonemeStateClassV1]
+    boundary_class: Union[int, BoundaryClassV1]
 
     activation: Callable[[], nn.Module]
     context_mix_mlp_dim: int
@@ -75,15 +55,15 @@ class DiphoneBackendV1Config(ModelConfiguration):
         assert 0.0 <= self.dropout <= 1.0, "dropout must be a probability"
 
 
-class DiphoneBackendV1(nn.Module):
+class FactoredDiphoneBlockV1(nn.Module):
     """
-    Diphone FH model backend.
+    Diphone FH model output block.
 
-    Consumes the output h(x) of a main encoder model and computes factored output
-    logits/probabilities for p(c|l,x) and p(l|x).
+    Consumes the output h(x) of a main encoder model and computes factored or joint
+    output logits/probabilities for p(c|l,x) and p(l|x).
     """
 
-    def __init__(self, cfg: DiphoneBackendV1Config):
+    def __init__(self, cfg: FactoredDiphoneBlockV1Config):
         super().__init__()
 
         self.n_contexts = cfg.n_contexts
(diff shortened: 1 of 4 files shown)


i6_models/parts/factored_hybrid/util.py
```diff
@@ -0,0 +1,63 @@
+from enum import Enum
+from typing import Callable, Union
+
+from torch import nn
+
+
+class PhonemeStateClassV1(Enum):
```
michelwi, line 7:
```suggestion
class PhonemeBoundaryClassV1(Enum):
```
"state" to me sounds like it would be related to `num_hmm_states_per_phone`

i6_models/parts/factored_hybrid/util.py
```diff
@@ -0,0 +1,63 @@
...
+) -> int:
+    """
+    :return: number of center phonemes given the augmentation values
+    """
+
+    factor = ph_class.factor() if isinstance(ph_class, PhonemeStateClassV1) else ph_class
+    return n_contexts * num_hmm_states_per_phone * factor
+
+
+def get_mlp(
```
michelwi, line 42: There is also https://github.com/rwth-i6/i6_models/blob/main/i6_models/parts/ffnn.py that one could reuse. But some linear layers seems simple enough to me that I do not care about code duplication here.

i6_models/parts/factored_hybrid/diphone.py
```diff
@@ -0,0 +1,167 @@
...
+            center_logits = self.center_encoder(center_features)  # C, B, T, F'
+            center_probs = F.log_softmax(center_logits, -1)
+            center_probs = center_probs.permute((1, 2, 3, 0))  # B, T, F', C
+            left_probs = F.log_softmax(left_logits, -1)
+            left_probs = left_probs.unsqueeze(-2)  # B, T, 1, C
+
+            joint_probs = center_probs + left_probs  # B, T, F', C
+            joint_probs = torch.flatten(joint_probs, start_dim=2)  # B, T, joint
+
+            return DiphoneProbsV1(
```
michelwi, line 163: In the full diphone case we do not need the `output_left` or the embeddings, do we?

Would it make sense to use a different Dataclass here?

i6_models/parts/factored_hybrid/diphone.py
```diff
@@ -0,0 +1,167 @@
...
+            num_layers=cfg.context_mix_mlp_num_layers,
+            dropout=cfg.dropout,
+            activation=cfg.activation,
+        )
+
+    def forward(
+        self,
+        features: Tensor,  # B, T, F
+        contexts_left: Optional[Tensor] = None,  # B, T
+    ) -> Union[DiphoneLogitsV1, DiphoneProbsV1]:
```
michelwi, line 113: I know what you semantically want to say here, but since DiphoneProbsV1 inherits from DiphoneLogitsV1, programmatically I think this would be equivalent to
```suggestion
    ) -> DiphoneLogitsV1:
```

This change directly contradicts one other comment I made below, pls apply only one of those :P

i6_models/parts/factored_hybrid/diphone.py
```diff
@@ -0,0 +1,167 @@
...
+        phoneme_state_class: the phoneme state augmentation to apply
+        activation: activation function to use in the context mixing MLP.
+    """
+
+    left_context_embedding_dim: int
+    n_contexts: int
+    num_hmm_states_per_phone: int
+    phoneme_state_class: Union[int, PhonemeStateClassV1]
+
+    activation: Callable[[], nn.Module]
```
michelwi, line 60: would that be a function that takes no input but returns a Module?

In other places I have seen the typing
```suggestion
    activation: Union[nn.Module, Callable[[torch.Tensor], torch.Tensor]]
```
which seems more appropriate to me

i6_models/parts/factored_hybrid/diphone.py
```diff
@@ -0,0 +1,167 @@
...
+            # here we forward every context to compute p(c, l|x) = p(c|l, x) * p(l|x)
+
+            features = features.expand((self.n_contexts, -1, -1, -1))  # C, B, T, F
+            embedded_left_contexts_ = embedded_left_contexts.reshape((self.n_contexts, 1, 1, -1)).expand(
+                (-1, *features.shape[1:3], -1)
+            )  # C, B, T, E
+            center_features = torch.cat((features, embedded_left_contexts_), -1)  # C, B, T, F+E
+            center_logits = self.center_encoder(center_features)  # C, B, T, F'
+            center_probs = F.log_softmax(center_logits, -1)
+            center_probs = center_probs.permute((1, 2, 3, 0))  # B, T, F', C
```
michelwi, line 156: I hope you checked that this leaves the outputs in the same order as expected in RASR diphone-dense tying?

i6_models/parts/factored_hybrid/diphone.py
```diff
@@ -0,0 +1,167 @@
...
+        assert self.context_mix_mlp_dim > 0
+        assert self.context_mix_mlp_num_layers > 0
+        assert self.n_contexts > 0
+        assert self.num_hmm_states_per_phone > 0
+        assert self.num_inputs > 0
+        assert self.left_context_embedding_dim > 0
+        assert 0.0 <= self.dropout <= 1.0, "dropout must be a probability"
+
+
+class DiphoneBackendV1(nn.Module):
```
michelwi, line 78: I am a bit unhappy about the name "backend" as that term says exactly nothing. I once talked to guys from Paderborn who were doing denoising and source separation of audio spectra and when they said "backend" they meant "feature extraction and other ASR model stuff" 🤷 

I was trying to find a good name that implies that here embeddings are converted to (log-)probabilities.. but I failed.. pls help :)

JackTemaki: +1 for Willis statement. Please really do not call it Backend. Maybe `DiphoneOutputBlock`?

i6_models/parts/factored_hybrid/diphone.py
JackTemaki: I just want to raise that "maybe" dataclasses as output might not be optimal for some graph tracing/scripting logic from PyTorch.

i6_models/parts/factored_hybrid/diphone.py
Marvin84: +1 for Moritz idea.
In defense of the naming, in the literature and commonly within ASR community, the feature extraction and all respective methods such as source separation, beam forming, and enhancement are denoted as ASR frontend and the model itself is called backend. I find the naming well chosen since it reminds the user that this can be combined with any encoder architecture or speech representation h(x).
The new generation calls the similar part of an RNN-T as the "decoder", which is certainly not a better term.

Marvin84: We just had a live discussion with @JackTemaki and @ZijYang and we think a nice name for this can be `DiphoneJoinerBlock` reminding the similar RNN-T case

i6_models/parts/factored_hybrid/diphone.py
Marvin84: We typically keep separate outputs for training and combine the two outputs only during inference, meaning that we do need a loss computation also on the `left_output`

michelwi: this is an argument in favor of having different kinds of outputs depending on train yes/no i.e. using different dataclasses (or e.g. tuples if we want to follow @JackTemaki s suggestion)

i6_models/parts/factored_hybrid/util.py
Marvin84: >  it would be related to num_hmm_states_per_phone

Yes, I also agree.

i6_models/parts/factored_hybrid/diphone.py
michelwi: offline discussion: With the fear of a statefull activation module this is indeed intended as a function that creates new instances every time it is called.

i6_models/parts/factored_hybrid/diphone.py
Marvin84: Another aspect to consider @NeoLegends  is whether we want to have two separate modules for the architecture that uses frame-wise cross-entropy and the one that uses sequence-level cross-entropy, i.e. fullsum. Since in the latter also training would rely on the combination of the outputs.

i6_models/parts/factored_hybrid/util.py
NeoLegends: That one does not work with `nn.Sequential`, which is why I don't use it here.

i6_models/parts/factored_hybrid/diphone.py
michelwi: Maybe a 
```
Tuple[Optional[Tensor], Optional[Tensor], Optional[Tensor], Optional[Tensor]]
```
with [joined, center, left, left_embedding] and depending on the conditions, each can be None if not applicable.?

Or should we just always return all of the above and then in training / inference some are just not used?

i6_models/parts/factored_hybrid/util.py
michelwi: offline discussion: this does not work for nn.Sequential

i6_models/parts/factored_hybrid/diphone.py
NeoLegends: I'm moving to tuples.

i6_models/parts/factored_hybrid/diphone.py
NeoLegends: I'm going to verify what I'm doing here is right before this PR can be merged. Either via test or via setup.

i6_models/parts/factored_hybrid/diphone.py
michelwi: I like "Joiner" better than "Backend" and "Output" most of them. Although this is not necessarily the (final) output of the model.

`DiphoneProbabilityBlock` ? or even just (`Factored`-)`DiphoneBlock` ?

i6_models/parts/factored_hybrid/diphone.py
michelwi: > sequence-level cross-entropy, i.e. fullsum.

*"e.g. fullsum"

Daniel has already offered to do a pr for the LFMMI op xP

i6_models/parts/factored_hybrid/diphone.py
Marvin84: I also like `FactoredDiphoneBlock`. 
BTW, one feedback I got from people outside our research group is that there is a marketing problem with the name of factored hybrid, and if we had a fancy new name people would be more open to try it out. Independently of the name, using up-to-date terminologies that do not remind the GMM-HMM systems is preferable, IMO.

i6_models/parts/factored_hybrid/diphone.py
Marvin84: > how easy it is to change model classes from training to inference in torch

Let's also ask @SimBe195 and @Judyxujj, they are also running torch trainings for a while now

i6_models/parts/factored_hybrid/diphone.py
michelwi: > if we had a fancy new name people would be more open to try it out. 

_seigh_

> Independently of the name, using up-to-date terminologies that do not remind the GMM-HMM systems is preferable, IMO.

Nothing inherently needs this to be for "phone"s it could also be FactoredDichar or DiBPE.. (except for the adherence to rasrs phoneme numbering scheme in `PhonemeStateClassV1`).. 

So should we drop the "Phone" branding?^^

i6_models/parts/factored_hybrid/diphone.py
JackTemaki: I would not do it in separate modules, so that you always load everything consistently, but you can also not use `self.training`. What you can have is separate functions. `self.training` has the problem that it is really for "training" only, not e.g. for cross validation during training. We had bugs in setups before where we then applied dropout and specaugment for the validation sets.

So when you use `train_step` or `forward_step / onnx_export` you can call two different functions

i6_models/parts/factored_hybrid/diphone.py
```diff
@@ -0,0 +1,138 @@
...
+        :return: tuple of logits for p(c|l,x), p(l|x) and the embedded left context values.
+        """
+
+        left_logits = self.left_context_encoder(features)  # B, T, C
+        # in training we forward exactly one context per T, so: B, T, E
+        contexts_embedded_left = self.left_context_embedding(contexts_left)
+        features_center = torch.cat((features, contexts_embedded_left), -1)  # B, T, F+E
+        logits_center = self.center_encoder(features_center)  # B, T, C
+
+        return logits_center, left_logits, contexts_embedded_left
```
michelwi, line 110: `logits_center` vs `left_logits` 
consistent order of "position" and "what is it"  would be appreciated

NeoLegends: Ah, yes.

i6_models/parts/factored_hybrid/diphone.py
NeoLegends: I have decoding results now with reasonable WERs and these verify that what we're doing here is right. :)

i6_models/parts/factored_hybrid/diphone.py
```diff
@@ -0,0 +1,145 @@
...
+        context_mix_mlp_dim: inner dimension of the context mixing MLP layers
+        context_mix_mlp_num_layers: how many hidden layers on the MLPs there should be
+        dropout: dropout probabilty
+        left_context_embedding_dim: embedding dimension of the left context
+            values. Good choice is in the order of num_contexts.
+        num_contexts: the number of raw phonemes/acoustic contexts
+        num_hmm_states_per_phone: the number of HMM states per phoneme
+        num_inputs: input dimension of the output block, must match w/ output dimension
+            of main encoder (e.g. Conformer)
+        phoneme_state_class: the phoneme state augmentation to apply
```
michelwi, line 31:
```suggestion
        boundary_class: the phoneme state augmentation to apply
```

also make a conscious decision on the order of parameters and then order the docstring in the same order as the parameters defined pls.

i6_models/parts/factored_hybrid/diphone.py
```diff
@@ -0,0 +1,145 @@
...
+            num_layers=cfg.context_mix_mlp_num_layers,
+            dropout=cfg.dropout,
+            activation=cfg.activation,
+        )
+        self.left_context_embedding = nn.Embedding(cfg.num_contexts, cfg.left_context_embedding_dim)
+        self.center_encoder = get_mlp(
+            num_input=cfg.num_inputs + cfg.left_context_embedding_dim,
+            num_output=self.num_center,
+            hidden_dim=cfg.context_mix_mlp_dim,
+            num_layers=cfg.context_mix_mlp_num_layers,
```
michelwi, line 89: Is this intentionally using the same dim/num_layers as the (left-)context MLP? Maybe rename those parameters to a name that reflects that both left-context and center have these dims?

NeoLegends: My intuition says the context mixing MLPs themselves don't have a big impact on model performance as long as they're sized sufficiently to perform their context mixing job, but maybe for the higher order contexts the size becomes more relevant due to the additional dependencies. Then again, this is also just intuition.

@Marvin84 Do you have any experience about the difference the sizing of these layers makes on the overall WER? The dim and number of layers here are the same because (I think) we never really experimented around much with them.

NeoLegends: It would certainly help if it turned out that these layers could even be sized down, and thereby speed up the score computation vs. a CART model (joint FH diphone is slower than triphone CART due to batching the context mixing MLPs over all contexts, whereas the CART just upsizes the model output via one linear layer to the final output dim).

Marvin84: I experimented with BLSTM based models, and in that case one layer against two layers had max `0.2%` absolute WER degradation on Hub500 (trained on SWB300h). Moreover, lower embed dimension could lead to absolute `0.1%` WER degradation. 
However, I noticed that for from-scratch Conformed based alignment models (monophone PHMM) having or not having MLP layers matter a lot. I have not checked the effect of number of MLP layers for Viterbi training with fixed path, though.

> speed up the score computation vs. a CART model 

How much speedup do you expect to get?

NeoLegends: image

Upper line is triphone CART, lower line diphone FH. Would be nice if the TF cost for both was at least equal. It's unlikely we can get FH any better, though.

michelwi: > Upper line is triphone CART, lower line diphone FH.

should we not compare explicit full-diphone to joint forwarding of factored diphone?

NeoLegends: In any case I think making this any more configurable can be postponed to a V2.

> should we not compare explicit full-diphone to joint forwarding of factored diphone?

Technically, yes. There won't be a noticeable difference to a triphone CART model wrt. the NN cost, however, as the size of the output layers is almost equal (at least for Librispeech, where it's 12k CART vs 11.5k diphone).

michelwi (approved):

Marvin84 (approved):

PR #58, opened 2024-08-22
Add FactoredTriphoneBlockV1

NeoLegends:
Based on #57, to be merged afterwards.

### DIFF
--- a/i6_models/parts/factored_hybrid/triphone.py
+++ b/i6_models/parts/factored_hybrid/triphone.py
@@ -1,0 +1,69 @@
+__all__ = [
+    "FactoredTriphoneBlockV1Config",
+    "FactoredTriphoneBlockV1",
+]
+
+from dataclasses import dataclass
+from typing import Tuple
+
+import torch
+from torch import nn, Tensor
+
+from .diphone import FactoredDiphoneBlockV1, FactoredDiphoneBlockV2Config
+from .util import get_mlp
+
+
+@dataclass
+class FactoredTriphoneBlockV1Config(FactoredDiphoneBlockV2Config):
+    """
+    Attributes:
+        Same as the FactoredDiphoneBlockV2Config.
+    """
+
+
+class FactoredTriphoneBlockV1(FactoredDiphoneBlockV1):
+    """
+    Triphone FH model output block.
+
+    Consumes the output h(x) of a main encoder model and computes factored logits/probabilities
+    for p(c|l,h(x)), p(l|h(x)) and p(r|c,l,h(x)).
+    """
+
+    def __init__(self, cfg: FactoredTriphoneBlockV1Config):
+        super().__init__(cfg)
+
+        self.center_state_embedding = nn.Embedding(self.num_center, cfg.center_state_embedding_dim)
+        self.right_context_encoder = get_mlp(
+            num_input=cfg.num_inputs + cfg.center_state_embedding_dim + cfg.left_context_embedding_dim,
+            num_output=self.num_contexts,
+            hidden_dim=cfg.context_mix_mlp_dim,
+            num_layers=cfg.context_mix_mlp_num_layers,
+            dropout=cfg.dropout,
+            activation=cfg.activation,
+        )
+
+    def forward(
+        self,
+        features: Tensor,  # B, T, F
+        contexts_left: Tensor,  # B, T
+        contexts_center: Tensor,  # B, T
+    ) -> Tuple[Tensor, Tensor, Tensor, Tensor, Tensor]:
+        """
+        :param features: Main encoder output. shape B, T, F. F=num_inputs
+        :param contexts_left: The left contexts used to compute p(c|l,x), shape B, T.
+            Valid values range from [0, num_contexts).
+        :param contexts_center: The center states used to compute p(r|c,l,x), shape B, T.
+            Given that the center state also contains the word-end class and HMM state ID, the valid values
+            range from [0, num_center_states), where num_center_states >= num_contexts.
+        :return: tuple of logits for p(c|l,x), p(l|x), p(r|c,l,x) and the embedded left context and center state values.
+        """
+
+        logits_center, logits_left, contexts_left_embedded = super().forward(features, contexts_left)
+
+        # This logic is very similar to FactoredDiphoneBlockV2.forward, but not the same.
+        # This class computes `p(r|c,l,h(x))` while FactoredDiphoneBlockV2 computes `p(r|c,h(x))`.
+        center_states_embedded = self.center_state_embedding(contexts_center)  # B, T, E'
+        features_right = torch.cat((features, center_states_embedded, contexts_left_embedded), -1)  # B, T, F+E'+E
+        logits_right = self.right_context_encoder(features_right)  # B, T, C
+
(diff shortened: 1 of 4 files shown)


i6_models/parts/factored_hybrid/triphone.py
```diff
@@ -0,0 +1,69 @@
...
+from .diphone import FactoredDiphoneBlockV1, FactoredDiphoneBlockV2Config
+from .util import get_mlp
+
+
+@dataclass
+class FactoredTriphoneBlockV1Config(FactoredDiphoneBlockV2Config):
+    """
+    Attributes:
+        Same as the FactoredDiphoneBlockV2Config.
+    """
```
michelwi, line 21: would `from .diphone import FactoredDiphoneBlockV2Config as FactoredTriphoneBlockV1Config` also work?

i6_models/parts/factored_hybrid/triphone.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class FactoredTriphoneBlockV1(FactoredDiphoneBlockV1):
+    """
+    Triphone FH model output block.
+
+    Consumes the output h(x) of a main encoder model and computes factored logits/probabilities
+    for p(c|l,h(x)), p(l|h(x)) and p(r|c,l,h(x)).
```
michelwi, line 29:
```suggestion
    for p(r|c,l,h(x)), p(c|l,h(x)), and p(l|h(x)).
```
my ocd proposes to reorder to match order of decomposition.

i6_models/parts/factored_hybrid/triphone.py
```diff
@@ -0,0 +1,69 @@
...
+
+        logits_center, logits_left, contexts_left_embedded = super().forward(features, contexts_left)
+
+        # This logic is very similar to FactoredDiphoneBlockV2.forward, but not the same.
+        # This class computes `p(r|c,l,h(x))` while FactoredDiphoneBlockV2 computes `p(r|c,h(x))`.
+        center_states_embedded = self.center_state_embedding(contexts_center)  # B, T, E'
+        features_right = torch.cat((features, center_states_embedded, contexts_left_embedded), -1)  # B, T, F+E'+E
+        logits_right = self.right_context_encoder(features_right)  # B, T, C
+
+        return logits_center, logits_left, logits_right, contexts_left_embedded, center_states_embedded
```
michelwi, line 69: Ok, now I see this has the same order as the docstring. I do agree that the docstring order should match the returned order.

I would also propose to return in order of decomposition, but then the order would be different compared to DiphoneV2 which is also undesirable. So our options would be
1. deviate from the natural order of decomposition
2. deviate from the order of DiphoneV2 (source of error/confusion)
3. change the order also in #57 (but then the order might be less natural for diphones)

I don't have a clear preference any more..

i6_models/parts/factored_hybrid/triphone.py
```diff
@@ -0,0 +1,69 @@
...
+        contexts_center: Tensor,  # B, T
+    ) -> Tuple[Tensor, Tensor, Tensor, Tensor, Tensor]:
+        """
+        :param features: Main encoder output. shape B, T, F. F=num_inputs
+        :param contexts_left: The left contexts used to compute p(c|l,x), shape B, T.
+            Valid values range from [0, num_contexts).
+        :param contexts_center: The center states used to compute p(r|c,l,x), shape B, T.
+            Given that the center state also contains the word-end class and HMM state ID, the valid values
+            range from [0, num_center_states), where num_center_states >= num_contexts.
+        :return: tuple of logits for p(c|l,x), p(l|x), p(r|c,l,x) and the embedded left context and center state values.
```
michelwi, line 58:
```suggestion
        :return: tuple of logits for p(r|c,l,x), p(c|l,x), p(l|x) and the embedded left context and center state values.
```
same as above

i6_models/parts/factored_hybrid/triphone.py
NeoLegends: The advantage of this order is that you don't need to do much work in the `train_step` function to distinguish a monophone from a diphone from a triphone model, and you can even train different models w/ the same `train_step` function.

You really just need to look at the length of the returned tuple to determine on which output tensors to compute the losses. Since the tensors don't change "position" in the tuple, that amounts to only a few lines of code, and more if they would change position based on e.g. the factorization order. :)

i6_models/parts/factored_hybrid/triphone.py
NeoLegends: It would, but I kind of would like this to be a real type instead of only an alias, in case we ever need to add more properties? Maybe it shouldn't inherit from the other class at all then -- but that would also reduce DRY-ness. 🤔

i6_models/parts/factored_hybrid/triphone.py
NeoLegends: The way it is now is kind of an "order by model complexity" -- you always have the (monophone) center state, so it comes first. Then you increase the model complexity to a diphone, and you get another output. Increase once more to triphone -> one more output.

michelwi: ok, I accept this argument.

i6_models/parts/factored_hybrid/triphone.py
```diff
@@ -0,0 +1,69 @@
...
+        self.right_context_encoder = get_mlp(
+            num_input=cfg.num_inputs + cfg.center_state_embedding_dim + cfg.left_context_embedding_dim,
+            num_output=self.num_contexts,
+            hidden_dim=cfg.context_mix_mlp_dim,
+            num_layers=cfg.context_mix_mlp_num_layers,
+            dropout=cfg.dropout,
+            activation=cfg.activation,
+        )
+
+    def forward(
```
michelwi, line 45: Here we overload the `forward` method. Which means that still the `FactoredDiphoneBlockV1.forward_factored` method exists in this class. Should we maybe overload `forward_factored` instead and then the `forward` inherited from `FactoredDiphoneBlockV1` would automatically do the correct thing.. except for typing...  so.. maybe inherit from V2 instead?

Related question: `forward_joint` is also still active, but only forwards the full diphone output ignoring the right context. Should we disable that? should we warn users? should we implement full triphone just in case anyone was so bold to ever try it?

NeoLegends: Good point. I don't think we should inherit from V2, because then we have some duplicate work done when calling `super().forward_factored(...)`, but I should definitely override `forward_factored` instead, and then also update the typings for `forward()`, as in Diphone V2.

NeoLegends: Wrt. `forward_joint`, a 10ms triphone LS960 model has about 450k outputs. I don't think this is ever going to be feasible, in whatever combination, w/o some clever optimizations. I think we should leave it at the diphone level for now, as this also allows decoding the triphone-trained models. It's a bit unfortunate that the `forward_joint` name does not include the term "diphone" anywhere, so that this would be clearer.

michelwi (approved):

curufinwe (approved):

NeoLegends:
🎉

PR #59, opened 2024-08-23
RASR FSA Builder

DanEnergetics:
Components for building FSAs with the `librasr` python package and using them in a PyTorch training

### DIFF
--- a/i6_models/parts/fsa.py
+++ b/i6_models/parts/fsa.py
@@ -1,11 +1,13 @@
+from __future__ import annotations
+
 __all__ = ["TorchFsaBuilder", "WeightedFsa"]
 
-from __future__ import annotations
 from functools import reduce
 from typing import Iterable, NamedTuple, Tuple, TypeVar
 
 import numpy as np
 import torch
+import librasr
 
 
 class WeightedFsa(NamedTuple):
@@ -13,6 +15,7 @@ class WeightedFsa(NamedTuple):
     Convenience class that represents an FSA. It supports scaling the weights of the
     fsa by simple left-multiplication and moving the tensors to a different device.
     It can simply be passed to :func:`i6_native_ops.fbw.fbw_loss` and :func:`i6_native_ops.fast_viterbi.align_viterbi`.
+
     :param num_states: the total number of all states S
     :param edges: a [4, E] tensor of edges with number of edges E and where each column is an edge
         consisting of from-state, to-state, emission idx and the index of the sequence it belongs to
@@ -34,7 +37,7 @@ class WeightedFsa(NamedTuple):
             self.start_end_states,
         )
 
-    def to(self, device: str) -> WeightedFsa:
+    def to(self, device: torch.device) -> WeightedFsa:
         """Move the tensors to a given device. This wraps around the
         PyTorch `Tensor.to(device)` method."""
         return WeightedFsa._make(tensor.to(device) for tensor in self)
@@ -48,13 +51,12 @@ class TorchFsaBuilder:
     This class provides an explicit implementation of the `__getstate__` and `__setstate__`
     functions, necessary for pickling as the C++-class `librasr.AllophoneStateFsaBuilder`
     is not picklable.
+
     :param config_path: path to the RASR fsa exporter config
     :param tdp_scale: multiply the weights by this scale
     """
 
     def __init__(self, config_path: str, tdp_scale: float = 1.0):
-        import librasr
-
         self.config_path = config_path
         config = librasr.Configuration()
         config.set_from_file(self.config_path)
@@ -75,7 +77,7 @@ class TorchFsaBuilder:
     def build_single(self, seq_tag: str) -> Tuple[int, int, np.ndarray, np.ndarray]:
         """
         Build the FSA for the given sequence tag in the corpus.
-        
+
         :param seq_tag: sequence tag
         :return: FSA as a tuple containing
             * number of states S
@@ -96,7 +98,7 @@ class TorchFsaBuilder:
         the state IDs of each single FSA are incrememented and made unique in
         the batch.
         Additionally we apply an optional scale to the weights.
-        
+
         :param seq_tags: an iterable object of sequence tags
         :return: a concatenated FSA
         """
--- a/i6_models/parts/fsa.py
+++ b/i6_models/parts/fsa.py
@@ -83,7 +83,7 @@ class TorchFsaBuilder:
         raw_fsa = self.builder.build_by_segment_name(seq_tag)
         return raw_fsa
 
-    def build_batch(self, seq_tags: Iterable[str]) -> TWeightedFsa:
+    def build_batch(self, seq_tags: Iterable[str]) -> WeightedFsa:
         """
         Build and concatenate the FSAs for a batch of sequence tags
         and reformat as an input to `i6_native_ops.fbw.fbw_loss`.
--- a/i6_models/parts/fsa.py
+++ b/i6_models/parts/fsa.py
@@ -8,6 +8,7 @@ import torch
 
 TWeightedFsa = TypeVar("TWeightedFsa", bound="WeightedFsa")
 
+
 class WeightedFsa(NamedTuple):
     """
     Convenience class that represents an FSA. It supports scaling the weights of the
@@ -20,6 +21,7 @@ class WeightedFsa(NamedTuple):
     :param weights: a [E,] tensor of weights for each edge scaled by the tdp_scale
     :param start_end_states: a [N, 2] tensor of start and end states for each of the N sequences
     """
+
     num_states: torch.IntTensor
     edges: torch.IntTensor
     weights: torch.FloatTensor
@@ -27,11 +29,8 @@ class WeightedFsa(NamedTuple):
 
     def __mul__(self: TWeightedFsa, scale: float) -> TWeightedFsa:
         """Multiply the weights, i.e. the third element, with a scale."""
-        return WeightedFsa._make(
-            tensor * scale if i == 2 else tensor
-            for i, tensor in enumerate(self)
-        )
-    
+        return WeightedFsa._make(tensor * scale if i == 2 else tensor for i, tensor in enumerate(self))
+
     def to(self: TWeightedFsa, device: str) -> TWeightedFsa:
         """Move the tensors to a given device. This wraps around the
         PyTorch `Tensor.to(device)` method."""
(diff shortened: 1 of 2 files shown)


DanEnergetics:
I'm not sure about a test for this. It does require the librasr package, which I'm not sure if we should put it into the requirements and to initialize the `AllophoneStateFsaBuilder` we would need a RASR config file pointing to a valid lexicon afaik

JackTemaki:
You "could" add extra dependencies just for a test, including a small dummy lexicon and corpus for testing. While this would be preferable to have, I am not sure if this is within your time-scope.

DanEnergetics:
I would like to do it, actually. Having a small dummy corpus and lexicon is a good idea. However, the installation of `librasr` is not that simple and I'm not sure how we could make it work with the automatic `model_tests`.

i6_models/parts/fsa.py
```diff
@@ -0,0 +1,137 @@
...
+            * number of states S
+            * number of edges E
+            * integer edge array of shape [E, 3] where each row is an edge
+                consisting of from-state, to-state and the emission idx
+            * float weight array of shape [E,]
+        """
+        raw_fsa = self.builder.build_by_segment_name(seq_tag)
+        return raw_fsa
+
+    def build_batch(self, seq_tags: Iterable[str]) -> TWeightedFsa:
```
michelwi, line 86:
```suggestion
    def build_batch(self, seq_tags: Iterable[str]) -> WeightedFsa:
```
I think here not the TypeVar but the concrete Type that will be returned should be used.

i6_models/parts/fsa.py
```diff
@@ -0,0 +1,137 @@
...
+TWeightedFsa = TypeVar("TWeightedFsa", bound="WeightedFsa")
+
+
+class WeightedFsa(NamedTuple):
+    """
+    Convenience class that represents an FSA. It supports scaling the weights of the
+    fsa by simple left-multiplication and moving the tensors to a different device.
+    It can simply be passed to `i6_native_ops.fbw.fbw_loss` and `i6_native_ops.fast_viterbi.align_viterbi`.
+    :param num_states: the total number of all states S
+    :param edges: a [4, E] tensor of edges where each column is an edge consisting
```
michelwi, line 18: Nitpick: `E` (I guess number of edges) is undefined here.

michelwi (commented):
Excellent documentation, thank you!

i6_models/parts/fsa.py
```diff
@@ -0,0 +1,136 @@
...
+        self.tdp_scale = tdp_scale
+
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        del state["builder"]
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        config = librasr.Configuration()
```
JackTemaki, line 67: Does this actually work? Even with `librasr` only imported during init? I am surprised this is still the same scope.

i6_models/parts/fsa.py
```diff
@@ -0,0 +1,136 @@
...
+
+TWeightedFsa = TypeVar("TWeightedFsa", bound="WeightedFsa")
+
+
+class WeightedFsa(NamedTuple):
+    """
+    Convenience class that represents an FSA. It supports scaling the weights of the
+    fsa by simple left-multiplication and moving the tensors to a different device.
+    It can simply be passed to `i6_native_ops.fbw.fbw_loss` and `i6_native_ops.fast_viterbi.align_viterbi`.
+    :param num_states: the total number of all states S
```
albertz, line 17:
```suggestion
    It can simply be passed to `i6_native_ops.fbw.fbw_loss` and `i6_native_ops.fast_viterbi.align_viterbi`.

    :param num_states: the total number of all states S
```
There should be an empty line between the description and the params in the docstring.

i6_models/parts/fsa.py
```diff
@@ -0,0 +1,136 @@
...
+import torch
+
+TWeightedFsa = TypeVar("TWeightedFsa", bound="WeightedFsa")
+
+
+class WeightedFsa(NamedTuple):
+    """
+    Convenience class that represents an FSA. It supports scaling the weights of the
+    fsa by simple left-multiplication and moving the tensors to a different device.
+    It can simply be passed to `i6_native_ops.fbw.fbw_loss` and `i6_native_ops.fast_viterbi.align_viterbi`.
```
albertz, line 16:
```suggestion
    It can simply be passed to :func:`i6_native_ops.fbw.fbw_loss` and :func:`i6_native_ops.fast_viterbi.align_viterbi`.
```
Or otherwise use two backsticks (``) (Rest syntax).

i6_models/parts/fsa.py
```diff
@@ -0,0 +1,136 @@
+__all__ = ["TorchFsaBuilder", "WeightedFsa"]
+
+from functools import reduce
+from typing import Iterable, NamedTuple, Tuple, TypeVar
+
+import numpy as np
+import torch
+
+TWeightedFsa = TypeVar("TWeightedFsa", bound="WeightedFsa")
+
```
albertz, line 10: I don't think you need this.
```suggestion
```

i6_models/parts/fsa.py
```diff
@@ -0,0 +1,136 @@
...
+    :param weights: a [E,] tensor of weights for each edge scaled by the tdp_scale
+    :param start_end_states: a [N, 2] tensor of start and end states for each of the N sequences
+    """
+
+    num_states: torch.IntTensor
+    edges: torch.IntTensor
+    weights: torch.FloatTensor
+    start_end_states: torch.IntTensor
+
+    def __mul__(self: TWeightedFsa, scale: float) -> TWeightedFsa:
```
albertz, line 29:
```suggestion
    def __mul__(self: TWeightedFsa, scale: float) -> WeightedFsa:
```

i6_models/parts/fsa.py
```diff
@@ -0,0 +1,136 @@
...
+        """Multiply the weights, i.e. the third element, with a scale."""
+        return WeightedFsa._make(tensor * scale if i == 2 else tensor for i, tensor in enumerate(self))
+
+    def to(self: TWeightedFsa, device: str) -> TWeightedFsa:
```
albertz, line 33:
```suggestion
    def to(self: TWeightedFsa, device: str) -> WeightedFsa:
```

i6_models/parts/fsa.py
```diff
@@ -0,0 +1,136 @@
+__all__ = ["TorchFsaBuilder", "WeightedFsa"]
+
```
albertz, line 2:
```suggestion

from __future__ import annotations
```

i6_models/parts/fsa.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+
+    num_states: torch.IntTensor
+    edges: torch.IntTensor
+    weights: torch.FloatTensor
+    start_end_states: torch.IntTensor
+
+    def __mul__(self: TWeightedFsa, scale: float) -> TWeightedFsa:
+        """Multiply the weights, i.e. the third element, with a scale."""
+        return WeightedFsa._make(tensor * scale if i == 2 else tensor for i, tensor in enumerate(self))
```
albertz, line 31: I think having this explicit is better:
```suggestion
        return WeightedFsa(
            self.num_states,
            self.edges,
            self.weights * scale,
            self.start_end_states,
        )
```

i6_models/parts/fsa.py
albertz, line 33: Is `device: str` correct? I think it must be `torch.device`. I.e.:
```suggestion
    def to(self, device: torch.device) -> WeightedFsa:
```

i6_models/parts/fsa.py
```diff
@@ -0,0 +1,136 @@
...
+    def build_batch(self, seq_tags: Iterable[str]) -> WeightedFsa:
+        """
+        Build and concatenate the FSAs for a batch of sequence tags
+        and reformat as an input to `i6_native_ops.fbw.fbw_loss`.
+        Here the FSAs are concatenated to a long FSA with multiple start and
+        end states corresponding to each single FSA. For the concatenation,
+        the state IDs of each single FSA are incrememented and made unique in
+        the batch.
+        Additionally we apply an optional scale to the weights.
+        :param seq_tags: an iterable object of sequence tags
```
albertz, line 94:
```suggestion
        Additionally we apply an optional scale to the weights.
        
        :param seq_tags: an iterable object of sequence tags
```

i6_models/parts/fsa.py
DanEnergetics: Yes that is indeed what it says in the PyTorch source code. The docs say that properly formatted strings would be accepted as well but we can decide on the `torch.device` here

albertz (approved):

i6_models/parts/fsa.py
michelwi: > If nothing speaks against it I'd just go with a normal import.

It would be unfortunate if every user of i6_models needed to have a functioning `librasr` install. 

But since nothing from `i6_models.parts.fsa` is imported into `i6_models.parts` I think we are good and only if one really uses the `TorchFsaBuilder`, one needs `librasr`

i6_models/parts/fsa.py
```diff
@@ -0,0 +1,144 @@
...
+            self.start_end_states,
+        )
+
+    def to(self, device: Union[str, torch.device]) -> WeightedFsa:
+        """Move the tensors to a given device. This wraps around the
+        PyTorch `Tensor.to(device)` method."""
+        return WeightedFsa._make(tensor.to(device) for tensor in self)
+
+
+class TorchFsaBuilder:
```
michelwi, line 46: I think everything in i6_models is torch related. The specialty of this class is that it uses rasr as a source for the fsas.
Would then 
```suggestion
class RasrFsaBuilder:
```
be a more appropriate branding for this class here?

Also, would it make sense to rename `fsa.py` to `rasr_fsa.py` just in case we will have different fsa related functions that should not be in the same scope as the `librasr` import?

i6_models/parts/fsa.py
albertz: If there is at any point in the future some config class here, which the user would import in its Sisyphus setup, then having the `import librasr` in the global scope can be problematic.

DanEnergetics: > [...] config class here, which the user would import in its Sisyphus setup

Yes this is indeed a good argument against the global import.

michelwi (approved):
lgtm now. Would you give it another quick test before merging?

PR #64, opened 2024-12-18
Add sampler: log uniform

christophmluscher:
Adds a sampler for NCE loss

Implemented by @Gerstenberger


i6_models/samplers/log_uniform.py
```diff
@@ -0,0 +1,30 @@
...
+
+        # assumes count-sorted vocabulary, descending
+        self.num_classes = num_classes
+
+        # approximately zipf distribution
+        self._distribution = [
+            (math.log1p(w + 1) - math.log1p(w)) / math.log1p(self.num_classes) for w in range(self.num_classes)
+        ]
+        self._distribution = torch.tensor(self._distribution).clamp(min=1e-10)
+        self._distribution /= self._distribution.sum()
```
albertz, line 22: Why first make a Python list of Python floats? Why not use Torch right away?
```suggestion
        ws = torch.arange(self.num_classes, dtype=torch.get_default_dtype())
        self._distribution = (torch.log1p(ws + 1) - torch.log1p(ws)) / torch.log1p(torch.tensor(self.num_classes))
        self._distribution.clamp_(min=1e-10)
        self._distribution /= self._distribution.sum()
```

i6_models/samplers/log_uniform.py
```diff
@@ -0,0 +1,30 @@
...
+
+        self._cat_sampler = torch.distributions.categorical.Categorical(probs=self._distribution.cuda())
```
albertz, line 24:
```suggestion
        self._cat_sampler = torch.distributions.categorical.Categorical(probs=self._distribution)
```

i6_models/samplers/log_uniform.py
```diff
@@ -0,0 +1,34 @@
+__all__ = ["LogUniformSampler"]
+
+
+import torch
+from torch import nn
+from typing import Optional
+
+
+class LogUniformSampler(nn.Module):
```
curufinwe, line 9: This class is a subclass of nn.Module, which makes it a part and thus should be in the parts folder instead of a dedicated samplers folder.

curufinwe (changes requested):

i6_models/parts/samplers/log_uniform.py
```diff
@@ -0,0 +1,37 @@
...
+
+        # approximately zipf distribution
+        ws = torch.arange(self.num_classes, dtype=torch.get_default_dtype(), device=device)
+        self._distribution = (torch.log1p(ws + 1) - torch.log1p(ws)) / torch.log1p(torch.tensor(self.num_classes))
+        self._distribution.clamp_(min=1e-10)
+        self._distribution /= self._distribution.sum()
+
+        self._cat_sampler = torch.distributions.categorical.Categorical(probs=self._distribution)
+
+    def sample(self, num_samples: int) -> torch.Tensor:
```
JackTemaki, line 33: It would be nice to have the shapes documented for sample and log_prob.

i6_models/parts/samplers/log_uniform.py
```diff
@@ -0,0 +1,37 @@
...
+        """
+        super().__init__()
+
+        # assumes count-sorted vocabulary, descending
+        self.num_classes = num_classes
+
+        # approximately zipf distribution
+        ws = torch.arange(self.num_classes, dtype=torch.get_default_dtype(), device=device)
+        self._distribution = (torch.log1p(ws + 1) - torch.log1p(ws)) / torch.log1p(torch.tensor(self.num_classes))
+        self._distribution.clamp_(min=1e-10)
```
JackTemaki, line 28: Is this always a safe choice, or does this need to be configurable?

JackTemaki (changes requested):

i6_models/parts/samplers/log_uniform.py
christophmluscher: not sure. I ll make it configurable but define a default value.

JackTemaki (approved):

curufinwe (approved):

PR #65, opened 2024-12-18
Add loss: Noise Constrastive Estimation

christophmluscher:
Add NCE loss

implemented by @Gerstenberger

depends on #64


i6_models/losses/nce.py
```diff
@@ -17,31 +16,45 @@ class NoiseContrastiveEstimationLossV1(nn.Module):
...
 
         with torch.no_grad():
             samples = self.noise_distribution_sampler.sample(self.num_samples).cuda()
 
             # log-probabilities for the noise distribution k * q(w|h)
-            sampled_prob = math.log(self.num_samples) + self.noise_distribution_sampler.log_prob(
-                samples
-            )  # [num_samples]
-            true_sample_prob = math.log(self.num_samples) + self.noise_distribution_sampler.log_prob(target)  # [B x T]
+            ws = torch.log(torch.Tensor([self.num_samples]))
```
albertz, line 55: Don't you want that this is a scalar? Also, not use `torch.tensor`?

christophmluscher: Idea was to get rid of the `import math` and I think the result should be the same... or not?!

albertz: Here you get a tensor of shape [1], but I think you want a tensor of shape []. I was just a bit confused about that.
```suggestion
            ws = torch.log(torch.tensor(float(self.num_samples)))
```

i6_models/losses/nce.py
```diff
@@ -0,0 +1,105 @@
...
+    "NoiseContrastiveEstimationLossV1",
+]
+
+import torch
+from torch import nn
+from torch.nn import functional as F
+from typing import Optional
+
+
+class NoiseContrastiveEstimationLossV1(nn.Module):
```
curufinwe, line 11: This class is a subclass of nn.Module, which makes it a part and thus should be in the parts folder instead of a dedicated samplers folder.

curufinwe (changes requested):

Gerstenberger:
Added some small comments about documentation. Otherwise looks fine to me.

i6_models/parts/losses/nce.py
```diff
@@ -0,0 +1,111 @@
...
+        self.noise_distribution_sampler = noise_distribution_sampler
+        self.log_norm_term = log_norm_term
+        self.device = device
+
+        self._bce = nn.BCEWithLogitsLoss(reduction=reduction)
+
+    def forward(self, data: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+        """
+
+        :param data: the tensor for the input data, where batch and time are flattened resulting in [B x T, F] shape.
```
JackTemaki, line 53: The comments here are more confusing than helpful. So it should be specified what data is (or even rename it), First I assumed it is logits, but then I guess this should be the input to the "output" layer of the provided model? This is very confusing and unclear. Then it should also be clear what is the input shape. When you say "where batch and time are flattened resulting in" this sounds like this is done internally, but what you mean it expects [B x T, F] as only valid input, correct?

JackTemaki (changes requested):

i6_models/parts/losses/nce.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        Noise contrastive estimation (NCE) loss implementation.
+        Used to estimate the softmax. Normally for very large softmax sizes, for example word-level LM.
+
+        :param num_samples: num of samples for the estimation, normally a value between 1000-4000.
+                            2000 is a good starting point.
+        :param model: model on which the NCE loss is to be applied. The model requires a member called `output`, which
+            is expected to be a linear layer with bias and weight members. The member `output` represents the output
+            layer of the model allowing access to the parameters during loss computation.
+        :param noise_distribution_sampler: for example `i6_model.samplers.LogUniformSampler`.
```
curufinwe, line 35:
```suggestion
        :param noise_distribution_sampler: for example `i6_models.parts.samplers.LogUniformSampler`.
```

i6_models/parts/losses/nce.py
```diff
@@ -0,0 +1,114 @@
...
+        reduction: str = "none",
+        device: Optional[str] = None,
+    ) -> None:
+        """
+        Noise contrastive estimation (NCE) loss implementation.
+        Used to estimate the softmax. Normally for very large softmax sizes, for example word-level LM.
+
+        :param num_samples: num of samples for the estimation, normally a value between 1000-4000.
+                            2000 is a good starting point.
+        :param model: model on which the NCE loss is to be applied. The model requires a member called `output`, which
```
curufinwe, line 32: Why not pass in these two weights explicitly?

curufinwe (changes requested):

i6_models/parts/losses/nce.py
christophmluscher: hmmm I think that would be possible... I dont have a strong preference either way.. do you prefer explicitly passing of the weights?
@Gerstenberger or is there a specific reason why you pass the model and not the weights?

i6_models/parts/losses/nce.py
christophmluscher: Are the new docs clearer?

i6_models/parts/losses/nce.py
Gerstenberger: Passing the weights is fine.
There is no particular reason for passing the model instead. Initially, i did pass the weights itself but changed that for some reason i cannot remember.

curufinwe: Let's pass the weights then directly. This way you don't need to have the model follow some convention.

curufinwe (approved):

i6_models/parts/losses/nce.py
```diff
@@ -0,0 +1,116 @@
...
+class NoiseContrastiveEstimationLossV1(nn.Module):
+    __constants__ = ["num_samples", "log_norm_term", "reduction"]
+    num_samples: int
+    log_norm_term: float
+
+    def __init__(
+        self,
+        num_samples: int,
+        *,
+        # TODO model: nn.Module,
```
christophmluscher, line 20:
```suggestion
```
cleanup

JackTemaki (approved):

PR #66, opened 2024-12-18
Add part: lstm block

christophmluscher:
Adds LSTM Block

### DIFF
--- a/i6_models/parts/lstm.py
+++ b/i6_models/parts/lstm.py
@@ -1,4 +1,4 @@
-__all__ = ["LstmBlockV1Config", "LstmBlockV1", "LstmBlockV2Config", "LstmBlockV2"]
+__all__ = ["LstmBlockV1Config", "LstmBlockV1"]
 
 from dataclasses import dataclass
 import torch
@@ -60,17 +60,5 @@ class LstmBlockV1(nn.Module):
             padding_value=0.0,
             batch_first=True,
         )
-
-        return lstm_out, seq_len
-
-
-@dataclass
-class LstmBlockV2Config(LstmBlockV1Config):
-    pass
-
-
-class LstmBlockV2(LstmBlockV1):
-    def forward(self, x: torch.Tensor, seq_len: torch.Tensor) -> (torch.Tensor, torch.Tensor):
-        lstm_out, _ = self.lstm_stack(x)
 
         return lstm_out, seq_len
--- a/i6_models/parts/lstm.py
+++ b/i6_models/parts/lstm.py
@@ -1,0 +1,76 @@
+__all__ = ["LstmBlockV1Config", "LstmBlockV1", "LstmBlockV2Config", "LstmBlockV2"]
+
+from dataclasses import dataclass
+import torch
+from torch import nn
+from typing import Dict, Union
+
+from i6_models.config import ModelConfiguration
+
+
+@dataclass
+class LstmBlockV1Config(ModelConfiguration):
+    input_dim: int
+    hidden_dim: int
+    num_layers: int
+    bias: bool
+    dropout: float
+    bidirectional: bool
+    enforce_sorted: bool
+
+    @classmethod
+    def from_dict(cls, model_cfg_dict: Dict):
+        model_cfg_dict = model_cfg_dict.copy()
+        return cls(**model_cfg_dict)
+
+
+class LstmBlockV1(nn.Module):
+    def __init__(self, model_cfg: Union[LstmBlockV1Config, Dict], **kwargs):
+        super().__init__()
+
+        self.cfg = LstmBlockV1Config.from_dict(model_cfg) if isinstance(model_cfg, Dict) else model_cfg
+
+        self.dropout = self.cfg.dropout
+        self.enforce_sorted = None
+        self.lstm_stack = nn.LSTM(
+            input_size=self.cfg.input_dim,
+            hidden_size=self.cfg.hidden_dim,
+            num_layers=self.cfg.num_layers,
+            bias=self.cfg.bias,
+            dropout=self.dropout,
+            batch_first=True,
+            bidirectional=self.cfg.bidirectional,
+        )
+
+    def forward(self, x: torch.Tensor, seq_len: torch.Tensor) -> (torch.Tensor, torch.Tensor):
+        if not torch.jit.is_scripting() and not torch.jit.is_tracing():
+            if seq_len.get_device() >= 0:
+                seq_len = seq_len.cpu()
+
+        lstm_packed_in = nn.utils.rnn.pack_padded_sequence(
+            input=x,
+            lengths=seq_len,
+            enforce_sorted=self.enforce_sorted,
+            batch_first=True,
+        )
+
+        lstm_out, _ = self.lstm_stack(lstm_packed_in)
+        lstm_out, _ = nn.utils.rnn.pad_packed_sequence(
+            lstm_out,
+            padding_value=0.0,
+            batch_first=True,
+        )
+
+        return lstm_out, seq_len
+
+
+@dataclass
+class LstmBlockV2Config(LstmBlockV1Config):
+    pass
+
+
+class LstmBlockV2(LstmBlockV1):
+    def forward(self, x: torch.Tensor, seq_len: torch.Tensor) -> (torch.Tensor, torch.Tensor):
+        lstm_out, _ = self.lstm_stack(x)
+
+        return lstm_out, seq_len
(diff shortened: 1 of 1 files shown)


i6_models/parts/lstm.py
```diff
@@ -0,0 +1,64 @@
...
+    input_dim: int
+    hidden_dim: int
+    num_layers: int
+    bias: bool
+    dropout: float
+    bidirectional: bool
+    enforce_sorted: bool
+
+    @classmethod
+    def from_dict(cls, model_cfg_dict: Dict):
```
NeoLegends, line 22: Same Q as in the other PR: why is this necessary now, and hasn't been for the other assemblies?

NeoLegends (approved):
One minor Q.

i6_models/parts/lstm.py
```diff
@@ -0,0 +1,64 @@
...
+            input_size=self.cfg.input_dim,
+            hidden_size=self.cfg.hidden_dim,
+            num_layers=self.cfg.num_layers,
+            bias=self.cfg.bias,
+            dropout=self.dropout,
+            batch_first=True,
+            bidirectional=self.cfg.bidirectional,
+        )
+
+    def forward(self, x: torch.Tensor, seq_len: torch.Tensor) -> (torch.Tensor, torch.Tensor):
```
albertz, line 45:
```suggestion
    def forward(self, x: torch.Tensor, seq_len: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
```

i6_models/parts/lstm.py
```diff
@@ -0,0 +1,70 @@
...
+    enforce_sorted: bool
+
+    @classmethod
+    def from_dict(cls, model_cfg_dict: Dict):
+        model_cfg_dict = model_cfg_dict.copy()
+        return cls(**model_cfg_dict)
+
+
+class LstmBlockV1(nn.Module):
+    def __init__(self, model_cfg: Union[LstmBlockV1Config, Dict], **kwargs):
```
albertz, line 28:
```suggestion
    def __init__(self, model_cfg: Union[LstmBlockV1Config, Dict[str, Any]], **kwargs):
```

i6_models/parts/lstm.py
```diff
@@ -0,0 +1,70 @@
...
+            bias=self.cfg.bias,
+            dropout=self.dropout,
+            batch_first=True,
+            bidirectional=self.cfg.bidirectional,
+        )
+
+    def forward(self, x: torch.Tensor, seq_len: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+        if not torch.jit.is_scripting() and not torch.jit.is_tracing():
+            if seq_len.get_device() >= 0:
+                seq_len = seq_len.cpu()
```
albertz, line 54:
```suggestion
            seq_len = seq_len.cpu()
```

i6_models/parts/lstm.py
```diff
@@ -0,0 +1,70 @@
...
+            hidden_size=self.cfg.hidden_dim,
+            num_layers=self.cfg.num_layers,
+            bias=self.cfg.bias,
+            dropout=self.dropout,
+            batch_first=True,
+            bidirectional=self.cfg.bidirectional,
+        )
+
+    def forward(self, x: torch.Tensor, seq_len: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+        if not torch.jit.is_scripting() and not torch.jit.is_tracing():
```
albertz, line 52: Why only when not scripting? Don't you want that `seq_len` is always on CPU?

christophmluscher: I followed the example in the blstm part.
```
        if not torch.jit.is_scripting() and not torch.jit.is_tracing():
            # during graph mode we have to assume all Tensors are on the correct device,
            # otherwise move lengths to the CPU if they are on GPU
            if seq_len.get_device() >= 0:
                seq_len = seq_len.cpu()
```
I did not copy the comment over... I did not yet get to look why this is necessary
@JackTemaki you implemented the BLSTM IIRC. You remember why this was done in this way?

i6_models/parts/lstm.py
```diff
@@ -0,0 +1,70 @@
...
+from dataclasses import dataclass
+import torch
+from torch import nn
+from typing import Dict, Tuple, Union
+
+from i6_models.config import ModelConfiguration
+
+
+@dataclass
+class LstmBlockV1Config(ModelConfiguration):
```
Atticus1806, line 12: Could you add docs here please? You can probably copy them from Blst

i6_models/parts/lstm.py
```diff
@@ -0,0 +1,70 @@
...
+    input_dim: int
+    hidden_dim: int
+    num_layers: int
+    bias: bool
+    dropout: float
+    bidirectional: bool
+    enforce_sorted: bool
+
+    @classmethod
+    def from_dict(cls, model_cfg_dict: Dict[str, Any]):
```
Atticus1806, line 22: I don't see this for other part configs, why do we need this here?

i6_models/parts/lstm.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@dataclass
+class LstmBlockV1Config(ModelConfiguration):
+    input_dim: int
+    hidden_dim: int
+    num_layers: int
+    bias: bool
+    dropout: float
+    bidirectional: bool
```
Atticus1806, line 18: should we allow this? I feel like if we have bidirectional here, the BLSTM part becomes redundant, which is maybe okay, but might also cause two different branches that do the same, which I am not sure we want (if there are potential extensions later). We could maybe also just deprecate the BLSTM block?

i6_models/parts/lstm.py
```diff
@@ -0,0 +1,70 @@
...
+    @classmethod
+    def from_dict(cls, model_cfg_dict: Dict[str, Any]):
+        model_cfg_dict = model_cfg_dict.copy()
+        return cls(**model_cfg_dict)
+
+
+class LstmBlockV1(nn.Module):
+    def __init__(self, model_cfg: Union[LstmBlockV1Config, Dict[str, Any]], **kwargs):
+        """
+        Model definition of LSTM block. Contains single lstm stack and padding sequence in forward call.
```
Atticus1806, line 30: Also add the "including dropout, batch-first variant,  hardcoded to use B,T,F input" part please.

Atticus1806: Could also add the supports scripting part.

i6_models/parts/lstm.py
```diff
@@ -0,0 +1,70 @@
...
+            input_size=self.cfg.input_dim,
+            hidden_size=self.cfg.hidden_dim,
+            num_layers=self.cfg.num_layers,
+            bias=self.cfg.bias,
+            dropout=self.dropout,
+            batch_first=True,
+            bidirectional=self.cfg.bidirectional,
+        )
+
+    def forward(self, x: torch.Tensor, seq_len: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
```
Atticus1806, line 51: doc pls.

i6_models/parts/lstm.py
```diff
@@ -0,0 +1,70 @@
...
+
+        lstm_packed_in = nn.utils.rnn.pack_padded_sequence(
+            input=x,
+            lengths=seq_len,
+            enforce_sorted=self.enforce_sorted,
+            batch_first=True,
+        )
+
+        lstm_out, _ = self.lstm_stack(lstm_packed_in)
+        lstm_out, _ = nn.utils.rnn.pad_packed_sequence(
```
Atticus1806, line 64: Just out of curiosity: why does black force the new lines here but not for the blstm? Shouldnt it be the same line length?

Atticus1806 (changes requested):

i6_models/parts/lstm.py
JackTemaki: The question is, is this still relevant? This was something I added at some point, but if this is not needed for ONNX export this should be removed until there is actually a reason for it.

i6_models/parts/lstm.py
christophmluscher: I use it in the model definition to conert a dict to the config class. Need might be a bit strong but I like it :D

i6_models/parts/lstm.py
christophmluscher: well this is dependent if you manually set the last commata. if set it will force new lines

i6_models/parts/lstm.py
christophmluscher: I use it in the model definition to convert a dict to the config class. Necessary might be a bit strong but I like it :D

Atticus1806 (approved):

PR #67, opened 2024-12-18
Add LstmEncoderV1 assembly

christophmluscher:
Adds LSTM encoder

depends on #66

### DIFF
--- a/i6_models/assemblies/lstm/lstm_v1.py
+++ b/i6_models/assemblies/lstm/lstm_v1.py
@@ -22,17 +22,17 @@ class LstmEncoderV1Config(ModelConfiguration):
     init_args: Optional[Dict[str, Any]] = None
 
     @classmethod
-    def from_dict(cls, model_cfg_dict: Dict):
+    def from_dict(cls, model_cfg_dict: Dict[str, Any]):
         model_cfg_dict = model_cfg_dict.copy()
         model_cfg_dict["lstm_layers_cfg"] = LstmBlockV1Config.from_dict(model_cfg_dict["lstm_layers_cfg"])
         return cls(**model_cfg_dict)
 
 
 class LstmEncoderV1(nn.Module):
-    def __init__(self, model_cfg: Union[LstmEncoderV1Config, Dict], **kwargs):
+    def __init__(self, model_cfg: Union[LstmEncoderV1Config, Dict[str, Any]], **kwargs):
         super().__init__()
 
-        self.cfg = LstmEncoderV1Config.from_dict(model_cfg) if isinstance(model_cfg, Dict) else model_cfg
+        self.cfg = LstmEncoderV1Config.from_dict(model_cfg) if isinstance(model_cfg, dict) else model_cfg
 
         self.embedding = nn.Embedding(self.cfg.input_dim, self.cfg.embed_dim)
         self.embed_dropout = nn.Dropout(self.cfg.embed_dropout)
--- a/i6_models/assemblies/lstm/lstm_v1.py
+++ b/i6_models/assemblies/lstm/lstm_v1.py
@@ -16,12 +16,12 @@ from i6_models.parts.lstm import LstmBlockV1Config, LstmBlockV1
 class LstmEncoderV1Config(ModelConfiguration):
     """
     :param init_args: used to initialize parameters of modules, example:
-                      ```
-                        {
-                            "init_args_w": {"func": "normal", "arg": {"mean": 0.0, "std": 0.1}},
-                            "init_args_b": {"func": "normal", "arg": {"mean": 0.0, "std": 0.1}},
-                        }
-                      ```
+        ```
+        {
+            "init_args_w": {"func": "normal", "arg": {"mean": 0.0, "std": 0.1}},
+            "init_args_b": {"func": "normal", "arg": {"mean": 0.0, "std": 0.1}},
+        }
+        ```
     """
 
     input_dim: int
--- a/i6_models/assemblies/lstm/lstm_v1.py
+++ b/i6_models/assemblies/lstm/lstm_v1.py
@@ -1,18 +1,15 @@
 __all__ = [
     "LstmEncoderV1Config",
     "LstmEncoderV1",
-    "LstmEncoderV2Config",
-    "LstmEncoderV2",
 ]
 
 from dataclasses import dataclass
-import math
 import torch
 from torch import nn
 from typing import Any, Dict, Optional, Tuple, Union
 
 from i6_models.config import ModelConfiguration
-from i6_models.parts.lstm import LstmBlockV1Config, LstmBlockV1, LstmBlockV2Config, LstmBlockV2
+from i6_models.parts.lstm import LstmBlockV1Config, LstmBlockV1
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


i6_models/assemblies/lstm/lstm_v1.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@dataclass
+class LstmEncoderV1Config(ModelConfiguration):
+    input_dim: int
+    embed_dim: int
+    embed_dropout: float
+    lstm_layers_cfg: LstmBlockV1Config
+    lstm_dropout: float
+    init_args: Optional[Dict[str, Any]] = None
```
NeoLegends, line 22: Can you add some docs on what can go in here?

i6_models/assemblies/lstm/lstm_v1.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @classmethod
+    def from_dict(cls, model_cfg_dict: Dict):
```
NeoLegends, line 25: I haven't seen these functions for other config classes. What is the reason they're needed here? Should we add generic "serialization" support to all other parameter dataclasses as well?

NeoLegends (commented):
Just two minor Qs, rest LGTM!

i6_models/assemblies/lstm/lstm_v1.py
christophmluscher: I think this depends on how you import/call your model and how the params are then passed to the model.. 
I like to have a dataclass in order to set the params in my config. and I like to have a single attribute `cfg` instead of numerous attributes for the model..
But passing a dict instead of the dataclass allows me to write someting like this in the returnn config without the need to import the data classes:
```
get_model = __import__("functools").partial(
    __import__(
        "apptek_asr.users.cluescher.lib.returnn.pytorch_networks.lm.lstm.alex_variant_v6",
        fromlist=["Model"],
    ).Model,
    **{
        "model_config_dict": {
            "vocab_dim": 249352,
            "embed_dim": 128,
            "hidden_dim": 512,
            "n_lstm_layers": 2,
            "init_args": {
                "init_args_w": {"func": "normal", "arg": {"mean": 0.0, "std": 0.1}},
                "init_args_b": {"func": "normal", "arg": {"mean": 0.0, "std": 0.1}},
            },
            "num_noise": 25,
            "bias": True,
            "bottle_neck_dim": 512,
            "dropout": 0.2,
        }
    }
)
```
maybe needed is therefor a bit strong.. :D

i6_models/assemblies/lstm/lstm_v1.py
```diff
@@ -0,0 +1,70 @@
...
+        model_cfg_dict = model_cfg_dict.copy()
+        model_cfg_dict["lstm_layers_cfg"] = LstmBlockV1Config.from_dict(model_cfg_dict["lstm_layers_cfg"])
+        return cls(**model_cfg_dict)
+
+
+class LstmEncoderV1(nn.Module):
+    def __init__(self, model_cfg: Union[LstmEncoderV1Config, Dict], **kwargs):
+        super().__init__()
+
+        self.cfg = LstmEncoderV1Config.from_dict(model_cfg) if isinstance(model_cfg, Dict) else model_cfg
```
albertz, line 35:
```suggestion
        self.cfg = LstmEncoderV1Config.from_dict(model_cfg) if isinstance(model_cfg, dict) else model_cfg
```
`Dict` is from `typing`, used to express K/V type of a `dict`.

i6_models/assemblies/lstm/lstm_v1.py
albertz, line 25:
```suggestion
    def from_dict(cls, model_cfg_dict: Dict[str, Any]):
```

i6_models/assemblies/lstm/lstm_v1.py
```diff
@@ -0,0 +1,87 @@
...
+
+from i6_models.config import ModelConfiguration
+from i6_models.parts.lstm import LstmBlockV1Config, LstmBlockV1
+
+
+@dataclass
+class LstmEncoderV1Config(ModelConfiguration):
+    """
+    :param init_args: used to initialize parameters of modules, example:
+                      ```
```
albertz, line 19: I think we usually indent by 4 on a continued doc line, or not?

i6_models/assemblies/lstm/lstm_v1.py
```diff
@@ -0,0 +1,87 @@
...
+class LstmEncoderV1Config(ModelConfiguration):
+    """
+    :param init_args: used to initialize parameters of modules, example:
+        ```
+        {
+            "init_args_w": {"func": "normal", "arg": {"mean": 0.0, "std": 0.1}},
+            "init_args_b": {"func": "normal", "arg": {"mean": 0.0, "std": 0.1}},
+        }
+        ```
+    """
```
Atticus1806, line 25: Again also please add the docs for the other parameters.

i6_models/assemblies/lstm/lstm_v1.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @classmethod
+    def from_dict(cls, model_cfg_dict: Dict[str, Any]):
+        model_cfg_dict = model_cfg_dict.copy()
+        model_cfg_dict["lstm_layers_cfg"] = LstmBlockV1Config.from_dict(model_cfg_dict["lstm_layers_cfg"])
+        return cls(**model_cfg_dict)
+
+
+class LstmEncoderV1(nn.Module):
+    def __init__(self, model_cfg: Union[LstmEncoderV1Config, Dict[str, Any]], **kwargs):
```
Atticus1806, line 42: Do we need kwargs here?

i6_models/assemblies/lstm/lstm_v1.py
```diff
@@ -0,0 +1,87 @@
...
+
+        self.lstm_block = LstmBlockV1(self.cfg.lstm_layers_cfg)
+        self.lstm_dropout = nn.Dropout(self.cfg.lstm_dropout)
+
+        if self.cfg.init_args is not None:
+            self._param_init(**self.cfg.init_args)
+
+    def _param_init(self, init_args_w=None, init_args_b=None):
+        for m in self.modules():
+            for name, param in m.named_parameters():
```
Atticus1806, line 65: I think I would be a bit more cautious here. This would fail with `KeyError` if given something that does not have `func` and `arg` in it. Maybe assert this beforehand? should we maybe even enforce it in the config above via post init?

Atticus1806 (changes requested):

i6_models/assemblies/lstm/lstm_v1.py
christophmluscher: Added a post init to handle this.

Atticus1806 (approved):

i6_models/assemblies/lstm/lstm_v1.py
JackTemaki: I would not have a dict where flexibility is not needed. So maybe `init_args_w` and `init_args_b` directly separated? Or do you want this to be in this "loose" format to later extend it to separate LSTM and embedding parameters? Usually I would say more explicit is always better, because here you can easily add entries that do nothing and you will never detect the error.

JackTemaki: Ah, I take it back, the dict is added as kwargs to _param_init, I think then it crashes if you have wrong entries. Well, not optimal but maybe fine then.

curufinwe (approved):

PR #68, opened 2025-03-04
flip the mask in BestRQ

Judyxujj:
For the mask generated in RandomMask, True should indicate that the frames are not masked and False indicate the frames are masked, now it is flipped and thus incorrect.


albertz:
The docstring of RandomMask should describe this behavior.

And where is RandomMask used? Is it properly documented there what kind of masked is required?

PR #69, opened 2025-03-07
Add AED+CTC model glue code

NeoLegends:
This PR adds the glue code for the RF AED+CTC model. The Conformer encoder and Transformer decoder themselves are part of RETURNN.

Open Qs:

- Should the glue code even go in here or into a different repository? At AppTek's internal meeting I'm not sure we made it clear where exactly the individual parts should go, but i6_models was suggested. Maybe apptek_asr is a better place.
- Should the callbacks be part of the PR? For torch models we don't have these callbacks (yet?) but only the model code. W/o them, the model isn't of much use, however. 
- The callbacks this PR adds cannot be used as `forward_step` or `get_model` directly, but require some small, generic wrapper functions for RETURNN compatibility. The callbacks use a set of interfaces called `ModelDef`/`RecogDef`/`TrainDef`, which are not part of this PR yet. Should these interfaces be included here, or would you prefer having them e.g. in `i6_experiments/common`, and should the wrapper functions that make e.g. the `train_def` compatible as `train_step` be here as well? They can also just be inlined and removed from this PR.
- This PR leaves the properties used for scaling up the model configurable and the rest it doesn't. Should everything be configurable? (I don't think so)
- Feature extraction should at best be RASR compatible for future-proofing the model. I don't know how to do that (see comment).

### DIFF
--- a/i6_models/assemblies/rf/aed_ctc_v1.py
+++ b/i6_models/assemblies/rf/aed_ctc_v1.py
@@ -1,0 +1,151 @@
+__all__ = ["ConformerAedCtcModelV1"]
+
+from dataclasses import dataclass, field
+from typing import Any, Dict, Optional, Sequence, Tuple
+
+
+from i6_models.config import ModelConfiguration
+
+from returnn.tensor import Tensor, Dim
+import returnn.frontend as rf
+from returnn.frontend.encoder.conformer import ConformerEncoder, ConformerConvSubsample
+from returnn.frontend.decoder.transformer import TransformerDecoder
+
+
+@dataclass
+class ConformerAedCtcModelV1Config(ModelConfiguration):
+    """
+    Attributes:
+        blank_idx: index of the blank label
+        eos_idx: index of the end-of-sequence label
+        bos_idx: index of the begin-of-sequence label
+        num_enc_layers: number of Conformer encoder layers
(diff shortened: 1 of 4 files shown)


i6_models/assemblies/rf/aed_ctc_v1.py
```diff
@@ -0,0 +1,151 @@
...
+            "num_spatial_mask_factor": config.typed_value("specaugment_num_spatial_mask_factor") or 100,
+        }
+
+    def encode(
+        self, source: Tensor, *, in_spatial_dim: Dim, collected_outputs: Optional[Dict[str, Tensor]] = None
+    ) -> Tuple[rf.State, Dim]:
+        """encode, and extend the encoder output for things we need in the decoder"""
+        source, in_spatial_dim = rf.audio.log_mel_filterbank_from_raw(
+            source, in_spatial_dim=in_spatial_dim, out_dim=self.in_dim, sampling_rate=self.sampling_rate
+        )
```
NeoLegends, line 146: I don't know RF enough yet to know how I would use `i6_models.primitives.feature_extraction.RasrCompatibleLogMelFeatureExtractionV1` here. How would I reconstruct the `in_spatial_dim` that would normally result from the call?

NeoLegends:
Superseded by PR in apptek_asr.

PR #71, opened 2025-03-28
Implement masked batch normalization

NeoLegends:
We commonly use the regular torch `nn.BatchNorm1d`, which does not mask non-sequence positions when computing the batch statistics. Masking them away is more correct.

Based on https://gist.github.com/amiasato/902fc14afa37a7537386f7b0c5537741.

### DIFF
--- a/i6_models/parts/masked_norm.py
+++ b/i6_models/parts/masked_norm.py
@@ -1,0 +1,113 @@
+__all__ = ["MaskedBatchNorm1dV1"]
+
+from typing import Optional
+
+import torch
+from torch import nn, Tensor
+
+
+def _lengths_to_mask(lengths: Tensor, max_len=None, dtype=None) -> Tensor:
+    """
+    Converts a "lengths" tensor to its binary mask representation.
+    """
+    assert len(lengths.shape) == 1, "Length shape should be 1 dimensional."
+    max_len = max_len or lengths.max().item()
+    mask = torch.arange(max_len, device=lengths.device, dtype=lengths.dtype).expand(
+        len(lengths), max_len
+    ) < lengths.unsqueeze(1)
+    if dtype is not None:
+        mask = mask.to(dtype=dtype, device=lengths.device)
+    return mask
+
+
+class MaskedBatchNorm1dV1(nn.BatchNorm1d):
+    """
+    Masked version of the 1D Batch normalization.
+
+    Receives a N-dim tensor of sequence lengths per batch element
+    along with the regular input for masking.
+
+    Same construction arguments as pytorch's BatchNorm1d.
+    """
+
+    def __init__(
+        self,
+        num_features: int,
+        *,
+        eps: float = 1e-5,
+        momentum: float = 0.1,
+        affine: bool = True,
+        track_running_stats: bool = True,
+        device: Optional[torch.device] = None,
+        dtype: Optional[torch.dtype] = None,
+    ):
+        super().__init__(
+            num_features,
+            eps=eps,
+            momentum=momentum,
+            affine=affine,
+            track_running_stats=track_running_stats,
+            device=device,
+            dtype=dtype,
+        )
+
+    def forward(self, inp: Tensor, lengths_or_mask: Tensor):
+        """
+        Applies batch norm to `inp`.
+
+        :param input: data to normalize, shape (B..., T, F)
+        :param lengths_or_mask: seq length tensor if shape (B...,),
+            or mask tensor if the shape is (B..., T).
+        """
+
+        self._check_input_dim(inp)
+
+        exponential_average_factor = 0.0
+        if self.training and self.track_running_stats and self.num_batches_tracked is not None:
+            self.num_batches_tracked += 1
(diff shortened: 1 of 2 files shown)


i6_models/parts/masked_norm.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # we use the mask to calculate the mean
+        mask = mask / n
+        mask = mask.unsqueeze(-2)
+
+        reduce_dims = list(range(inp.ndim - 2)) + [inp.ndim - 1]
+        if not self.track_running_stats:
+            mean = (mask * inp).sum(reduce_dims)
+            var = ((mask * inp**2).sum(reduce_dims) - mean**2) * n / (n - 1)
+        elif self.training and n > 1:
```
kuacakuaca, line 95: you don't need n > 1 check anymore.

i6_models/parts/masked_norm.py
```diff
@@ -0,0 +1,116 @@
...
+
+import torch
+from torch import nn, Tensor
+
+
+def _lengths_to_mask(lengths: Tensor, max_len=None, dtype=None) -> Tensor:
+    """
+    Converts a "lengths" tensor to its binary mask representation.
+    """
+    assert len(lengths.shape) == 1, "Length shape should be 1 dimensional."
```
kuacakuaca, line 13: does this have to be one-dimensional? if so, then the dimensionality of `inp` in the forward function is also fixed.

kuacakuaca (changes requested):

i6_models/parts/masked_norm.py
```diff
@@ -0,0 +1,127 @@
...
+        mask = mask.unsqueeze(-2)
+
+        reduce_dims = list(range(inp.ndim - 2)) + [inp.ndim - 1]
+        if not self.track_running_stats:
+            mean = (mask * inp).sum(reduce_dims)
+            var = ((mask * inp**2).sum(reduce_dims) - mean**2) * n / (n - 1)
+        elif self.training:
+            mean = (mask * inp).sum(reduce_dims)
+            # Var(X) = E[X^2] - E[X]^2
+            var = (mask * inp**2).sum(reduce_dims) - mean**2
```
michelwi, line 109: Why is there a distinction of having/not-having the factor `n / (n - 1)` in training vs (not training or not track_running_stats)? especially if we add the factor again to the running_var that we are tracking.

i6_models/parts/masked_norm.py
```diff
@@ -0,0 +1,127 @@
...
+
+            with torch.no_grad():
+                self.running_mean = (
+                    exponential_average_factor * mean + (1 - exponential_average_factor) * self.running_mean
+                )
+                # torch updates running statistics with unbiased var
+                self.running_var = (
+                    exponential_average_factor * var * n / (n - 1) + (1 - exponential_average_factor) * self.running_var
```
michelwi, line 117: is exponentially averaging the variances the correct thing to do? The result will be different from the variance of the whole data.. e.g.
```
# prepare 10 batches of random data, each shifted by a different amount
datas = [torch.tensor(np.random.random(data.shape) + i) for i in range(10)]
norm = MaskedBatchNorm1dV1(num_features=data.shape[1], momentum=None, affine=False).train()
_ = [norm(d, data_lens) for d in datas]

norm.running_var, norm.running_mean  # 0.0856, 4.9849
torch.var_mean(torch.cat(datas))     # 8.3816, 4.9849
```
In practice I doubt the inter-batch variance will be much larger than the intra-batch variance.. so maybe we do not need to be concerned?

i6_models/parts/masked_norm.py
```diff
@@ -0,0 +1,127 @@
...
+        super().__init__(
+            num_features,
+            eps=eps,
+            momentum=momentum,
+            affine=affine,
+            track_running_stats=track_running_stats,
+            device=device,
+            dtype=dtype,
+        )
+
```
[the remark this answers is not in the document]
NeoLegends: IIRC torch modules require a constructor to be defined.

i6_models/parts/masked_norm.py
NeoLegends: Citing torch docs: 

> At train time in the forward pass, the variance is calculated via the biased estimator, equivalent to torch.var(input, unbiased=False). However, the value stored in the moving average of the variance is calculated via the unbiased estimator, equivalent to torch.var(input, unbiased=True).

i6_models/parts/masked_norm.py
```diff
@@ -0,0 +1,127 @@
+__all__ = ["MaskedBatchNorm1dV1"]
+
+from typing import Optional
+
+import torch
+from torch import nn, Tensor
+
+
+def _lengths_to_mask(lengths: Tensor, max_len: Optional[int] = None, dtype=None) -> Tensor:
```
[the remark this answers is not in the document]
NeoLegends: I'd leave this to a follow-up.

kuacakuaca (approved):

PR #72, opened 2025-03-28
Add support for masked batch norm to `ConformerRelPosEncoderV1`

NeoLegends:
I implemented a new convolution block to avoid breaking the API of the old one if that is used on its own elsewhere. I assumed the internals (i.e. what type of model block is used) of `ConformerRelPosEncoderV1` are relatively private/encapsulated, which is why I simply changed the existing implementation instead of adding in another version. If that is unwanted I can create a `ConformerRelPosEncoderV2`, but it would be a bit more verbose.

Depends on #71.


NeoLegends:
Superseded by https://github.com/rwth-i6/i6_models/pull/73

PR #73, opened 2025-03-28
Allow using masking batch norm in `ConformerRelPosEncoderV1`

NeoLegends:
I implemented a new convolution block to avoid breaking the API of the old one if that is used on its own elsewhere. I assumed the internals (i.e. what type of model block is used) of `ConformerRelPosEncoderV1` are relatively private/encapsulated, which is why I simply changed the existing implementation instead of adding in another version. If that is unwanted I can create a `ConformerRelPosEncoderV2`, but it would be a bit more verbose.

Second part of #71. Does not depend on the concrete implementation though.


i6_models/assemblies/conformer/conformer_rel_pos_v1.py
```diff
@@ -70,7 +70,7 @@ def __init__(self, cfg: ConformerRelPosBlockV1Config):
             elif module_name == "mhsa":
                 modules.append(ConformerMHSARelPosV1(cfg=cfg.mhsa_cfg))
             elif module_name == "conv":
-                modules.append(ConformerConvolutionV2(model_cfg=cfg.conv_cfg))
+                modules.append(ConformerConvolutionV3(model_cfg=cfg.conv_cfg))
```
michelwi, line 73: would this not change the behavior of all setups that used `ConformerRelPosBlockV1`? I think we should have the new behavior either in a V2 or enable it wit ha flag.

NeoLegends: You always had to provide the norm class/`nn.Module` from outside as `ModelConfiguration`. Existing setups won't change in behavior unless you dig into the internals of the `ConformerRelPosBlockV1`, where the API of the convolution module will change (but not its behavior).

So, knowing this, you say, do a V2 of the complete assembly?

michelwi: ah.. the new behavior is guarded by `len(signature(self.norm.forward).parameters) > 1` which is False for all norms that have worked in the past.. ok, then I am happy, I guess.

PR #74, opened 2025-04-10
Add Transformer attention decoder

NeoLegends:


i6_models/assemblies/transformer/transformer_decoder_v1.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def __init__(self, cfg: TransformerDecoderV1Config):
+        """
+        :param cfg: configuration with subunits for transformer blocks
+        """
+        super().__init__()
+
+        self.model_dim = cfg.block_cfg.ff_cfg.input_dim
+
+        self.input_dropout = nn.Dropout(cfg.input_dropout)
```
kuacakuaca, line 169: would it make sense to use `BroadcastDropout` also here?

kuacakuaca (approved):

PR #75, opened 2025-04-29
Fbw 2 fsa builder

Stefanwuu:
To support a new version of fbw loss that accepts a different fsa interface from the previous fbw loss(with additional num_edges), we add a new RASR FSA builder.

### DIFF
--- a/i6_models/parts/rasr_fsa.py
+++ b/i6_models/parts/rasr_fsa.py
@@ -42,6 +42,51 @@ class WeightedFsa(NamedTuple):
         return WeightedFsa._make(tensor.to(device) for tensor in self)
 
 
+class WeightedFsaV2(NamedTuple):
+    """
+    Convenience class that represents an FSA for the fbw2 loss from i6_native_ops. It supports scaling the weights of
+    the FSA by simple left-multiplication and moving the tensors to a different device.
+    It can be passed to :func:`i6_native_ops.fbw2.fbw2_loss`.
+
+    :param num_states: the number of states per sequence (shape: [N])
+    :param num_edges: the number of edges per sequence (shape: [N])
+    :param edges: a [3, E] tensor of edges with number of edges E and where each column is an edge
+        consisting of from-state, to-state and emission idx
+    :param weights: a [E,] tensor of weights for each edge scaled by the tdp_scale
+    :param start_end_states: a [N, 2] tensor of start and end states for each of the N sequences
+    """
+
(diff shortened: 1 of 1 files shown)


i6_models/parts/rasr_fsa.py
```diff
@@ -42,6 +42,51 @@ def to(self, device: Union[str, torch.device]) -> WeightedFsa:
...
+
+    num_states: torch.IntTensor
+    num_edges: torch.IntTensor
+    edges: torch.IntTensor
+    weights: torch.FloatTensor
+    start_end_states: torch.IntTensor
+
+    def __mul__(self, scale: float) -> WeightedFsa:
+        """Multiply the weights, i.e. the forth element, with a scale."""
+        return WeightedFsa(
```
DanEnergetics, line 67: Should this be V2?

i6_models/parts/rasr_fsa.py
```diff
@@ -42,6 +42,51 @@ def to(self, device: Union[str, torch.device]) -> WeightedFsa:
...
 
+class WeightedFsaV2(NamedTuple):
+    """
+    Convenience class that represents an FSA for the fbw2 loss from i6_native_ops. It supports scaling the weights of
+    the FSA by simple left-multiplication and moving the tensors to a different device.
+    It can be passed to :func:`i6_native_ops.fbw2.fbw2_loss`.
+
+    :param num_states: the number of states per sequence (shape: [N])
+    :param num_edges: the number of edges per sequence (shape: [N])
+    :param edges: a [3, E] tensor of edges with number of edges E and where each column is an edge
```
DanEnergetics, line 53: The [docstring of fbw_loss2](https://github.com/rwth-i6/i6_native_ops/blob/d70308a0722f0632f70a690974f3c51444664e13/i6_native_ops/fbw2/__init__.py#L77) says that the expected shape is [4,E] + the sequence idx

DanEnergetics (changes requested):

i6_models/parts/rasr_fsa.py
curufinwe: Ups, this is an error in the documentation of fbw_loss2. The new op does not have the sequence ID anymore for each edge separately. It's rather a separate tensor in the FSA.

DanEnergetics: Ah yes that's what I remember from reviewing the code, as well, just wanted to make sure.

PR #76, opened 2025-05-08
Use ruff format instead of black

curufinwe:


.github/workflows/ruff.yml
```diff
@@ -7,13 +7,13 @@ on:
...
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
     - uses: actions/setup-python@v4
       with:
         python-version: 3.8
         cache: 'pip'
-    - run: pip install black==22.3.0
-    - run: black --diff --check .
+    - run: pip install ruff==0.9.8
```
albertz, line 18: Why this version? It's already old.

(And also, I think it would be good to have it consistent across repos.)

michelwi (approved):

NeoLegends (approved):

Issue #81, opened 2025-06-26
__all__ contains wrong names


jtrmal:
Hi, in here https://github.com/rwth-i6/i6_models/blob/90537d492d33d6aab3c9d1764b4b82a7953c97e8/i6_models/parts/ffnn.py#L1C13-L1C30 the `__all__` variable contains wrong class names. Not sure if intentional or how big of an issue it is. Probably not much of an issue?

PR #82, opened 2025-06-26
fix incorrect export names in __all__, adresses #81

jtrmal:


jtrmal:
this should address #81

albertz (approved):

PR #84, opened 2025-08-07
add pca projection

Judyxujj:
Replace the fixed random projection layer in BestRQ with PCA projection learned from training data

### DIFF
--- a/i6_models/parts/best_rq/quantizer_with_PCA.py
+++ b/i6_models/parts/best_rq/quantizer_with_PCA.py
@@ -1,0 +1,251 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch.linalg import vector_norm
+from i6_experiments.users.jxu.experiments.pretrain.pytorch_networks.pca_module import IncrementalPCAonGPU
+
+__all__ = [
+    "PCAProjectionQuantizer",
+]
+
+
+class IncrementalPCAonGPU:
+    """
+    An implementation of Incremental Principal Components Analysis (IPCA) that leverages PyTorch for GPU acceleration.
+    The code is taken from https://github.com/dnhkng/PCAonGPU/blob/main/gpu_pca/pca_module.py.
+
+    This class provides methods to fit the model on data incrementally in batches, and to transform new data
+    based on the principal components learned during the fitting process.
+    """
+
+    def __init__(self, n_components=None, *, whiten=False, copy=True, batch_size=None):
+        """
+        :param n_components: Number of components to keep. If `None`, it's set to the minimum of the number of samples
+                             and features. Defaults to None.
+        :param whiten: When True, the `components_` vectors are divided to ensure uncorrelated outputs with unit
+                       component-wise variances. Defaults to False.
+        :param copy: If False, input data will be overwritten. Defaults to True.
+        :param batch_size: The number of samples to use for each batch. If `None`, it's inferred from the data and set to
+                           `5 * n_features`. Defaults to None.
+        """
+        self.n_components = n_components
+        self.whiten = whiten
+        self.copy = copy
+        self.batch_size = batch_size
+        self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+        # Set n_components_ based on n_components if provided
+        if n_components:
+            self.n_components_ = n_components
+
+        # Initialize attributes to avoid errors during the first call to partial_fit
+        self.mean_ = None  # Will be initialized properly in partial_fit based on data dimensions
+        self.var_ = None  # Will be initialized properly in partial_fit based on data dimensions
+        self.n_samples_seen_ = 0
+
+    def _validate_data(self, X, dtype=torch.float32, copy=True):
+        """
+        Validates and converts the input data `X` to the appropriate tensor format.
+
+        This method ensures that the input data is in the form of a PyTorch tensor and resides on the correct device (CPU or GPU).
+        It also provides an option to create a copy of the tensor, which is useful when the input data should not be overwritten.
+        """
+        if not isinstance(X, torch.Tensor):
+            X = torch.tensor(X, dtype=dtype).to(self.device)
+        if X.device == torch.device("cpu"):
+            X = X.to(self.device)
+        if copy:
+            X = X.clone()
+        return X
+
+    @staticmethod
+    def _incremental_mean_and_var(X, last_mean, last_variance, last_sample_count):
+        """
+        Computes the incremental mean and variance for the data `X`.
+
+        :param X: The batch input data tensor with shape (n_samples, n_features).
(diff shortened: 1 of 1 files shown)


curufinwe:
> I think the filename should be changed to all lowercase.
Agreed

i6_models/parts/best_rq/quantizer_with_PCA.py
```diff
@@ -0,0 +1,251 @@
...
+        """
+        self.n_components = n_components
+        self.whiten = whiten
+        self.copy = copy
+        self.batch_size = batch_size
+        self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+        # Set n_components_ based on n_components if provided
+        if n_components:
+            self.n_components_ = n_components
```
albertz, line 39:
```suggestion
        self.n_components_ = n_components
```

i6_models/parts/best_rq/quantizer_with_PCA.py
```diff
@@ -0,0 +1,251 @@
...
+
```
albertz, line 40:
```suggestion
```
This is not really needed.

Also, you should never only conditionally assign some attrib. The attrib should always be there, but maybe None.

i6_models/parts/best_rq/quantizer_with_PCA.py
```diff
@@ -0,0 +1,251 @@
...
+        # Initialize attributes to avoid errors during the first call to partial_fit
+        self.mean_ = None  # Will be initialized properly in partial_fit based on data dimensions
+        self.var_ = None  # Will be initialized properly in partial_fit based on data dimensions
+        self.n_samples_seen_ = 0
+
+    def _validate_data(self, X, dtype=torch.float32, copy=True):
```
albertz, line 46:
```suggestion
    def _validate_data(self, x, dtype=torch.float32, copy=True):
```
You should not have uppercase vars (I think this violates PEP8).

i6_models/parts/best_rq/quantizer_with_PCA.py
```diff
@@ -0,0 +1,251 @@
...
+        if not isinstance(X, torch.Tensor):
+            X = torch.tensor(X, dtype=dtype).to(self.device)
+        if X.device == torch.device("cpu"):
+            X = X.to(self.device)
+        if copy:
+            X = X.clone()
+        return X
+
+    @staticmethod
+    def _incremental_mean_and_var(X, last_mean, last_variance, last_sample_count):
```
albertz, line 62:
```suggestion
    def _incremental_mean_and_var(x, last_mean, last_variance, last_sample_count):
```

i6_models/parts/best_rq/quantizer_with_PCA.py
```diff
@@ -0,0 +1,251 @@
...
+        :param codebook_dim: number of dimension for vocab in the codebook
+        :param codebook_num_vars: vocab size of the codebook
+        :param use_pca_projection: whether to use pca projection, if set to False, random
+        """
+        super().__init__()
+
+        self.input_dim = input_dim
+        self.distance_meature = distance_meature
+
+        self.PCA = IncrementalPCAonGPU(n_components=codebook_dim)
```
albertz, line 218:
```suggestion
        self.pca = IncrementalPCAonGPU(n_components=codebook_dim)
```

i6_models/parts/best_rq/quantizer_with_PCA.py
```diff
@@ -0,0 +1,251 @@
...
+        self.register_buffer("PCA_components", torch.empty((codebook_dim, input_dim)))
```
albertz, line 219:
```suggestion
        self.register_buffer("pca_components", torch.empty((codebook_dim, input_dim)))
```

i6_models/parts/best_rq/quantizer_with_PCA.py
```diff
@@ -0,0 +1,251 @@
...
+                self.PCA_components = self.PCA.components_
+
+        if self.use_pca_projection and PCA_update_steps > 0:
+            x = x @ self.PCA_components.T
+            if normalise_after_PCA:
+                x = F.normalize(x, dim=-1)
+        else:
+            x = F.normalize(x @ self.P, dim=-1)
+
+        targets = vector_norm((self.CB.unsqueeze(1) - x.unsqueeze(1)), dim=-1).argmin(dim=1)
```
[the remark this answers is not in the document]
Judyxujj: because self.cb has shape (8192, 16) and x has shape (B, T, 16), directly do Subtraction `self.cb - x` will give error `RuntimeError: The size of tensor a (30) must match the size of tensor b (50) at non-singleton dimension 1`

i6_models/parts/best_rq/quantizer_with_pca.py
```diff
@@ -0,0 +1,250 @@
...
+import torch.nn.functional as F
+from torch.linalg import vector_norm
+from i6_experiments.users.jxu.experiments.pretrain.pytorch_networks.pca_module import IncrementalPCAonGPU
+
+__all__ = [
+    "PCAProjectionQuantizer",
+]
+
+
+class IncrementalPCAonGPU:
```
albertz: Why is the GPU aspect relevant? Wouldn't the same code work for any device?

Maybe make `device: torch.device` just an argument to the class?

Also, I see below that you actually also use CPU if GPU is not available...

Also, I wonder, shouldn't this be a Torch module?

```suggestion
class IncrementalPCA(nn.Module):
```

curufinwe (approved):

i6_models/parts/best_rq/quantizer_with_pca.py
```diff
@@ -0,0 +1,261 @@
...
+    """
+
+    def __init__(
+        self,
+        n_components: Optional[int] = None,
+        *,
+        whiten: bool = False,
+        copy: bool = True,
+        batch_size: Optional[int] = None,
+    ):
```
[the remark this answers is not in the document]
Judyxujj: I added `device: Optional[Literal["cpu", "cuda"]] = None` and as you suggested, if it is set to None, x.device from the first input x would be used.

i6_models/parts/best_rq/quantizer_with_pca.py
```diff
@@ -0,0 +1,271 @@
...
+        :param batch_size: The number of samples to use for each batch. If `None`, it's inferred from the data and set to
+                           `5 * n_features`. Defaults to None.
+        :param device: if set to None, then it use the x.device from the first input data x
+        """
+        self.n_components = n_components
+        self.whiten = whiten
+        self.copy = copy
+        self.batch_size = batch_size
+        if device is not None:
+            self.device = device
```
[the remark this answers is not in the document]
Judyxujj: I missed this one, just pushed the update

PR #91, opened 2025-12-22
Update documentation of input_embedding_scale parameter in TransformerDecoderV1Config

NeoLegends:
Mohammad did not use the scale in a pure LM setup.


michelwi (approved):

i6_models/assemblies/transformer/transformer_decoder_v1.py
```diff
@@ -128,7 +128,7 @@ class TransformerDecoderV1Config(ModelConfiguration):
         block_cfg: Configuration for TransformerDecoderV1.
         input_dropout: Dropout applied to the input embedding.
         input_embedding_scale: Scale applied to the input embedding.
-            Set to `None` to apply a (tuned) default.
+            Set to `None` to apply a default that is suitable for ASR AED decoder models.
```
albertz: I would not mention any specific model at all here. I think this is just confusing. I would instead just say what default you use.

albertz (approved):

PR #93, opened 2026-01-29
TransformerDecoder: optional positional encoding and final matmul

Gerstenberger:
Changes for positional encoding and the final matrix multiplication of model output and output embedding matrix to be both optional.

This allows us to use the implementation for self-normalized LM Transformer training, where positional encoding is not required and the final matmul is replaced by another matmul in the sampling loss.

My only question is: should this be a `TransformerDecoderV2` instead?

### DIFF
--- a/i6_models/assemblies/transformer/transformer_decoder_v1.py
+++ b/i6_models/assemblies/transformer/transformer_decoder_v1.py
@@ -20,7 +20,7 @@ from torch import nn, Tensor
 import torch.nn.functional as F
 
 from dataclasses import dataclass, field
-from typing import List, Optional, Tuple, TypedDict, Union
+from typing import List, Optional, Tuple, TypedDict, Union, NotRequired
 
 from i6_models.config import ModelConfiguration
 from i6_models.parts.conformer import (
@@ -141,6 +141,8 @@ class TransformerDecoderV1Config(ModelConfiguration):
         logits_bias: Whether to add a bias to the output logits.
             Usually False is a good choice.
         share_embedding: Whether to share the input and output embedding.
+        use_positional_encoding: use a sinus positional encoding on the initial input
+        do_output_embedding_matmul: apply the final model output x output embedding matmul
     """
 
     block_cfg: TransformerDecoderBlockV1Config
@@ -150,13 +152,15 @@ class TransformerDecoderV1Config(ModelConfiguration):
     num_output: int
     logits_bias: bool
     share_embedding: bool
+    use_positional_encoding: bool = True
+    do_output_embedding_matmul: bool = True
 
 
 class TransformerDecoderV1State(TypedDict):
     """Recurrent state of the transformer decoder."""
 
     block_state: List[TransformerDecoderBlockV1State]
-    pos: Tensor
+    pos: NotRequired[Tensor]
 
 
 class TransformerDecoderV1(nn.Module, ModuleWithState[TransformerDecoderV1State]):
@@ -190,12 +194,19 @@ class TransformerDecoderV1(nn.Module, ModuleWithState[TransformerDecoderV1State]
         else:
             self.out_logits = nn.Linear(self.model_dim, cfg.num_output, bias=cfg.logits_bias)
 
+        self.use_positional_encoding = cfg.use_positional_encoding
+        self.do_output_embedding_matmul = cfg.do_output_embedding_matmul
+
     def get_initial_state(self) -> TransformerDecoderV1State:
         """:return: initial decoder state"""
-        return {
+        state: TransformerDecoderV1State = {
             "block_state": [block.get_initial_state() for block in self.module_list],
-            "pos": torch.tensor(0, dtype=torch.int32),
         }
+
+        if self.use_positional_encoding:
+            state["pos"] = torch.tensor(0, dtype=torch.int32)
+
+        return state
 
     def transform_encoder_output(
         self,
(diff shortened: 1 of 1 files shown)


i6_models/assemblies/transformer/transformer_decoder_v1.py
```diff
@@ -150,13 +152,15 @@ class TransformerDecoderV1Config(ModelConfiguration):
     num_output: int
     logits_bias: bool
     share_embedding: bool
+    use_positional_encoding: bool = True
```
NeoLegends: I wonder if, instead of being a flag, this should be a configurable module instead, which you simply replace with a noop if you don't want any positional encoding. This would allow using other positional encoding schemes other than sinusoidal as well.

i6_models/assemblies/transformer/transformer_decoder_v1.py
```diff
@@ -150,13 +152,15 @@ class TransformerDecoderV1Config(ModelConfiguration):
...
+    do_output_embedding_matmul: bool = True
```
NeoLegends: Perhaps 
```suggestion
    embed_outputs_to_vocab_dim: bool = True
```
is clearer naming-wise?

albertz:
As a first comment (I will try to comment in more detail later): The same questions have been thought about in the RF implementation, for Transformer encoder, decoder, and very related also Conformer encoder (to make the frontend optional, etc).

[Current RF TransformerDecoder implementation](https://github.com/rwth-i6/returnn/blob/master/returnn/frontend/decoder/transformer.py). It already has the `pos_enc` configurable, and you can pass `None` to it to disable this. It doesn't have the logits optional though yet. But I have variants of this (in own code, not in RETURNN) where I have removed this part, or made it optional.

[RF ConformerEncoderV2](https://github.com/rwth-i6/returnn/blob/master/returnn/frontend/encoder/conformer_v2.py).

i6_models/assemblies/transformer/transformer_decoder_v1.py
albertz: Yes, agree, better would be to have this more dynamic.

`ConformerMHSARelPosV1._sinusoidal_pe` should maybe be moved to a separate function, and then you would have `positional_encoding=absolute_sinusoidal_positional_encoding` as default, and `None` is also allowed.

i6_models/assemblies/transformer/transformer_decoder_v1.py
```diff
@@ -190,13 +194,20 @@ def __init__(self, cfg: TransformerDecoderV1Config):
         else:
             self.out_logits = nn.Linear(self.model_dim, cfg.num_output, bias=cfg.logits_bias)
```
albertz, line 253: I just realize, this sharing is weird. I would always set `self.out_logits`. If sharing, you can just do `self.out_logits.weights = self.input_embedding.weight`. That would simplify the other code.

Also, `self.out_logits` should always be set (be `None` if not used). But with my suggestion, you don't need to care about this.

And then you would also allow to have logits_bias=True with share_embedding=True.

i6_models/assemblies/transformer/transformer_decoder_v1.py
albertz: I don't think it's cleaner. But I also don't like the original name. But I'm also not sure whether I like the logic at all (see my separate comment on this, why to have the `out_logits` at all if it is not used).

albertz: If this is False, and `not cfg.share_embedding`, the `out_logits` are not used at all. Does it make sense to even have them then?

Gerstenberger:
I made a initial proposal but I am not really sure about the changes. Maybe the proposal is too complicated/not straightforward enough, which I tend towards.

In general, would a `Callable` be sufficient or you want a `ModuleFactoryV1`?
If a `ModuleFactoryV1` would be optional, i.e. we do not apply a positional encoding at all, how do you specify 'not specified' and have the sinusoidal positional encoding as a default?
Having the callable is easier to handle.

Please let me know.

i6_models/assemblies/transformer/transformer_decoder_v1.py
```diff
@@ -128,6 +127,55 @@ def forward(
...
+
+    def forward(self, inputs: Tensor, lengths: Tensor, state: PositionalEncodingV1State):
+        """
+        Apply sinusoidal positional encoding on the inputs.
+
+        :param inputs: tensor to apply the positional encoding on
+        :param lengths: input lengths
+        :param state: current state of positional encoding.
+        """
+        sinus_pe = ConformerMHSARelPosV1._sinusoidal_pe(
```
Gerstenberger, line 157: Maybe should be moved to `primitives`?

i6_models/assemblies/transformer/transformer_decoder_v1.py
```diff
@@ -128,6 +127,56 @@ def forward(
...
+        """
+        Apply sinusoidal positional encoding on the inputs.
+
+        :param inputs: input embeddings
+        :param labels: input labels
+        :param lengths: input lengths
+        :param state: current state of positional encoding.
+        """
+        sinus_pe = ConformerMHSARelPosV1._sinusoidal_pe(
+            torch.arange(labels.shape[-1], device=labels.device) + state["pos"], self.embed_dim
```
Gerstenberger: Do we have the constraint `labels.shape[-1] == lenghts.max()`? If so labels input can be removed.
Or do we only return `sinus_pre.unsqueeze(0)` and apply the addition later?

i6_models/assemblies/transformer/transformer_decoder_v1.py
```diff
@@ -150,13 +201,15 @@ class TransformerDecoderV1Config(ModelConfiguration):
...
+    positional_encoding: Optional[ModuleFactoryV1]
+    output_linear_projection: bool = True
 
 
 class TransformerDecoderV1State(TypedDict):
     """Recurrent state of the transformer decoder."""
 
     block_state: List[TransformerDecoderBlockV1State]
-    pos: Tensor
+    pos_state: NotRequired[PositionalEncodingV1State]
```
Gerstenberger: Okay, maybe should not change names and type as this breaks existing setups.

i6_models/assemblies/transformer/transformer_decoder_v1.py
```diff
@@ -184,19 +237,33 @@ def __init__(self, cfg: TransformerDecoderV1Config):
         )
         self.out_norm = nn.LayerNorm(self.model_dim)
         self.share_embedding = cfg.share_embedding
-        if cfg.share_embedding:
-            assert not cfg.logits_bias, "Cannot use logits bias with shared embedding"
-            nn.init.xavier_uniform_(self.input_embedding.weight)  # bad convergence with default init
+
+        cfg.positional_encoding = None
```
hannah220: `cfg.positional_encoding` is set to None above, so the condition below will always be false. 
I think you meant to do `self.positional_encoding = None`.

